PR #194, opened 2022-09-29
Show shared balances

valzav:


brandonfancher (changes requested):
Hey! Coming together nicely!

1. When I send to an account with Manual Debit mode _on_, I have to refresh before I see it appear in the Pending balances list. If I send to someone with Manual Debit _off_, however, I see it appear right away in the bottom list without refreshing. Can you invalidate the query similarly so that the user doesn't have to refresh in the former case?
2. Same thing for when I "Accept" as the recipient or "Cancel" as the sender. The user shouldn't have to refresh to see it move to the Transfer history below.
3. Let's remove the "Cancel" button that isn't implemented yet for now.

valzav:
image

services/user/TokenSys/ui/src/App.tsx
```diff
@@ -76,6 +83,7 @@ function App() {
         try {
             await transfer(data);
             reset();
+            invalidateSharedBalancesQuery();
```
brandonfancher, line 86: This causes the query above to refetch...but not the instance of that query being consumed by `shared-balances.tsx` so it has no effect.

services/user/TokenSys/ui/src/App.tsx
```diff
@@ -36,6 +38,8 @@ function App() {
     const [formSubmitted, setFormSubmitted] = useState(false);
     const [transferHistoryResult, invalidateTransferHistoryQuery] =
         useTransferHistory(userName);
+    const [manualDebitMode, setManualDebitMode] = useState(false);
+    const [, invalidateSharedBalancesQuery] = useSharedBalances();
```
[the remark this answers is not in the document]
brandonfancher: Grab the first item in the destructured array `sharedBalancesResult` and pass it into the `<SharedBalances />` component as a prop. Don't call `useSharedBalances()` again/separately in `<SharedBalances />`.

services/user/TokenSys/ui/src/App.tsx
brandonfancher: I would still call this here (given my proposal above), but I do recommend moving it down a few lines next to `invalidateTransferHistoryQuery();`.

services/user/TokenSys/ui/src/App.tsx
```diff
@@ -232,6 +258,10 @@ function App() {
                     )}
                 </div>
             </form>
+            <SharedBalances
+                tokens={tokens}
+                currentUser={userName}
+            />
```
brandonfancher, line 264:
```js
            <SharedBalances
                tokens={tokens}
                currentUser={userName}
                queryResult={sharedBalancesResult}
                refetch={async () => {
                    await wait(1000);
                    invalidateTransferHistoryQuery();
                    invalidateSharedBalancesQuery();
                }}
            />
```

(Although it would be even better if you declare that async function separately and include it there like `refetch={onTokenAction}`. Then you could even call that in the `onSubmit()` above instead of calling the two invalidate functions separately.

Also, I did find that waiting 500-1000 ms makes this more reliable, since the blockchain can take a moment to update its state.

services/user/TokenSys/ui/src/views/shared-balances.tsx
```diff
@@ -0,0 +1,150 @@
...
+    tokens?: TokenBalance[];
+    currentUser: string;
+}
+
+export const SharedBalances = ({
+                                    tokens,
+                                    currentUser,
+                                }: Props) => {
+    const [result, invalidateSharedBalancesQuery] =
+        useSharedBalances();
```
brandonfancher, line 18: You'll want to remove this and use the props you pass in instead.

services/user/TokenSys/ui/src/views/shared-balances.tsx
```diff
@@ -0,0 +1,150 @@
...
+        if (incoming) {
+            console.log("Cancel deposit is not implemented yet");
+        } else {
+            await executeUncredit({
+                symbol,
+                receiver,
+                maxAmount,
+                memo: "Uncredit",
+            });
+            invalidateSharedBalancesQuery();
```
brandonfancher, line 35: pass in and call the `refetch` prop instead

services/user/TokenSys/ui/src/views/shared-balances.tsx
```diff
@@ -0,0 +1,150 @@
...
+
+    const onAccept = async (sender: string, symbol: string, amount: string) => {
+        console.log("Accept", currentUser, symbol, amount);
+        await executeDebit({
+            symbol,
+            sender,
+            amount,
+            memo: "Debit",
+        });
+        invalidateSharedBalancesQuery();
```
brandonfancher, line 47: pass in and call the `refetch` prop instead

brandonfancher (changes requested):
This is how you'd make the transfer tables update after the various actions. Happy to walk you through it if helpful.

services/user/TokenSys/ui/.eslintrc.cjs
```diff
@@ -10,12 +16,12 @@ module.exports = {
...
-            {
-                devDependencies: ["**/*.config.ts", "**/*.config.js"],
-            },
-        ],
+        // "import/no-extraneous-dependencies": [
+        //     "error",
+        //     {
+        //         devDependencies: ["**/*.config.ts", "**/*.config.js"],
+        //     },
+        // ],
```
brandonfancher, line 24: If we don't need this, let's just remove it instead of commenting it out.

services/user/TokenSys/ui/src/App.tsx
```diff
@@ -77,6 +89,7 @@ function App() {
         try {
             await transfer(data);
             reset();
+            refetchData();
```
brandonfancher, line 92: Let's move this down to line 105 and replace (the now duplicate) `invalidateTransferHistoryQuery()` there. Better it refetches, even after a failure.

services/user/TokenSys/ui/src/contracts.ts
```diff
@@ -42,27 +73,110 @@ export class TokenContract extends Service {
...
+            }
+            await this.debit(token.id, sender, amount, memo);
+        } catch (e) {
+            console.error("Debit operation failed:", e);
+        }
+    }
 
+    @Action
+    setUserConf(flag: string, enable: boolean) {
+        console.log("setUserConf Action", flag, enable);
```
brandonfancher, line 164: Let's remove this log

services/user/TokenSys/ui/src/contracts.ts
```diff
@@ -42,27 +73,110 @@ export class TokenContract extends Service {
...
+        return {
+            flag,
+            enable,
+        };
+    }
 
+    @Op()
+    public async setUserConfOp({ flag, enable }: SetUserConfPayload) {
+        console.log("TokenSys Operation: setUserConf", flag, enable);
```
brandonfancher, line 173: Here too. Unless you think there's value in keeping them around.

services/user/TokenSys/ui/src/queries.ts
```diff
@@ -61,3 +65,27 @@ export const useTransferHistory = (user?: string) => {
...
+        }
+      }
+    }
+  }
+}`;
+
+export const useSharedBalances = (user?: string) => {
+    const query = querySharedBalances();
+    return useGraphQLQuery("/graphql", query);
+};
```
brandonfancher, line 91: The `user` parameter is unused. Remove it.

Out of curiosity, how does the query know _who_ to get shared balances for? I tested it, and it's not picking up other users' shared balances IIRC. So that's good.

services/user/TokenSys/ui/src/views/shared-balances.tsx
```diff
@@ -0,0 +1,185 @@
...
+        (e: any) => e.node
+    );
+
+    const onCancel = async (
+        incoming: boolean,
+        receiver: string,
+        symbol: string,
+        maxAmount: string
+    ) => {
+        console.log("Cancel", incoming);
```
brandonfancher, line 30: Do we need these logs throughout?

src
```diff
@@ -0,0 +1,279 @@
+import React, { useEffect, useState } from "react";
```
brandonfancher, line 1: I imagine this file wasn't supposed to be checked in?

brandonfancher (changes requested):
Looking good. Just a few minor changes.

services/user/TokenSys/ui/src/queries.ts
valzav: It does fetch all users shared balances (last 1000 records), making it return only user-related shared balances requires some backend work. I believe this task should address this if implemented https://github.com/orgs/gofractally/projects/4/views/1

brandonfancher: Ah, that makes sense. 👍

brandonfancher (approved):
Looks good!

PR #198, opened 2022-10-01
Add psispace wrapper

sparkplug0025:
Add rust wrappers for the following services:
- auth-ec-sys
- common-sys
- nft-sys
- proxy-sys
- psispace-sys

### DIFF
--- a/rust/psibase/src/lib.rs
+++ b/rust/psibase/src/lib.rs
@@ -3,6 +3,7 @@ mod block;
 mod crypto;
 mod db;
 mod from_bin;
+mod http;
 mod method_number;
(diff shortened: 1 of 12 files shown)


rust/psibase/src/lib.rs
```diff
@@ -3,6 +3,7 @@ mod block;
 mod crypto;
 mod db;
 mod from_bin;
+mod http;
```
tbfleming, line 6: Needs a `pub use` below

tbfleming (approved):
Nice work! I need this for what I'm working on now, so I'll go ahead and merge then make http pub.

PR #204, opened 2022-10-07
Store and consume fonts from common location

brandonfancher:
- Removes references to external Google font resources
- Places variable Raleway and Red Hat Mono fonts in `/common/fonts/`
- Updates `DefaultTestChain.cpp`, `boot.rs`, and `CMakeLists.txt` to include those
- Updates the CommonSys ui, React applet template/generator, and all applets to reference those fonts externally
- Updates all vite configurations to properly handle those external assets in both dev and release mode


generator/tools/templates/react/vite.config.ts
```diff
@@ -49,13 +49,21 @@ const psibase = (appletContract: string) => {
...
-                            "/common": path.resolve("../../CommonSys/common"),
-                        },
+                            },
+                            {
+                                // bundle non-external (above) common files except fonts (which should only be referenced)
+                                find: /^\/common(?!\/(?:fonts))(.*)$/,
+                                replacement: path.resolve(
+                                    "../../CommonSys/common$1"
+                                ),
+                            },
```
brandonfancher, line 65: At first, I tried simply adding the font files to the `rollupOptions.external` array above. This works fine for js modules. But I learned that `external` is _only_ for js modules and applies that mime type to any files in that array. It works in dev (hmr) mode, but not when running from a build; some browsers reject the css and fonts because they're improperly mime typed as javascript modules.

services/user/TokenSys/ui/vite.config.ts
```diff
@@ -76,11 +83,15 @@ const psibase = (appletContract: string) => {
...
+        react({
+            babel: {
+                parserOpts: {
+                    plugins: ["decorators-legacy"],
+                },
+            },
+        }),
+        svgr({ exportAsDefault: true }),
+        psibase("token-sys"),
+    ],
```
brandonfancher, line 96: Just a reformat

tbfleming (approved):

PR #205, opened 2022-10-10
Init rust tables

sparkplug0025:
Draft of a rust service example.

With this PR we are kicking off the table operations in rust.

We now have the following kv operations: put, get, last.

To come in next PRs:
- Macros to ease the tables handlers
- ~~Implementation of a DoubleEndedIterator~~
- Secondary keys management


rust/psibase/src/native.rs
```diff
@@ -51,6 +51,22 @@ pub fn check(condition: bool, message: &str) {
...
+    }
+    opt_value.unwrap()
+}
+
+/// Abort with message if optional has value
+pub fn check_none<T>(opt_value: Option<T>, message: &str) -> T {
+    if !opt_value.is_none() {
+        abort_message_bytes(message.as_bytes());
+    }
+    opt_value.unwrap()
```
tbfleming, line 67: This will always abort

rust/psibase/src/native.rs
```diff
@@ -218,39 +235,36 @@ pub fn kv_get<V: PackableOwned, K: ToKey>(
...
+    get_optional_result_bytes(size)
+}
+
+/// Get the maximum key-value pair which has key as a prefix
+///
+/// If one is found, then returns the value. Use [get_key] to
+/// get the found key.
+pub fn kv_max<K: ToKey, V: PackableOwned>(db_id: DbId, key: &K) -> Option<V> {
+    let bytes = kv_max_bytes(db_id, &key.to_key());
+    bytes.map(|v| V::unpack(&v[..], &mut 0).unwrap()) // unwrap won't panic
```
tbfleming, line 269: It will, and should, panic if the data is corrupt. e.g. someone didn't follow the rules at https://doc-sys.psibase.io/cpp-service/reference/table.html#schema-changes

tbfleming: `unpacked()` is more straight-forward than `unpack()`

rust/psibase/src/to_key.rs
```diff
@@ -45,6 +45,21 @@ pub trait ToKey {
     }
 }
 
+/// A serialized key (non-owning)
```
tbfleming, line 48: `Vec<u8>` is owning

rust/test_contract/src/lib.rs
```diff
@@ -5,7 +5,7 @@
 /// This service adds and multiplies `i32` numbers.
 ///
 /// This is where a detailed description would go.
-#[psibase::service]
+#[psibase::service()]
```
tbfleming, line 8: The macro should work without `()`

tbfleming: heh. The () was accidentally in my doc. Fixing...

rust/psibase/src/to_key.rs
sparkplug0025: Good point about this. I renamed to `RawKey` to avoid confusion with C++ side and kept it owned.

rust/psibase/src/native.rs
sparkplug0025: Ooops! Nice catch!

rust/psibase/src/db.rs
```diff
@@ -4,6 +4,7 @@
 /// various purposes. This enum identifies which database to
 /// use when invoking those functions.
 #[repr(u32)]
+#[derive(Clone)]
```
tbfleming, line 7: Make this Copy also. It will eliminate the need for .to_owned().

rust/psibase/src/table.rs
```diff
@@ -0,0 +1,170 @@
...
+
+impl<Key: ToKey, Record: TableRecord> Iterator for TableIndex<Key, Record> {
+    type Item = Record;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        if self.is_end {
+            return None;
+        }
+
+        self.front_key.data.push(0);
```
tbfleming, line 107: This skips the first record

rust/psibase/src/to_key.rs
```diff
@@ -45,6 +45,28 @@ pub trait ToKey {
...
+
+impl RawKey {
+    pub fn new(data: Vec<u8>) -> Self {
+        RawKey { data }
+    }
+}
+
+impl ToKey for RawKey {
+    fn append_key(&self, key: &mut Vec<u8>) {
+        for byte in &self.data {
```
tbfleming, line 64: I forgot the name, but Vec has a method which appends data from an iterator

tbfleming (approved):
You may address the above in another PR.

rust/psibase/src/to_key.rs
sparkplug0025: `extend_from_slice`, done! 👍

PR #208, opened 2022-10-13
Init secondary keys

sparkplug0025:
Initialize secondary keys.

Currently working functionalities:
- inserting new keys
- updating keys when record is updated and change is relevant
- removing keys when record is removed

Next steps:
- Explore better type handling for secondary keys, heterogeneous lists, macros etc.
- Implement the macros itself for the tables

### DIFF
--- a/rust/psibase/src/native.rs
+++ b/rust/psibase/src/native.rs
@@ -180,6 +180,15 @@ pub fn kv_put<'a, K: ToKey, V: Packable<'a>>(db: DbId, key: &K, value: &V) {
     kv_put_bytes(db, &key.to_key(), &value.packed())
 }
 
+/// Set a key-value pair if the key is unique
+///
+/// If key already exists, throws an error
+pub fn kv_insert_unique<'a, K: ToKey, V: Packable<'a>>(db: DbId, key: &K, value: &V) {
+    let key_bytes = key.to_key();
+    check_none(kv_get_bytes(db, &key_bytes), "key already exists");
+    kv_put_bytes(db, &key_bytes, &value.packed())
+}
+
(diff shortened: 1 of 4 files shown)


rust/psibase/src/native.rs
```diff
@@ -180,6 +180,15 @@ pub fn kv_put<'a, K: ToKey, V: Packable<'a>>(db: DbId, key: &K, value: &V) {
     kv_put_bytes(db, &key.to_key(), &value.packed())
 }
 
+/// Set a key-value pair if the key is unique
+///
+/// If key already exists, throws an error
+pub fn kv_insert_unique_bytes<K: ToKey>(db: DbId, key: &K, bytes: &[u8]) {
```
tbfleming, line 186: Name it `.._put_..` for consistency

tbfleming (approved):

rust/psibase/src/native.rs
swatanabe: Not sure I agree on that.  The reason we settled on `put` in the first place is that `put` implies that it overwrites existing, which is precisely what this function doesn't do.

sparkplug0025: I left as `kv_insert_unique` to follow the same pattern as the c++ side: https://github.com/gofractally/psibase/blob/sparkplug0025/rust-elections-tables3/libraries/psibase/common/include/psibase/Table.hpp#L479-L490

I will merge as it is for now to unblock me, but let's keep the discussion here...

Maybe I should move this function out of this native file then and add it to the table abstraction keeping the `insert` name? (Since it's just a sugar around the kv get/put ops?)

PR #209, opened 2022-10-15
Table wrapper with secondary keys

sparkplug0025:
Found a better design for the table wrappers and getting secondary key index.

### DIFF
--- a/rust/psibase/src/table.rs
+++ b/rust/psibase/src/table.rs
@@ -29,62 +29,63 @@ pub trait TableRecord: PackableOwned {
     }
 }
 
-pub trait TableRecordWithSecondaryKey: TableRecord {
-    type SecondaryKey: ToKey;
-
-    fn get_secondary_key(&self) -> Self::SecondaryKey;
+pub struct Table {
+    pub db_id: DbId,
+    pub prefix: Vec<u8>,
 }
 
-pub trait TableHandler<Record: TableRecord> {
+pub trait TableBase {
+    fn prefix(&self) -> Vec<u8>;
+    fn db_id(&self) -> DbId;
+}
+
+pub trait TableHandler {
+    type TableType;
     const TABLE_INDEX: u16;
     const TABLE_SERVICE: AccountNumber;
 
-    fn open() -> Table<Record> {
+    fn new(table: Table) -> Self::TableType;
+
+    fn open() -> Self::TableType {
         let prefix = (Self::TABLE_SERVICE, Self::TABLE_INDEX).to_key();
-        Table {
+        let table = Table {
             db_id: DbId::Service,
             prefix,
-            phantom: PhantomData,
-        }
+        };
+        Self::new(table)
     }
 }
 
-pub struct Table<Record: TableRecord> {
-    db_id: DbId,
-    prefix: Vec<u8>,
-    phantom: PhantomData<Record>,
-}
-
-impl<Record: TableRecord> Table<Record> {
+pub trait TableWrapper<Record: TableRecord>: TableBase {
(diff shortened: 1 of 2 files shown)


rust/test_service_elections/src/lib.rs
```diff
@@ -153,7 +238,7 @@ mod service {
         );
 
         let table = CandidatesTable::open();
-        let mut idx = table.get_index::<ByElectionVotesCount>(1);
+        let mut idx = table.get_index_by_election_votes_key();
```
sparkplug0025, line 241: The idx is typed for the user, since we will auto generate the method from the macro. So I believe we are ok now.

rust/test_service_elections/src/lib.rs
```diff
@@ -18,20 +18,61 @@ mod service {
...
+    impl TableBase for ElectionsTable {
+        fn prefix(&self) -> Vec<u8> {
+            self.0.prefix.clone()
+        }
 
-    impl TableHandler<ElectionRecord> for ElectionsTable {
+        fn db_id(&self) -> DbId {
+            self.0.db_id
+        }
+    }
```
sparkplug0025, line 55: What I did not like was these repetitions of:

```rust
    impl TableBase for XTable {
        fn prefix(&self) -> Vec<u8> {
            self.0.prefix.clone()
        }

        fn db_id(&self) -> DbId {
            self.0.db_id
        }
    }
```

It will always be the same... Any suggestions?

rust/psibase/src/table.rs
```diff
@@ -124,20 +125,20 @@ impl<Record: TableRecord> Table<Record> {
...
                 old_key.append_key(&mut key_buffer);
-                kv_remove(self.db_id, &KeyView::new(&key_buffer));
-                key_buffer.truncate(self.prefix.len() + 1);
+                kv_remove(self.db_id(), &KeyView::new(&key_buffer));
+                key_buffer.truncate(self.prefix().len() + 1);
 
                 new_key.append_key(&mut key_buffer);
-                kv_insert_unique_bytes(self.db_id, &KeyView::new(&key_buffer), pk);
-                key_buffer.truncate(self.prefix.len());
+                kv_insert_unique_bytes(self.db_id(), &KeyView::new(&key_buffer), pk);
```
tbfleming, line 132: Long term, not needed in this PR, it'd be nice to do the dup check before any writes then return an error

tbfleming: The C++ code aborts because the error approach we once preferred ended up causing clang to bloat the wasms.

tbfleming:
```
    /// Set the range of available records for the iterator; useful for viewing a slice of the records
    ///
    /// - `from` is inclusive, `to` is not: [from..to)
    /// - using this after the start of an iteration process (for..loops, next(), next_back() etc)
    /// can mess the iteration sequence order, since it resets the front and the back cursors
    pub fn range(&mut self, from: Key, to: Key) -> &mut Self {
```

I think this weirdness exists because `TableIndex` is a cross between container-like functionality and iterators. We should probably split that up (future PR). We can then make it act more like BTreeMap, including its more flexible `range` function.

tbfleming:
`next_back` doesn't have that issue. Just `next`.

tbfleming:
Potential optimization: I suspect it's possible for `TableRecord::get_secondary_keys` to return a `&'static [RawKey]`. The implementation could return a const.

tbfleming:
scratch that. I forgot RawKey is a struct with a converted key, not a trait with a key conversion function.

sparkplug0025:
Thanks for all the comments, will address each item in a future PR.

PR #213, opened 2022-10-24
User/jmart/add events doc

James-Mart:
This PR breaks the wallet applet, front-end needs to be updated to use the graphql interface, rather than rest


James-Mart:
* `getTokenTypes` rest query must be replaced by the graphql query:
```
query {
    tokenTypes
    {
        pageInfo {
            hasNextPage
            endCursor
        }
        edges {
            node {
                id
                precision {
                    value
                }
                symbolId
            }
        }
    }
}
```
* `getUserConf` rest query must be replaced by the graphql query: 
```
query {
    getUserConf(user: "alice", flag: "manualDebit" )
}
```
* `balances` rest query must be replaced by the graphql query:
```
query {
    userBalances(user: "alice") {
        key {
            account
            tokenId
        }
        balance
    }
}
```

doc/psidk/src/concepts/events.md
```diff
@@ -0,0 +1,308 @@
...
+| 7   | `4`       | ...   | ...   | ...  |
+
+- event head = 7
+- event chain = 7, 4, 2
+
+<br/>
+
+For example, in the example Token Service, a `Transferred` event is defined, and is included in an index of events called `UserEvents`:
+
+<details>
```
tbfleming, line 65: Nit pick: I'm not a fan of pages that require you to click to reveal

libraries/psibase/common/include/psibase/eventIndex.hpp
```diff
@@ -0,0 +1,54 @@
...
+
+namespace psibase
+{
+   // This does nothing, but causes an error when called from a `consteval` function.
+   inline void expectedNullTerminatedArray() {}
+
+   // StringLiteral idea taken from:
+   // https://stackoverflow.com/questions/68024563/passing-a-string-literal-to-a-template-char-array-parameter
+   template <size_t N>
+   struct StringLiteral
```
tbfleming, line 11: Redundant with `psio::FixedString`

services/system/AuthEcSys/src/RAuthEcSys.cpp
```diff
@@ -9,16 +10,45 @@
...
+   struct AuthEcQuery
+   {
+      auto accWithKey(psibase::PublicKey key) const
+      {
+         auto idx = AuthEcSys::Tables{AuthEcSys::service}
+                        .open<AuthEcSys::AuthTable>()
+                        .getIndex<1>()
+                        .subindex(key);
+
+         optional<std::tuple<AccountNumber>> gt;
```
tbfleming, line 29: Make these args of `accWithKey`

tbfleming: But without tuple or the user side of the query gets icky. Unfortunately you'll have to translate between `optional<AccountNumber>` and `optional<tuple<AccountNumber>>`

tbfleming (approved):

PR #218, opened 2022-11-01
Doc tables (rebased)

sparkplug0025:


rust/test_service_elections/src/lib.rs
```diff
@@ -19,7 +19,7 @@ mod service {
     const MIN_ELECTION_TIME_SECONDS: TimePointSec = TimePointSec { seconds: 60 * 60 };
 
     #[table(name = "ElectionsTable", index = 0)]
-    #[derive(Debug, Reflect, Serialize, Deserialize)]
+    #[derive(Debug, Fracpack, Reflect, Serialize, Deserialize)]
```
sparkplug0025, line 22: Any reason why not to automagically derive Fracpack in the `table` macros?

tbfleming: It seems easier to document A requires B (common in Rust), than to document A provides B. The second leads to questions like why doesn't it also provide C, D, E, ...

PR #219, opened 2022-11-03
Replicates psispace in rust

sparkplug0025:
- Replicates psispace-sys in rust
- Implements table.erase
- Adds number of secondary keys in the `Table` trait (to optimize erase operation)
- General improvements on the election service (now it supports current time and serves the simple ui)
- Improves table interface adding the methods: `new(), with_service() and with_prefix()`

### DIFF
--- a/rust/psibase/src/table.rs
+++ b/rust/psibase/src/table.rs
@@ -47,8 +47,11 @@ pub trait TableRecord: PackableOwned {
 pub trait Table<Record: TableRecord>: Sized {
     const TABLE_INDEX: u16;
     const TABLE_SERVICE: AccountNumber;
+    const SECONDARY_KEYS: u8;
 
     fn new(db_id: DbId, prefix: Vec<u8>) -> Self;
+    fn prefix(&self) -> &[u8];
+    fn db_id(&self) -> DbId;
 
     fn open() -> Self {
         let prefix = (Self::TABLE_SERVICE, Self::TABLE_INDEX).to_key();
@@ -63,9 +66,6 @@ pub trait Table<Record: TableRecord>: Sized {
     fn create_table_from_prefix(prefix: Vec<u8>) -> Self {
         Self::new(DbId::Service, prefix)
     }
-
-    fn prefix(&self) -> &[u8];
-    fn db_id(&self) -> DbId;
 
     /// Returns one of the table indexes: 0 = Primary Key Index, else secondary indexes
     fn get_index<Key: ToKey>(&self, idx: u8) -> TableIndex<Key, Record> {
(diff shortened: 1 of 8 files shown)


rust/test_service_elections/src/lib.rs
```diff
@@ -71,6 +71,10 @@ mod service {
         }
     }
 
+    fn get_current_time() -> TimePointSec {
+        transaction_sys::Wrapper::call().headBlockTime()
```
tbfleming, line 75: headBlockTime is marked `TODO: remove`. Use `currentBlock()`, which is more suitable than `headBlock()` for most cases.

rust/psibase/src/table.rs
```diff
@@ -46,25 +47,24 @@ pub trait TableRecord: PackableOwned {
...
 
-    fn open() -> Self {
-        let prefix = (Self::TABLE_SERVICE, Self::TABLE_INDEX).to_key();
+    fn new() -> Self {
+        let prefix = (get_service(), Self::TABLE_INDEX).to_key();
         Self::create_table_from_prefix(prefix)
     }
 
-    fn open_custom(service: AccountNumber, table_index: u16) -> Self {
+    fn with_service(service: AccountNumber, table_index: u16) -> Self {
```
tbfleming, line 61: remove `table_index`. This function handles any time service A reads service B's tables.

rust/psibase/src/table.rs
```diff
@@ -46,25 +47,24 @@ pub trait TableRecord: PackableOwned {
...
         let prefix = (service, table_index).to_key();
         Self::create_table_from_prefix(prefix)
     }
 
     fn create_table_from_prefix(prefix: Vec<u8>) -> Self {
```
tbfleming, line 66: Users providing the prefix is a rare case, so having just the 2-argument version is enough.

tbfleming (approved):

rust/psibase/src/table.rs
sparkplug0025: What if service A, for convenience, also have the same table as table B under a different index? Eg. I could see services reusing the definition from `psispace-sys.ContentTable`. Should we always encourage duplicating the declaration?

rust/psibase/src/table.rs
sparkplug0025: the `create_table_from_prefix` is used from the `with_service()` and also from `new()`. I did not mean to expose this method but I guess we can't hide it in a trait.

sparkplug0025:
I'm merging but I can address above comments in a next PR.

rust/psibase/src/table.rs
tbfleming: See https://doc-sys.psibase.io/rust-service/tables.html#storing-structs-defined-elsewhere

rust/psibase/src/table.rs
tbfleming: It's small enough to inline in the other functions

rust/psibase/src/table.rs
sparkplug0025: I was working in the Global WebContent table and I realized that the consumer does not need to redeclare a shared struct keys like in the docs example.

Here's the WebContent global record struct: 
https://github.com/gofractally/psibase/pull/222/files#diff-64aad0707509e772a7c54f564cd388b06db283a4d6bb5042e9bcdd548a5d96cdR219

And here's how a simplified macro could be (to replace the following lines Table impl):
https://github.com/gofractally/psibase/pull/222/files#diff-ed73e7a1a9fad9c7b0753d5de5101331739bff714eef72fd1a054cbe72ba8e4cR75-R77

What do you think? Should I pursue this macro?

PR #221, opened 2022-11-04
update image builder workflow

James-Mart:


tbfleming:
Oh wow that deprecation is recent: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

PR #252, opened 2022-11-30
Add Contract Queries Macros

sparkplug0025:
Queries can be implemented by using the following macros:

```
    #[queries]
    impl Queries {
        /// List all the existing files
        table_query!(content, ContentTable, get_index_pk: ContentKeyNamed);

        /// List files content by account subindex
        table_query_subindex!(
            content_by_account,
            ContentTable,
            get_index_pk,
            account: AccountNumber
        );
    }
```

- `table_query!` is used to wrap an index function and its named keys type with the `first, last, before, after, gt, ge, lt, le` query functions
- `table_query_subindex!` is used to wrap a subindex for a given index function and the defined subindex partial fields.  
  - the last parameter of this query can set the subindex RemainingKey type if it's a single ident parameter such as `Vec<u8>`, `String`, `MyStruct`

### DIFF
--- a/rust/psibase/src/serve_http.rs
+++ b/rust/psibase/src/serve_http.rs
@@ -178,13 +178,21 @@ pub fn serve_graphql<Query: async_graphql::ObjectType + 'static>(
                 headers: vec![],
             })
         } else {
-            let request = receive_body(
+            println!("receiving graphql post...");
+            let request_result = receive_body(
                 Some(&request.contentType),
                 request.body.as_ref(),
                 Default::default(),
             )
-            .await
-            .unwrap();
+            .await;
+
+            if let Err(err) = &request_result {
+                check(false, &format!("err parsing graphql query {}", err));
+            }
+
+            let request = request_result.unwrap();
+            println!("received gql post!!! {:?}", request);
+
             let res = schema.execute(request).await;
             Some(HttpReply {
                 contentType: "application/json".into(),
--- a/rust/psibase/src/serve_http.rs
+++ b/rust/psibase/src/serve_http.rs
@@ -178,7 +178,6 @@ pub fn serve_graphql<Query: async_graphql::ObjectType + 'static>(
                 headers: vec![],
             })
         } else {
-            println!("receiving graphql post...");
             let request_result = receive_body(
                 Some(&request.contentType),
                 request.body.as_ref(),
@@ -190,10 +189,7 @@ pub fn serve_graphql<Query: async_graphql::ObjectType + 'static>(
                 check(false, &format!("err parsing graphql query {}", err));
             }
 
-            let request = request_result.unwrap();
-            println!("received gql post!!! {:?}", request);
-
-            let res = schema.execute(request).await;
+            let res = schema.execute(request_result.unwrap()).await;
             Some(HttpReply {
                 contentType: "application/json".into(),
                 body: serde_json::to_vec(&res).unwrap().into(),
--- a/rust/psibase_macros/src/graphql_macro.rs
+++ b/rust/psibase_macros/src/graphql_macro.rs
@@ -20,12 +20,21 @@ struct Args {
 
 impl Parse for Args {
     fn parse(input: ParseStream) -> Result<Self> {
-        let mut vars = Punctuated::<TypeParam, Token![,]>::parse_terminated(input)?.into_iter();
+        let query_name = input.parse().expect("Query name not present");
+        input.parse::<Token![,]>().expect("expected ,");
 
-        let query_name = vars.next().expect("Query name not present").ident;
-        let table = vars.next().expect("Table name not present").ident;
-        let key_fn = vars.next().expect("Table key function not present");
-        let subindex_params = vars.collect();
+        let table = input.parse().expect("Table name not present");
+        input.parse::<Token![,]>().expect("expected ,");
+
+        let key_fn = input.parse().expect("Table key function not present");
+
+        let subindex_params = if input.parse::<Token![,]>().is_ok() {
+            Punctuated::<TypeParam, Token![,]>::parse_terminated(input)?
+                .into_iter()
+                .collect()
+        } else {
+            Vec::new()
+        };
 
         Ok(Args {
             query_name,
@@ -95,7 +104,10 @@ pub fn table_query_macro_impl(item: TokenStream) -> TokenStream {
 
     let query_name = args.query_name;
     let table = args.table;
+
+    // TODO: This specific naming convention is a bit of a hack. Figure out a way of using an associated type
     let record = Ident::new(format!("{}Record", table).as_str(), table.span());
+
     let key_fn = args.key_fn.ident;
     let key_ty = args.key_fn.bounds.to_token_stream();
 
@@ -135,7 +147,10 @@ pub fn table_query_subindex_macro_impl(item: TokenStream) -> TokenStream {
 
     let query_name = args.query_name;
     let table = args.table;
+
+    // TODO: This specific naming convention is a bit of a hack. Figure out a way of using an associated type
     let record = Ident::new(format!("{}Record", table).as_str(), table.span());
+
     let key_fn = args.key_fn.ident;
     let subindex_params = args.subindex_params;
 
@@ -147,7 +162,7 @@ pub fn table_query_subindex_macro_impl(item: TokenStream) -> TokenStream {
     }
 
     let mut params = quote! {};
-    let mut subindex_args = quote! {};
+    let mut subkey_data = quote! {};
 
     // TODO: Should string be accepted as a default rest type of any subindex?
     let mut subindex_rest_ty = quote! { String };
@@ -163,9 +178,9 @@ pub fn table_query_subindex_macro_impl(item: TokenStream) -> TokenStream {
                 #param_name: #param_type,
             };
 
-            subindex_args = quote! {
-                #subindex_args
-                &#param_name,
+            subkey_data = quote! {
+                #subkey_data
+                #param_name.append_key(&mut subkey_data);
             };
         }
     }
@@ -180,7 +195,10 @@ pub fn table_query_subindex_macro_impl(item: TokenStream) -> TokenStream {
             after: Option<String>,
         ) -> async_graphql::Result<async_graphql::connection::Connection<psibase::RawKey, #record>> {
             let table_idx = #table::new().#key_fn ();
-            let sidx = psibase::TableQuery::subindex::<#subindex_rest_ty>(table_idx, #subindex_args);
+            let mut subkey_data: Vec<u8> = Vec::new();
+            #subkey_data
+            let subkey = psibase::RawKey::new(subkey_data);
+            let sidx = psibase::TableQuery::subindex::<#subindex_rest_ty>(table_idx, &subkey);
             sidx
                 .first(first)
                 .last(last)
--- a/rust/psibase_macros/src/graphql_macro.rs
+++ b/rust/psibase_macros/src/graphql_macro.rs
@@ -1,0 +1,164 @@
+use darling::ToTokens;
+use proc_macro::TokenStream;
+use proc_macro2::Ident;
+use proc_macro_error::abort;
+use quote::quote;
+use syn::{
+    parse::{Parse, ParseStream},
+    parse_macro_input,
+    punctuated::Punctuated,
+    AttrStyle, ImplItem, Item, Result, Token,
+};
+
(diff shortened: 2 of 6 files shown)


rust/psibase_macros/src/graphql_macro.rs
```diff
@@ -0,0 +1,164 @@
...
+    parse_macro_input,
+    punctuated::Punctuated,
+    AttrStyle, ImplItem, Item, Result, Token,
+};
+
+struct Args {
+    query_name: Ident,
+    table: Ident,
+    key_fn: Ident,
+    sub_index_params: Vec<(Ident, Ident)>,
```
swatanabe, line 17: Should we allow types that are represented by more than one token, i.e. using syn::TypeParam?

rust/test_service_psispace/src/lib.rs
```diff
@@ -39,6 +41,15 @@ mod service {
         }
     }
 
+    #[queries]
+    impl Queries {
+        /// List all the existing files
+        table_query!(all_content, ContentTable, get_index_pk);
+
+        /// List files content by account
+        table_query!(content, ContentTable, get_index_pk, account, AccountNumber);
```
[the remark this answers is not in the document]
sparkplug0025: Done!

rust/psibase_macros/src/graphql_macro.rs
sparkplug0025: Great idea, done!

rust/test_service_psispace/src/lib.rs
```diff
@@ -39,6 +41,40 @@ mod service {
         }
     }
 
+    // TODO: create a macro for unnamed field keys (eg as seen in tuples keys),
+    // so that we can support these types of tables in GQL queries.
+    #[derive(ToKey, InputObject)]
+    pub struct ContentKeyNamed {
+        pub account: AccountNumber,
+        pub path: String,
+    }
```
sparkplug0025, line 50: Or should we encourage using structs as keys and not tuples?
@tbfleming @swatanabe thoughts?

rust/psibase_macros/src/graphql_macro.rs
```diff
@@ -0,0 +1,195 @@
...
+
+    impl_query.into()
+}
+
+pub fn table_query_macro_impl(item: TokenStream) -> TokenStream {
+    let args = parse_macro_input!(item as Args);
+
+    let query_name = args.query_name;
+    let table = args.table;
+    let record = Ident::new(format!("{}Record", table).as_str(), table.span());
```
swatanabe, line 98: Requiring a specific naming convention seems dangerous. Can't it be deduced as an associated type of the table type?

rust/psibase_macros/src/graphql_macro.rs
```diff
@@ -0,0 +1,195 @@
...
+    let mut params = quote! {};
+    let mut subindex_args = quote! {};
+
+    // TODO: Should string be accepted as a default rest type of any subindex?
+    let mut subindex_rest_ty = quote! { String };
+
+    for param in subindex_params {
+        let param_name = param.ident;
+        let param_type = param.bounds.to_token_stream();
+        if param_name == "subindex_rest_ty" {
```
swatanabe, line 158: This seems really clunky, especially for a parameter that does not affect the behavior at all.

rust/psibase_macros/src/graphql_macro.rs
```diff
@@ -0,0 +1,195 @@
...
+    }
+
+    let query = quote! {
+        async fn #query_name(
+            &self,
+            #params
+            first: Option<i32>,
+            last: Option<i32>,
+            before: Option<String>,
+            after: Option<String>,
```
swatanabe, line 180: Is there are reason that `gt`, `lt`, &c. are not used for subindex?

rust/psibase/src/serve_http.rs
```diff
@@ -178,14 +178,18 @@ pub fn serve_graphql<Query: async_graphql::ObjectType + 'static>(
...
                 Some(&request.contentType),
                 request.body.as_ref(),
                 Default::default(),
             )
-            .await
-            .unwrap();
-            let res = schema.execute(request).await;
+            .await;
+
+            if let Err(err) = &request_result {
```
tbfleming, line 188: Rust's new let-else syntax is good for this pattern. We should gradually start moving code towards it.

rust/psibase_macros/src/graphql_macro.rs
```diff
@@ -0,0 +1,195 @@
...
+        ) -> async_graphql::Result<async_graphql::connection::Connection<psibase::RawKey, #record>> {
+            let table_idx = #table::new().#key_fn ();
+            let sidx = psibase::TableQuery::subindex::<#subindex_rest_ty>(table_idx, #subindex_args);
```
swatanabe, line 183: `subindex` only takes a single argument for the subkey.

rust/test_service_psispace/src/lib.rs
swatanabe: GraphQL doesn't have a notion of tuples. I think the GraphQL type should always be a struct. We can define `derive(ToKey)` in way that makes the struct usable where a tuple key is expected.

tbfleming: If we leave tuples unsupported and encourage structs for keys, then it leaves open the possibility of coming up with a good solution in the future.

rust/psibase_macros/src/graphql_macro.rs
tbfleming: Keeping these separate allows restricting the range with these args within a subindex

rust/psibase_macros/src/graphql_macro.rs
swatanabe: We don't necessarily need to insert the original name of the type, though. We only need a sequence of tokens that will resolve to that type.

rust/psibase_macros/src/graphql_macro.rs
tbfleming: Related: it'd be nice for table_query_subindex to support lt, gt, le, ge, but will require the macro user specifying a type for these. Maybe a future enhancement.

swatanabe: Isn't that type `subindex_rest_ty`?

tbfleming:
Attribute macros tend to be nicer than `foo!` macros since they don't introduce as much new syntax. Unfortunately attribute macros appear to be more clumsy for this case.

```rust
[TableQuery(table = "ContentTable", key_fn = "get_index_pk", key_ty = "ContentKeyNamed"]
fn content();

[TableQuerySubindex(table = "ContentTable", key_fn = "get_index_pk", 
 key_ty = "ContentKeyNamed"]
fn content_by_account(account: AccountNumber);
```

rust/psibase_macros/src/graphql_macro.rs
tbfleming: > If the user wants to specify these arguments explicitly, I can't see how to do so using this macro
I meant having the macro automatically add them. Since `subindex_rest_ty` already exists (I missed that), it be extended to do it now.

rust/psibase_macros/src/graphql_macro.rs
```diff
@@ -0,0 +1,195 @@
...
+    key_fn: TypeParam,
+    subindex_params: Vec<TypeParam>,
+}
+
+impl Parse for Args {
+    fn parse(input: ParseStream) -> Result<Self> {
+        let mut vars = Punctuated::<TypeParam, Token![,]>::parse_terminated(input)?.into_iter();
+
+        let query_name = vars.next().expect("Query name not present").ident;
+        let table = vars.next().expect("Table name not present").ident;
```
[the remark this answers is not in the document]
sparkplug0025: Agreed. Do you have any suggestions of how to use `Punctuated::<TypeParam, Token![,]>::parse_terminated` with a polymorphic type? Perhaps an enum example?

sparkplug0025:
>Attribute macros tend to be nicer than foo! macros since they don't introduce as much new syntax. Unfortunately attribute macros appear to be more clumsy for this case.

Yeah I tried to play with it a lot and the `!` was the easiest concise thing that I was able to come up with :/

rust/psibase_macros/src/graphql_macro.rs
sparkplug0025: Doh! That clarifies things and even solves the subindex_rest_ty clunkiness. I did not notice that, thank you!

rust/psibase_macros/src/graphql_macro.rs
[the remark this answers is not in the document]
swatanabe: You don't need to parse everything using Punctuated. You can parse the initial elements separately before constructing `Punctuated::<TypeParam, Token![,]>`.

rust/psibase/src/serve_http.rs
sparkplug0025: nice! but since we need the Error in this case how would you use it inside the else block?

rust/psibase_macros/src/graphql_macro.rs
```diff
@@ -0,0 +1,195 @@
...
+    impl_query.into()
+}
+
+pub fn table_query_macro_impl(item: TokenStream) -> TokenStream {
+    let args = parse_macro_input!(item as Args);
+
+    let query_name = args.query_name;
+    let table = args.table;
+    let record = Ident::new(format!("{}Record", table).as_str(), table.span());
+    let key_fn = args.key_fn.ident;
```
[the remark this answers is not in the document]
sparkplug0025: It won't compile if it isn't, ge, gt, lt, le would have wrong types -- the type is bound by the Table<Key>.

rust/test_service_psispace/src/lib.rs
sparkplug0025: work TBD for a next PR

rust/psibase_macros/src/graphql_macro.rs
```diff
@@ -28,19 +27,44 @@ impl Parse for Args {
...
-        };
+        let mut subindex_params = Vec::new();
+        let mut subindex_remaining_key_ty: Option<TypeParamBound> = None;
+
+        if input.parse::<Token![,]>().is_ok() {
+            loop {
+                if input.is_empty() {
+                    break;
+                } else if input.peek(Token![,]) {
+                    input.parse::<Token![,]>().expect("expected ,");
```
swatanabe, line 38: The `expect` will always pass, because `peek` just checked the same thing.

rust/psibase_macros/src/graphql_macro.rs
```diff
@@ -28,19 +27,44 @@ impl Parse for Args {
...
-            Vec::new()
-        };
+        let mut subindex_params = Vec::new();
+        let mut subindex_remaining_key_ty: Option<TypeParamBound> = None;
+
+        if input.parse::<Token![,]>().is_ok() {
+            loop {
+                if input.is_empty() {
+                    break;
+                } else if input.peek(Token![,]) {
```
swatanabe, line 37: This allows zero or more commas between elements of `subindex_params`. I don't think thats quite what we want.

rust/psibase_macros/src/graphql_macro.rs
sparkplug0025: Good catch, will unwrap it.

sparkplug0025: Actually with my new change it's irrelevant.

rust/psibase_macros/src/graphql_macro.rs
sparkplug0025: Thanks for catching it. Fixed.

rust/psibase_macros/src/graphql_macro.rs
```diff
@@ -0,0 +1,241 @@
...
+                    let param = input
+                        .parse::<TypeParam>()
+                        .expect("unable to parse subindex param type");
+                    subindex_params.push(param);
+                    if input.is_empty() {
+                        break;
+                    }
+                    input.parse::<Token![,]>().expect("expected ,");
+
+                // This is the last param, it should always be the remaining key TypeParamBound
```
swatanabe, line 49: If this needs to be the last param, then it needs to check that it is at the end after an optional trailing comma.

sparkplug0025: Fixed.

swatanabe (approved):

PR #265, opened 2022-12-06
Fracpack review

swatanabe:


libraries/psio/include/psio/fracpack.hpp
```diff
@@ -0,0 +1,2608 @@
...
+      }
+   }
+
+   template <typename T, typename R, typename... Args>
+   constexpr auto tuple_from_function_args(R (T::*func)(Args...))
+       -> std::tuple<std::decay_t<Args>...>;
+
+   template <typename T, typename R, typename... Args>
+   constexpr auto tuple_from_function_args(R (T::*func)(Args...) const)
+       -> std::tuple<std::decay_t<Args>...>;
```
swatanabe, line 2578: Duplicates `tuple_from_args` in reflect.hpp.

libraries/psio/include/psio/fracpack.hpp
```diff
@@ -0,0 +1,2608 @@
...
+   std::vector<char> to_frac(const T& v)
+   {
+      std::vector<char> result(fracpack_size(v));
+      fast_buf_stream   s(result.data(), result.size());
+      fracpack(v, s);
+      return result;
+   }
+
+   template <typename T>
+   T from_frac(const std::vector<char>& b)
```
swatanabe, line 2593: It isn't obvious to me why `from_frac` means checked while `convert_from_frac` means unchecked.

libraries/psio/include/psio/fracpack.hpp
```diff
@@ -0,0 +1,2608 @@
...
+   {
+      if (fracvalidate<T>(b.data(), b.data() + b.size()).valid)
+         return convert_from_frac<T>(b);
+      abort_error(stream_error::invalid_frac_encoding);
+   }
+
+   template <typename T>
+   auto make_view(const char* data, size_t size)
+   {
+      if (fracvalidate<T>(data, size).valid_and_known())
```
swatanabe, line 2603: Why a view need to have all fields `known`?

libraries/psio/include/psio/member_proxy.hpp
```diff
@@ -0,0 +1,84 @@
+#pragma once
```
swatanabe, line 1: We should remove `member_proxy`. It doesn't seem to serve any useful purpose and it creates additional possibility for error:
```
auto x = view->m1()->m2();
// BOOM: x is a member_proxy that holds a reference to the temprorary returned by m1.
```

libraries/psio/include/psio/unaligned_type.hpp
```diff
@@ -0,0 +1,43 @@
...
+
+   /**
+     *  Reading from a char* is potentially unaligned, so we always cast
+     *  through this type to warn the compiler that unaligned_type::val 
+     *  might be unaligned.
+     */
+   // TODO: rename; this is incredibly dangerous; the name should reflect that.
+   // TODO: better: remove this struct and use a template function which
+   //       makes the danger go away:
+   //          T f<T>(const char*)
```
swatanabe, line 16: +1000

libraries/psio/include/psio/get_type_name.hpp
```diff
@@ -0,0 +1,207 @@
...
+namespace psio
+{
+
+#define PSIO_REFLECT_TYPENAME(T)                 \
+   constexpr const char* get_type_name(const T*) \
+   {                                             \
+      return BOOST_PP_STRINGIZE(T);              \
+   }
+
+#define PSIO_REFLECT_TYPENAME_CUSTOM(T, CUSTOM)  \
```
swatanabe, line 23: This can be folded info `PSIO_REFLECT_TYPENAME` by counting arguments.

libraries/psio/include/psio/reflect.hpp
```diff
@@ -0,0 +1,760 @@
...
+   template <typename R, typename C, typename... Args>
+   constexpr std::tuple<std::decay_t<Args>...> args_as_tuple(R (C::*)(Args...) const)
+   {
+      return {};
+   }
+   template <typename R, typename... Args>
+   constexpr std::tuple<std::decay_t<Args>...> args_as_tuple(R (*)(Args...))
+   {
+      return {};
+   }
```
swatanabe, line 44: Should use `remove_cvref_t` instead of `decay_t`.  Array-to-pointer decay is definitely wrong here.

libraries/psio/include/psio/reflect.hpp
```diff
@@ -0,0 +1,760 @@
...
+   {
+      return std::make_tuple(std::get<I>(t)...);
+   }
+
+   template <typename... Ts>
+   constexpr auto prune_tuple(std::tuple<Ts...> t)
+   {
+      return prune_tuple_h(
+          t, std::make_index_sequence<std::tuple_size<std::tuple<Ts...>>::value - 1>{});
+   }
```
swatanabe, line 57: Seems to be unused

libraries/psio/include/psio/reflect.hpp
```diff
@@ -0,0 +1,760 @@
...
+
+   template <typename F>
+   constexpr void tuple_foreach_type(const std::tuple<>* obj, F&& f)
+   {
+   }
+
+   template <typename T, typename... Ts, typename F>
+   constexpr void tuple_foreach_type(const std::tuple<T, Ts...>* obj, F&& f)
+   {
+      f((T*)nullptr);
```
swatanabe, line 104: A fold expression would be simpler.

libraries/psio/include/psio/reflect.hpp
```diff
@@ -0,0 +1,760 @@
...
+   struct is_std_variant : std::false_type
+   {
+   };
+
+   template <typename... T>
+   struct is_std_variant<std::variant<T...>> : std::true_type
+   {
+      static std::string name() { return get_variant_typename<T...>(); }
+      template <typename First, typename... Rest>
+      static std::string get_variant_typename()
```
swatanabe, line 277: This is not used, and the name is also not unique. For example, what type is int&double|char? This is also using `get_type_name` to get the representation of the members, which does something totally different (and equally wrong). We need to consider where the names are being used (and therefore what characters are permitted) and work out a single encoding.

libraries/psio/include/psio/reflect.hpp
```diff
@@ -0,0 +1,760 @@
...
+             std::string("map<") + get_type_name<K>() + "," + get_type_name<V>() + ">";
+         return n;
+      }
+   };
+
+   template <typename T>
+   constexpr bool is_std_map_v = is_std_map<T>::value;
+
+   template <typename T>
+   struct remove_cvref
```
swatanabe, line 333: This is std now.

libraries/psio/include/psio/reflect.hpp
```diff
@@ -0,0 +1,760 @@
...
+      static constexpr int size = sizeof...(Ts);
+   };
+
+   template <typename... Ts>
+   std::tuple<Ts...> TupleFromTypeListImpl(TypeList<Ts...>&&);
+   template <typename T>
+   using TupleFromTypeList = decltype(TupleFromTypeListImpl(std::declval<T>()));
+
+   template <typename T>
+   struct MemberPtrType;
```
swatanabe, line 353: This only seems to be used in psibase/common/include/serveGraphQL.hpp and that only uses it for SimplifiedArgumentTypes which is equivalent to `args_as_tuple`.

libraries/psio/include/psio/reflect.hpp
```diff
@@ -0,0 +1,760 @@
...
+      {
+         f(p, i < names.size() ? names.begin()[i] : nullptr);
+         ++i;
+      };
+      (g((remove_cvref_t<Args>*)nullptr), ...);
+   }
+
+}  // namespace psio
+
+// TODO: not legal to add new definitions to std namespace
```
swatanabe, line 433: Why is it needed in the first place? There's an overload for `string` in `get_type_name.hpp`

libraries/psio/include/psio/reflect.hpp
```diff
@@ -0,0 +1,760 @@
...
+ *    * ident                          non-static data member
+ *    * definitionWillNotChange()      the struct's definition will not change in the future. This saves 2 bytes from the fracpack representation.
+ *    * allowHashedMethods()           allow hashed method identifiers
+ *    * base(ident)                    base class (not implemented)
+ *    * method(ident, arg1, ...)       method
+ *    * numbered(int, ident)           non-static data member with field number
+ */
+#define PSIO_REFLECT(STRUCT, ...)                                                                 \
+   PSIO_REFLECT_TYPENAME(STRUCT)                                                                  \
+   struct reflect_impl_##STRUCT                                                                   \
```
swatanabe, line 680: I would prefix all internal symbol names that are inserted in the user's namespace with `psio_` to avoid conflicts.

libraries/psio/include/psio/reflect.hpp
```diff
@@ -0,0 +1,760 @@
...
+   psio::member_proxy<i, psio::hash_name(BOOST_PP_STRINGIZE(PSIO_GET_IDENT(elem))),        \
+                                         &STRUCT::PSIO_GET_IDENT(elem), const ProxyObject> \
+                      PSIO_GET_IDENT(elem)() const                                         \
+   {                                                                                       \
+      return _psio_proxy_obj;                                                              \
+   }
+
+#define PSIO_PROXY_METHOD(r, STRUCT, i, elem)                                           \
+   template <typename... Args>                                                          \
+   auto PSIO_GET_IDENT(elem)(Args... args)                                              \
```
swatanabe, line 660: The return type should be `decltype(auto)`

swatanabe (commented):
Lots of dead files that should be removed

PR #274, opened 2022-12-13
updated dev environment sample files

James-Mart:


.vscode/psibase.code-snippets.sample
```diff
@@ -0,0 +1,42 @@
...
+			"PSIO_REFLECT($1, account);",
+			"using $1Table = Table<$1, &$1::account>;"
+		],
+		"description": "Create psibase service table skeleton"
+	},
+
+	"Create store/serve actions": {
+		"prefix": "psi-storeserve",
+		"body": [
+		  "auto serveSys(HttpRequest request) -> optional<HttpReply>",
```
swatanabe, line 19: If this is inlined in the class definition, it may be in a header, in which case `using namespace std` and `using namespace psibase` should not be available.

Issue #239, opened 2022-11-28
Audit boot size


tbfleming:
The boot block is approaching the ~16MB triedent record size limit; it exceeds it when `--no-doc` isn't used.

swatanabe:
It's now going over:
```
#8  0x0000000001b38eb9 in triedent::write_session::upsert (
    this=0x60c000000950, r=std::shared_ptr<triedent::root> (empty) = {...}, 
    key=std::span of length 4 = {...}, 
    val=std::span of length 16899736 = {...})
```

swatanabe:
I found one culprit: d28f5fa7f12708605268b694d92d974847599c66 

The actual command executed is:
```
[19/21] cd .../rust-build && rm -rf ".../psibase/rust/psibase/boot-image/*" && mkdir ...
```
The double quotes added by cmake suppress glob expansion, which completely stops rm from doing anything.

ExploreSys generates unique file names on every build:
```
-rw-r--r--. 1 swatanabe swatanabe 24981 Dec  1 12:05 rust/psibase/boot-image/ExploreSys/ui/dist/dist/_app/immutable/start-686196bc.js
-rw-r--r--. 1 swatanabe swatanabe 24981 Dec  1 13:12 rust/psibase/boot-image/ExploreSys/ui/dist/dist/_app/immutable/start-4641986e.js
-rw-r--r--. 1 swatanabe swatanabe 24981 Dec  2 11:54 rust/psibase/boot-image/ExploreSys/ui/dist/dist/_app/immutable/start-0d06e378.js
-rw-r--r--. 1 swatanabe swatanabe 24981 Dec  2 11:59 rust/psibase/boot-image/ExploreSys/ui/dist/dist/_app/immutable/start-a7adb66b.js
-rw-r--r--. 1 swatanabe swatanabe 24981 Dec  2 12:06 rust/psibase/boot-image/ExploreSys/ui/dist/dist/_app/immutable/start-8350f546.js
-rw-r--r--. 1 swatanabe swatanabe 24981 Dec  2 12:17 rust/psibase/boot-image/ExploreSys/ui/dist/dist/_app/immutable/start-bc015141.js
-rw-r--r--. 1 swatanabe swatanabe 24981 Dec  2 12:23 rust/psibase/boot-image/ExploreSys/ui/dist/dist/_app/immutable/start-caec9844.js
[...]

PR #276, opened 2022-12-13
add new subdirectory to boot image to avoid glob expansion issue

James-Mart:
Adds a new directory inside boot-image that can be entirely deleted & recreated by cmake. This avoids [this glob expansion issue](https://github.com/gofractally/psibase/issues/239#issuecomment-1343528947) and still allows `boot-image/` to be bind-mounted outside of a docker container.

After this update, devs should run `rm -rf /root/psibase/rust/psibase/boot-image/*` inside their environments to manually clear out the boot-image.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -131,7 +131,7 @@ endif()
 
 option(BUILD_RUST "Build rust" ON)
 if(BUILD_RUST)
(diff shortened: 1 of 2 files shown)


tbfleming:
Why not bind-mount higher up the tree?

James-Mart:
> Why not bind-mount higher up the tree?

I don't feel strongly about it, but I did it to keep the bind mount smaller / more focused. Didn't seem useful to expose src/ or tests/ to the host.

PR #278, opened 2022-12-16
Initial performance monitoring

swatanabe:
Only works on linux


swatanabe:
@sparkplug0025, I think this should be sufficient to get started. Let me now if there are other statistics that should be collected.

PR #302, opened 2023-01-30
Metrics

swatanabe:
Add a `/native/admin/metrics` endpoint. Some  and labels have been renamed compared to the json exporter.

I also restored the original access control on this and `/native/admin/perf`, because relaxing it wasn't enough to solve the problem. Instead, the `service` option now allows an empty path, which will cause it to serve no content besides the native interface.


sparkplug0025:
I'm receiving this error in Prometheus:
image

I believe it's because you need to rename `psinode_memory` to `psinode_memory_bytes`. 

According to their doc [here](https://prometheus.io/docs/practices/naming/#metric-names):
```
...should have a suffix describing the unit, in plural form. Note that an accumulating count has total as a suffix, in addition to the unit if applicable.
http_request_duration_seconds
node_memory_usage_bytes
http_requests_total (for a unit-less accumulating count)
process_cpu_seconds_total (for an accumulating count with unit)
foobar_build_info (for a pseudo-metric that provides [metadata](https://www.robustperception.io/exposing-the-software-version-to-prometheus) about the running binary)
data_pipeline_last_record_processed_timestamp_seconds (for a timestamp that tracks the time of the latest record processed in a data processing pipeline)
```

Is there a way that you can spin up prometheus via our docker instructions in my PR and test it to make sure the other labels does not error out too?

This is my prometheus config to add the new metrics endpoint:
```
  - job_name: "psinode_perf2"
    scrape_interval: 1s
    metrics_path: /native/admin/metrics
    static_configs:
      - targets: ["host.docker.internal:8080"] # change to localhost if not running in docker
```

swatanabe:
> I believe it's because you need to rename `psinode_memory` to `psinode_memory_bytes`.
> 
> According to their doc [here](https://prometheus.io/docs/practices/naming/#metric-names):
> 

That is a convention, not a requirement (note the work "should"). The spec I'm using is here: https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md

"Unit specifies MetricFamily units. If non-empty, it MUST be a suffix of the MetricFamily name separated by an underscore."

sparkplug0025 (approved):

PR #303, opened 2023-01-31
New sign trx flow

sparkplug0025:


Velua:
Converting back to a draft, there seems to be a bug where a transaction is being triggered with no actions, there may be a race condition where the call back system is resolving the wrong promise, still investigating.

Velua:
@sparkplug0025 lets get on a call whenever it suits you.

services/system/AccountSys/ui/src/appInit.ts
```diff
@@ -59,69 +207,78 @@ export const initAppFn = (setAppInitialized: () => void) =>
...
+                    return keystore.getAccounts();
+                },
+            },
+            {
+                id: "getProof",
+                exec: async (params: GetProofParams): Promise<any> => {
+                    console.info("getproof received!!", params);
+                    const proof = keystore.getProof(params);
+                    console.info("generated proof", proof);
+                    return { proof };
```
brandonfancher, line 275: Just a reminder to remove any console statements if they're no longer necessary.

services/user/InviteSys/ui/README.md
```diff
@@ -0,0 +1,68 @@
+# Fractally Webapp
```
brandonfancher, line 1: This isn't the Fractally Webapp, but rather InviteSys, right? Maybe I hadn't updated this when the app was in toolbox.

brandonfancher (approved):
Just a few comments. It'll be nice when we can point these to a shared components lib in psibase instead of bringing in the components directly. But we'll work towards thisl.

Issue #292, opened 2023-01-25
Watch main applet history changes


sparkplug0025:
Parent frame = Psibase main page with the whole common sys and applets infra
Main Applet = current active applet

- Overrides the parent frame address with the address from the main applet
- Watch the changes on the window.history of the main applet and replicate it to the parent frame
- The parent frame address needs to append the path from the main applet
  - If the Main applet address is `some-applet.psibase.io/foo/bar?favorite_num=123`
  - Parent frame should be `psibase.io/some-applet/foo/bar?favorite_num=123`
- The back button of the browser should work with these address changes

PR #312, opened 2023-02-22
History changes synced with main frame address

sparkplug0025:
Implements #292

### DIFF
--- a/services/user/CommonSys/common/rpc.mjs
+++ b/services/user/CommonSys/common/rpc.mjs
@@ -341,6 +341,7 @@ export const MessageTypes = {
     OperationResponse: "OperationResponse",
     TransactionReceipt: "TransactionReceipt",
     SetActiveAccount: "SetActiveAccount",
+    ChangeHistory: "ChangeHistory", // Fired when the applet history changes
 };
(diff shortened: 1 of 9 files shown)


services/user/CommonSys/common/rpc.mjs
```diff
@@ -627,9 +632,45 @@ export async function initializeApplet(initializer = () => { }) {
...
     await initializer();
 }
 
+const sendHistoryChange = () => {
+    const { pathname, href, search } = document.location;
+    console.info("sending change history message", {
+        href,
+        pathname,
+        search,
+    });
```
Velua, line 667: We happy to lose this or shall keep it for now?

Velua, line 667:
```suggestion
```

Velua (approved):

services/user/CommonSys/common/rpc.mjs
sparkplug0025: I think for debug purposes it's a good idea to keep it for now! then if we dont see any anomalies we can go ahead and clean it up!

PR #319, opened 2023-02-28
Add shared components / fractally applet

Velua:


sparkplug0025 (approved):
Components seems to be working from my basic navigation.

Please cleanup the `/service/user` root folder prior to merging as we talked about.

PR #324, opened 2023-03-07
Avoid race condition

Velua:
Me and @sparkplug0025 discovered an issue where when the app saves the token upon start up it can run too closely to when `account-sys` mutates local storage. This update changes the saving / mutating of the local key storage to when the user triggers an `accept` or `acceptCreate` action instead.


sparkplug0025 (approved):
lgtm. I will be creating an issue for us to review this issue. Like we talked about yesterday, ideally the applet should know how to handle and avoid race conditions before exposing its operations.

PR #326, opened 2023-03-08
Implements message metadata and add getproof check

### LINKED ISSUE #321: Restrict AccountSys.getProof calls
getProof needs to verify if the sender is common-sys and if not reject the operation.

As an alternative, common-sys layer can ban any `getProof` calls.

sparkplug0025:
Fix #321


Velua (approved):
Sweet!

PR #327, opened 2023-03-08
add fractal creation

Velua:


James-Mart:
I guess we should close this

PR #338, opened 2023-03-13
Typescriptify common RPC

sparkplug0025:
- Implement the `rpc.mjs` in typescript
- Also add the `keyConversions.mjs` dependency
- Imports it in psispace (use it as a reference to import it into other applets)


Velua (approved):
Awesome! 

I'd not bother with -lib appended to the end, but that's just preference. Super excited!

PR #343, opened 2023-03-14
Add https documentation

brandonfancher:


doc/https.md
```diff
@@ -0,0 +1,70 @@
+# HTTPS
+
+Some web technologies require a secure context when not operating over `localhost`. This includes things link clipboard access and WebRTC--two technologies used within the `FractalSys` applet. For this reason, we need a means of running psinode with an SSL.
```
James-Mart, line 3: things like* clipboard access

doc/https.md
```diff
@@ -0,0 +1,70 @@
...
+
+## Generating certs on local machine
+
+1. Install `mkcert` locally
```
James-Mart, line 7: What's "local," here? Is it within the container? Or on your host machine and then you copy them into the docker container if that's what you're using?

services/user/FractalSys/ui/vite.config.ts
```diff
@@ -52,11 +52,11 @@ const psibase = (options: Options) => {
             ...httpServerConfig,
             https: {
                 key: fs.readFileSync(
-                    "./psibase.127.0.0.1.sslip.io+1-key.pem",
+                    "../../../../psibase.127.0.0.1.sslip.io+1-key.pem",
```
James-Mart, line 55: Can the vite config read environment variables? Might be nice to set the path to the location of the cert and key as an environment variable and not have to locally touch the vite config.

doc/https.md
brandonfancher: host machine. I'll clarify.

services/user/FractalSys/ui/vite.config.ts
brandonfancher: Yeah, good call.

James-Mart (approved):

PR #378, opened 2023-04-04
decouple doc-sys account creation from doc deployment

James-Mart:


swatanabe (approved):
I think we'd be better off making the list of accounts local and appending the account names as they are created.

PR #388, opened 2023-04-14
Rename String to Memo and make it actually work

swatanabe:
- Add hook clio_validate_packable which applies to wrappers
- Add psio::view for std::string
- Enable view for wrappers that are also reflected when reflection appears to match clio_unwrap_packable.
- Serialization of Memo in both JSON and fracpack is identical to the serialization of std::string
- The requirements are checked on all the normal methods of creating a Memo (constructors and deserialization), and can be considered a class invariant.


swatanabe:
- [ ] Fix rust so psibase boot works

libraries/psibase/common/include/psibase/Memo.hpp
```diff
@@ -0,0 +1,63 @@
...
+      from_json(s.contents, stream);
+      Memo::validate(s.contents);
+   }
+
+   void to_json(const Memo& s, auto& stream)
+   {
+      to_json(s.contents, stream);
+   }
+
+   inline std::string& clio_unwrap_packable(Memo& s)
```
James-Mart, line 44: clio is psio now

James-Mart: Oh, I just saw it was still `clio` in PackableWrapper

James-Mart (approved):

PR #398, opened 2023-05-02
Remove print.hpp.

swatanabe:
Use `printf` instead
- It actually works
- It's far more generic
- We don't need to document it because it's part of the standard library


swatanabe:
There's still a corresponding print.cpp

James-Mart (approved):

James-Mart:
> There's still a corresponding print.cpp

That should be deleted?

swatanabe:
> > There's still a corresponding print.cpp
> 
> That should be deleted?

yes

Issue #374, opened 2023-03-31
Sync AdminSys with config API


swatanabe:
- port has been replaced with listen, which requires a list control of some sort
- The TLS options need to be added

James-Mart:
Added bug label because this expresses itself (at least in the case of port) with a value that appears incorrect compared to the cli params

James-Mart:
Closing in favor of https://github.com/gofractally/psibase/issues/742

PR #402, opened 2023-05-05
Change AdminSys to use "listen" instead of "port". Refs #374.

swatanabe:


services/user/AdminSys/ui/src/configuration/utils.ts
```diff
@@ -201,6 +201,57 @@ export function writeLoggers(loggers: { [index: string]: LogConfig }): {
...
+    if (listen.protocol == "http" || listen.protocol == "https") {
+        const ipv4 = listen.text?.match(/^(\d+\.\d+.\d+.\d+)(?::(\d+))$/);
+        if (ipv4) {
+            return {
+                protocol: listen.protocol,
+                address: ipv4[1],
+                port: Number(ipv4[2]),
+            };
+        }
+        const ipv6 = listen.text?.match(/^\[(.*)\](?::(\d+))$/);
```
James-Mart, line 236: The .* may be over-permissive

James-Mart (approved):

services/user/AdminSys/ui/src/configuration/utils.ts
swatanabe: It doesn't really matter here. This isn't being used for validation and it sufficient to distinguish the form of any valid inputs.

PR #425, opened 2023-05-23
Misc cleanup

James-Mart:


services/user/CommonSys/CMakeLists.txt
```diff
@@ -4,15 +4,4 @@ endfunction(add)
 
 include(ExternalProject)
 
-downloadThirdParty("htm" "https://unpkg.com/htm@3.1.0/dist/htm.module.js")
```
swatanabe, line 7: The build is failing because the root CMake is still trying to copy these.

James-Mart: Actually I'll do this cleanup in a different pr

swatanabe (approved):

PR #426, opened 2023-05-23
add gdbinit commands relative to workspace

James-Mart:
In .gdbinit I had `add-auto-load-safe-path /root/psibase/build/psidk/lib/debug/psitest-gdb.py`
But here I changed it to `add-auto-load-safe-path /root/psibase/build/psidk/lib/debug/`, because it's supposed to add a safe dir.

### DIFF
--- a/.vscode/launch.json.sample
+++ b/.vscode/launch.json.sample
@@ -30,11 +30,11 @@
                 },
                 {
                     "description": "Add directory from which to load auto-loaded gdb scripts",
-                    "text": "add-auto-load-scripts-directory ${workspaceFolder}/build/psidk/lib/debug/",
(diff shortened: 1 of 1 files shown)


.vscode/launch.json.sample
```diff
@@ -27,6 +27,14 @@
                 {
                     "description": "Enable pretty-printing for gdb",
                     "text": "-enable-pretty-printing",
+                },
+                {
+                    "description": "Add directory from which to load auto-loaded gdb scripts",
+                    "text": "add-auto-load-scripts-directory ${workspaceFolder}/build/psidk/lib/debug/",
```
swatanabe, line 33: I changed this in #415. It's not `.../psidk/share/gdb/auto-load`

James-Mart:
- [ ] Depends on #415

PR #429, opened 2023-05-31
Add simd to the default cmake toolchain. Fixes #383.

### LINKED ISSUE #383: Turn on SIMD by default in cmake config

swatanabe:
Also disable clang's automatic invocation of wasm-opt, because it causes too many problems.


swatanabe:
Note: My build randomly broke because the current version of `wasm-opt` turns on sign-ext by default.

James-Mart:
So just to be clear, wasm-opt is still running, just not through the automatic invocation?

swatanabe:
> So just to be clear, wasm-opt is still running, just not through the automatic invocation?

It's not running at all. I replaced it with a no-opt script, because I couldn't figure out a way to call it that would work reliably.

James-Mart:
> > So just to be clear, wasm-opt is still running, just not through the automatic invocation?
> 
> It's not running at all. I replaced it with a no-opt script, because I couldn't figure out a way to call it that would work reliably.

Are the wasm sizes still comparable?

swatanabe:
The difference is a lot more than I expected:
```
[swatanabe@localhost psibase]$ ls -l TransactionSys.wasm 
-rwxr-xr-x. 1 swatanabe swatanabe 167857 Jun  1 08:03 TransactionSys.wasm
```
```
[swatanabe@localhost psibase]$ ls -l TransactionSys.wasm 
-rw-r--r--. 1 swatanabe swatanabe 116186 Jun  1 09:13 TransactionSys.wasm
```

swatanabe:
Okay, it finally works on 20.04

James-Mart (approved):

PR #472, opened 2023-07-28
Implement multi-hop routing of consensus messages

swatanabe:


doc/psidk/src/consensus.md
```diff
@@ -57,24 +69,53 @@ BFT consensus has the following properties:
...
+Note: Since the condition for activating a view requires the leader's view change to refer to a block which not older than other nodes' view changes, this ordering guarantees that as long as a valid set of view changes exist, all connected, honest nodes will converge to the same valid set.
+
+### Safety
+
+Let block X and block Y be two conflicting blocks which are both irreversible
+- Without loss of generality, let X be better than Y
+- Y has been committed by 2f+1 producers, because that is the requirement for it to be considered irreversible
+- let Z be the earliest block that is an ancestor of X (inclusive of X), conflicts with Y, is better than Y, and is prepared by 2f+1 producers. Such a block is guaranteed to exist because X itself must be prepared by 2f+1 producers in order to be committed by any honest producer.
+
+The existence of Z implies that at least f+1 producers have violated the protocol
```
James-Mart, line 111: Doesn't the existence of X and Y already imply that at least f+1 producers have violated the protocol?

swatanabe: It does. This list is the breakdown of exactly how they violated the protocol.

libraries/net/include/psibase/shortest_path_routing.hpp
```diff
@@ -0,0 +1,676 @@
...
+            multicast(RouteUpdateMessage{self, seqno, RouteMetric{0}});
+         }
+         // Sync our routing table with all peers
+         multicast(RequestRoutesMessage{});
+      }
+      bool isFeasible(std::pair<const RouteKey, RouteData>& route)
+      {
+         if (route.second.metric == RouteMetric::infinite)
+         {
+            return true;
```
James-Mart, line 350: Shouldn't infinite metric return !feasible?

libraries/net/test/test_shortest_path_routing.cpp
James-Mart, line 1: Based on the name, I expected this to test that the routing would follow the shortest path, but the tests really seem only to verify that multi-hop routing works (and will only route messages to producers).

libraries/net/include/psibase/shortest_path_routing.hpp
swatanabe: https://datatracker.ietf.org/doc/html/rfc8966#feasibility-condition
"Note further that retractions (updates with infinite metric) are always feasible, since they cannot possibly cause a routing loop"

James-Mart (approved):

libraries/net/test/test_shortest_path_routing.cpp
swatanabe: I'm not exactly sure how I would test that. The exact path followed is kind of an implementation detail.

PR #475, opened 2023-08-15
Add version information for psinode and psitest

James-Mart:
image


libraries/psibase/common/include/psibase/version.hpp
```diff
@@ -0,0 +1,5 @@
+#pragma once
+
+#define PSINODE_VERSION_MAJOR 0
```
swatanabe, line 3: Since this is a single project-wide version number, it should probably be called `PSIBASE` rather than `PSINODE`

swatanabe (approved):

PR #489, opened 2023-09-14
Improvements to the infrastructure provider admin GUI 

### LINKED ISSUE #488: Subtask: Add boot button to admin-sys
See parent task #487 for a description of the admin-sys boot button.

James-Mart:
Closes #488


rust/psibase/src/boot.rs
```diff
@@ -537,3 +500,30 @@ pub fn create_boot_transactions(
...
+        .map_err(|e| JsValue::from_str(&e.to_string()))?;
+
+    let (boot_transactions, transactions) =
+        create_boot_transactions(&None, prod.into(), true, true, true, expiration);
+
+    let boot_transactions = boot_transactions.packed();
+    let transactions : Vec<Vec<u8>> = transactions
+        .into_iter()
+        .map(|tx| tx.packed())
+        .collect();
```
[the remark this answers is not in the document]
James-Mart: It worked :+1:

James-Mart:
Build error:
```
2023-09-16T17:16:04.3175153Z Scanning dependencies of target wasm-psibase
2023-09-16T17:16:07.0522020Z Error: Permission denied (os error 13)
2023-09-16T17:16:07.0522439Z Caused by: Permission denied (os error 13)
2023-09-16T17:16:07.0532465Z make[2]: *** [CMakeFiles/wasm-psibase.dir/build.make:57: CMakeFiles/wasm-psibase] Error 1
2023-09-16T17:16:07.0540583Z make[1]: *** [CMakeFiles/Makefile2:688: CMakeFiles/wasm-psibase.dir/all] Error 2
2023-09-16T17:16:07.0587715Z make: *** [Makefile:152: all] Error 2
2023-09-16T17:16:12.7223917Z ##[error]Process completed with exit code 2.
```

James-Mart:
Discovered the bug in wasm-pack that was preventing me from using the non-root docker image.
https://github.com/rustwasm/wasm-pack/issues/1335

For now, we can work around the bug by passing the `WASM_PACK_CACHE` environment variable to the build container, and then it creates the cache in a permitted directory.

CMakeLists.txt
```diff
@@ -150,44 +150,53 @@ set(JS_DIRS
...
+        DEPENDS ${CMAKE_CURRENT_SOURCE_DIR}/${PATH}/dist/${OUTPUT_FILE}
+    )
+endfunction()
 
 option(BUILD_JS "Build JS" ON)
 if(BUILD_JS)
     list(TRANSFORM JS_DIRS REPLACE "^[^:]+:([^:]+)$" \\1 OUTPUT_VARIABLE JS_DEPS)
     file(GLOB common LIST_DIRECTORIES false ${CMAKE_CURRENT_SOURCE_DIR}/services/user/CommonSys/common/*)
     file(GLOB common-fonts LIST_DIRECTORIES false ${CMAKE_CURRENT_SOURCE_DIR}/services/user/CommonSys/common/fonts/*)
+    file(GLOB admin-wasm LIST_DIRECTORIES false ${CMAKE_CURRENT_SOURCE_DIR}/services/user/AdminSys/ui/wasm-psibase/*)
```
swatanabe, line 180: This GLOB doesn't work correctly, because the files it's looking for are created by the build, which happens after cmake runs.

package-lock.json
swatanabe, line 1: What is this file for?

James-Mart: Added by accident. Good catch

CMakeLists.txt
James-Mart: I just deleted it because I think the dependency I was intending to add is already covered by `add_dependencies(AdminSys_js wasm-psibase)`

swatanabe: I don't think it quite works.
- `wasm-pack` should run unconditionally because it manages its own dependencies ✅
- AdminSys_js should be built iff any of its sources were modified OR `wasm-pack` updated its outputs. As it's currently set up, there is nothing that detects whether `wasm-pack` did anything.  I believe the correct way to handle this is to use `BYPRODUCTS` in `add_custom_target`

James-Mart: Added `BYPRODUCTS` and the corresponding dependencies

CMakeLists.txt
```diff
@@ -161,33 +179,33 @@ if(BUILD_JS)
...
+    add_js_target(${ADMIN_DIR} index.html
+        ${byproducts}/wasm-psibase_bg.js
+        ${byproducts}/wasm-psibase_bg.wasm
+        ${byproducts}/wasm-psibase_bg.wasm.d.ts
+        ${byproducts}/wasm-psibase.d.ts
+        ${byproducts}/wasm-psibase.js
+    )
+
+    add_custom_target(wasm-psibase ALL
+    COMMAND wasm-pack build ${CMAKE_CURRENT_SOURCE_DIR}/rust/psibase --out-dir ${byproducts} --out-name wasm-psibase
```
swatanabe, line 199: I'm seeing intermediate build products created in the source tree rather than in the build tree. I'm not sure whether this is fixable, because wasm-pack seems to hard-code the path.

CMakeLists.txt
```diff
@@ -161,33 +179,33 @@ if(BUILD_JS)
...
+
+    add_js_target(${ADMIN_DIR} index.html
+        ${byproducts}/wasm-psibase_bg.js
+        ${byproducts}/wasm-psibase_bg.wasm
+        ${byproducts}/wasm-psibase_bg.wasm.d.ts
+        ${byproducts}/wasm-psibase.d.ts
+        ${byproducts}/wasm-psibase.js
+    )
+
+    add_custom_target(wasm-psibase ALL
```
swatanabe, line 198: `wasm-pack` always updates the timestamp on its output files. I had assumed that it was checking the source timestamps because it uses cargo, but that only applies to the wasm build step. Other steps run unconditionally. This causes AdminSys to rebuild every time, which is what I was hoping to avoid.

These infinite rebuilds can be fixed by moving the `wasm-pack` command into an `add_custom_command` associated with the `wasm-psibase` target. In order to ensure that it rebuilds when the sources are modified, add `BUILD_BYPRODUCTS` pointing at the `psibase` executable to the `rust` external project and make the custom command depend on `psibase`. This is a somewhat hacky, as it actually depends on the *sources* of `psibase`, but it should work well enough in practice.

James-Mart: This setup still seems to build `wasm-psibase` on every build, presumably because `rust` is also building on every build.
It's not building everything, but it's doing something... I see this on every build, no matter if I change sources:
```
[  9%] Performing build step for 'rust'
   Compiling psibase v0.2.0 (/root/psibase/rust/psibase)
    Finished release [optimized] target(s) in 34.57s
[  9%] No install step for 'rust'
[  9%] Completed 'rust'
[ 10%] Built target rust
[ 10%] Generating ../services/user/AdminSys/ui/wasm-psibase/wasm-psibase_bg.js, ../services/user/AdminSys/ui/wasm-psibase/wasm-psibase_bg.wasm, ../services/user/AdminSys/ui/wasm-psibase/wasm-psibase_bg.wasm.d.ts, ../services/user/AdminSys/ui/wasm-psibase/wasm-psibase.d.ts, ../services/user/AdminSys/ui/wasm-psibase/wasm-psibase.js
[INFO]: 🎯  Checking for the Wasm target...
[INFO]: 🌀  Compiling to Wasm...
   Compiling psibase v0.2.0 (/root/psibase/rust/psibase)
    Finished release [optimized] target(s) in 22.57s
```

swatanabe:
The test failures on the last commit look like a bug in the test. It's trying to use the `/graphql` endpoint immediately after `psibase boot` completes. The service HTTP API isn't available until a block is produced, so it returns a 404. Since the graphql endpoint is being used to track block production, we have a bit of a chicken and egg problem here. I think it will just need to catch the 404 and treat it as meaning "I should wait some more, because it's not ready yet."

swatanabe (approved):

PR #491, opened 2023-09-22
Support using P-256 keys for block and transaction signing

swatanabe:
Motivation
===

The NIST curves are much more widely supported than secp256k1. Adding them will give us better compatibility with third party software and hardware.

Data formats
===
- The signature is encoded in compact form (64 bytes, r+s, big endian)
- The public key is encoded as an X.509 Subject Public Key Info
- The private key is encoded using `PKCS #8`

Verify Service
===

The VerifySys service (`verify-sys`) uses Botan.  Botan supports picking out only the required components much better than openssl does and compiles to WASM with only minimal tweaks. The main patch I needed to make was to add a stub exception library that simply aborts when an exception is thrown. VerifySys supports both named curves and arbitrary curves with explicit domain parameters. However, only 256-bit curves are useful right now, because SHA256 is applied by native.

Block Signing
===
Native block signing is supported using OpenSSL. Any private key supported by OpenSSL can be used to sign, however key generation is not configurable and always generates an r1 key. As with VerifySys, the use of SHA256 is hard-coded.

psibase command line support
===
All key command line arguments can be either base58 encoded keys as before or the name of a file containing a PEM or DER encoded key. A base58 key uses verifyec-sys/auth-ec-sys. ASN.1 encodings use verify-sys/auth-sys. Only k1 keys are supported right now. The public key must be in compressed form and must use the named curve representation of the parameters.

js frontend support
===
Keys can be either PEM or Base58 encoded. When creating an account, account-sys always uses auth-sys. When signing or importing keys, the auth service will be deduced by querying the chain.


James-Mart:
Am I understanding correctly that verify-sys can be used to verify nist and k1 curves?

swatanabe:
> 
> Am I understanding correctly that verify-sys can be used to verify nist and k1 curves?

Yes. Also brainpool and other more obscure curves.

James-Mart:
> Yes. Also brainpool and other more obscure curves

Any reason not to switch everything to this and delete verify-ec-sys?

swatanabe:
> Any reason not to switch everything to this and delete verify-ec-sys?

I haven't checked performance. secp256k1 is more optimized.

programs/psinode/main.cpp
```diff
@@ -362,14 +363,23 @@ namespace psibase
...
+         std::shared_ptr<Prover> result;
+
+         if (key.service.str() == "verifyec-sys")
+         {
+            result = std::make_shared<EcdsaSecp256K1Sha256Prover>(
+                key.service, psio::from_frac<PrivateKey>(key.rawData));
+         }
+         else if (key.service.str() == "verify-sys")
+         {
+            result = std::make_shared<OpenSSLProver>(key.service, key.rawData);
```
James-Mart, line 376: Shouldn't the key data be fracpacked?

rust/psibase/src/boot.rs
```diff
@@ -52,6 +52,7 @@ const ACCOUNTS: [AccountNumber; 27] = [
     account!("token-sys"),
     transaction_sys::SERVICE,
     account!("verifyec-sys"),
+    account!("verify-sys"),
```
James-Mart, line 55: The names are a bit misleading now, since nist curves are ec, but can only be verified with verify-sys. Maybe rename verifyec-sys to verifyk1-sys?

programs/psinode/main.cpp
swatanabe: Why? It's already in a binary format.

James-Mart: Just noticed that the k1 key is expected to be packed, why is it different?

swatanabe: The k1 key is actually a `variant<k1_key, r1_key>`, so it needs additional encoding. The BER encoded key already contains a discriminator to identify the type of key. I don't see any reason to add another layer of wrapping.

swatanabe:
The current performance numbers don't look good. I'm going to check how much of that is global initialization.

|ns| verify-sys | verifyec-sys |
|---|---|---|
||43013901|733819|
||42598396|2903760|
||45292236|778158|
||66910510|2885490|
||42495448|1679233|

swatanabe:
Tests are failing because binaryen 105 (from ubuntu 22.04) doesn't fully support v128.

swatanabe:
> Can't we just change the 22.04 image to use the version of binaryen currently installed in 20.04?

We could, but I'd rather minimize our dependence on non-system packages. I didn't see any significant performance difference for this service.  We can reconsider minimum versions in six months when 24.04 ships.

James-Mart (approved):

PR #493, opened 2023-09-28
Simplify psibase CLI - Remove no-doc option

James-Mart:
- [ ] #489

Since the docs are now added to psinode by default, I also took some time to clean up how they work. I made doc-sys into an applet, and improved the synchronization between URL changes within the iframe and the displayed URL in the web browser to better conform to user expectations.

I also added the docs to the "home page" for now, and removed the "about" link which was broken on local deployments anyway.


James-Mart:
> The default theme I'm seeing with the docs in dark mode but a white background for the common navbar is a bit jarring.

Definitely agree. I think nav bar should be generally themeable, and default dark mode.

PR #498, opened 2023-10-05
eliminate unneeded third-party dependencies

James-Mart:
Some of the older proof-of-concept UIs made use of third-party resources we stores on chain, such as react. 
But all the newer apps just bundle dependencies as usual. We can delete these third-party resources.


swatanabe:
I think #502 should fix the test failures here.

PR #505, opened 2023-10-31
Switch keyring to use PKCS #11

swatanabe:
Example usage:
```sh
./psinode ./db -l 8080 --pkcs11-module p11-kit-proxy.so
```

The HTTP API can list and login to tokens:
```sh
$ curl http://localhost:8080/native/admin/keys/devices | jq
[
  {
    "name": "psibase keyring",
    "id": "pkcs11:token=psibase%20keyring;manufacturer=SoftHSM%20project;model=SoftHSM%20v2;serial=9faefe5e5a9b49c2",
    "unlocked": true
  },
  {
    "name": "",
    "id": "pkcs11:token=;manufacturer=SoftHSM%20project;model=SoftHSM%20v2;serial=",
    "unlocked": false
  }
]
```

```sh
$ curl -d '{"device":"pkcs11:serial=9faefe5e5a9b49c2","pin":"password"}' -H 'Content-Type: application/json' http://localhost:8080/native/admin/keys/unlock
```

the token must be logged in before its keys are accessible

I tried using a Yubkey 5 NFC with `libykcs11.so`. It didn't work properly, because key generation requires some attributes to be set to yubikey specific values. It will probably work if the keys are set up separately, but I'm going to defer that to a future PR. Currently, `psinode` will only use keys that were created by `psinode`.

The id is a PKCS #11 URI specified by [RFC 7512](https://datatracker.ietf.org/doc/html/rfc7512)


doc/psidk/src/run-infrastructure/cli/psinode.md
```diff
@@ -40,6 +40,10 @@ psinode - The psibase blockchain server
 
   Use this private key to sign blocks. Any number of keys may be provided, but only the one that matches the public key corresponding to the producer name will be used. If the correct key is not provided, then `psinode` will be unable to produce blocks.
 
+- `--pkcs11-module` *filename*
```
James-Mart, line 43: I assume I can use multiple of these flags to add multiple modules?

programs/psinode/main.cpp
```diff
@@ -1751,96 +1904,113 @@ void run(const std::string&              db_path,
...
-              &tls_key, &root_ca, callback = std::move(callback)]() mutable
+              &tls_key, &root_ca, &pkcs11_modules, callback = std::move(callback)]() mutable
              {
                 PsinodeConfig result;
                 result.p2p                                       = http_config->enable_p2p;
                 std::tie(result.peers, result.autoconnect.value) = node.autoconnect();
                 result.producer                                  = node.producer_name();
-                result.host                                      = host;
-                result.listen                                    = http_config->listen;
+                result.pkcs11_modules = pkcs11_modules, result.host = host;
```
James-Mart, line 2012: formatting ^

programs/psinode/main.cpp
```diff
@@ -1920,12 +2091,36 @@ void run(const std::string&              db_path,
                          result = std::make_shared<OpenSSLProver>(key.service);
                       }
                    }
-                   std::vector<Claim> existing;
-                   prover->get(existing);
+                   else
```
James-Mart, line 2094: Looks like this else statement assumes there's a key device, but `key.device.empty()` could still be empty

James-Mart, line 2094: Should this explicitly require key.device to be nonempty?

doc/psidk/src/run-infrastructure/cli/psinode.md
swatanabe: Yes

doc/psidk/src/run-infrastructure/administration.md
```diff
@@ -91,6 +94,28 @@ Each peer has the following fields:
...
+|------------|---------|------------------------------------------------------|
+| `name`     | String  | The human readable name of the device                |
+| `id`       | String  | An opaque string that uniquely identifies the device |
+| `unlocked` | Boolean | `true` if the device is currently unlocked           |
+
+`POST` to `/native/admin/keys/unlock` unlocks a cryptographic device.
+
+| Field    | Type   | Description                 |
+|----------|--------|-----------------------------|
+| `device` | String | The device id.              |
```
James-Mart, line 111: It would be nice if the device could be unlocked with the device name/label

swatanabe: This is machine interface, not a human interface. I expect the UI to list devices by name, but use the id for requests. The name has a higher chance of being non-unique.

James-Mart:
It appears nothing in the config file can be used to set the pkcs11-module so we have to use the CLI flag?

swatanabe:
> It appears nothing in the config file can be used to set the pkcs11-module so we have to use the CLI flag?

The config file and command line share most of their definitions. The only options that are available on the command line and not the config file are `--help` and the positional database.

swatanabe:
> The config file and command line share most of their definitions. The only options that are available on the command line and not the config file are `--help` and the positional database.

`--version` should be excluded from the config file, but isn't.

James-Mart:
Though it possible to set the pkcs11-module within the config file, it does not automatically get set if you use the command line flag

James-Mart:
For documentation purposes, here are some of the steps I performed while testing:

- install the pkcs11-module softhsm2, install xxd, install jq, install curl

- initialize a new token in softhsm2, using a command:
`softhsm2-util --init-token --slot 0 --label "psibase SoftHSM" --pin 1215 --so-pin 1215`

- run psinode with a pkcs11-module, host is psibase.127.0.0.1.sslip.io, and admin interface available on admin-sys.*, using: `psinode psinode_db --pkcs11-module /usr/lib/softhsm/libsofthsm2.so`. I tested that the pkcs11-module can be configured in either the command line or config file. However, using the command line doesn't automatically update the config file.

- get device id, using: `curl http://admin-sys.psibase.127.0.0.1.sslip.io:8079/native/admin/keys/devices | jq`
```
[
  {
    "name": "psibase SoftHSM",
    "id": "pkcs11:token=psibase%20SoftHSM;manufacturer=SoftHSM%20project;model=SoftHSM%20v2;serial=d74fc56b9ce782c3",
    "unlocked": false
  },
  ...
]
```

- unlock the device within psinode, using: `curl -d '{"device":"pkcs11:serial=d74fc56b9ce782c3","pin":"1215"}' -H 'Content-Type: application/json' http://admin-sys.psibase.127.0.0.1.sslip.io:8079/native/admin/keys/unlock`

- check that it was unlocked:
```
{
    "name": "psibase SoftHSM",
    "id": "pkcs11:token=psibase%20SoftHSM;manufacturer=SoftHSM%20project;model=SoftHSM%20v2;serial=d74fc56b9ce782c3",
    "unlocked": true
  },
```

- generate a new key in the device, using: `curl -d '{"service":"verify-sys","device":"pkcs11:token=psibase%20SoftHSM;manufacturer=SoftHSM%20project;model=SoftHSM%20v2;serial=d74fc56b9ce782c3"}' -H 'Content-Type: application/json' http://admin-sys.psibase.127.0.0.1.sslip.io:8079/native/admin/keys`
```
[{
    "service": "verify-sys",
    "rawData": "3059301306072A8648CE3D020106082A8648CE3D030107034200049283B84EFCEC275D310BA957FE6B6EFACF6AE0E247E6A0AD855F9A04E1C0D74AA7382C23BB3D770C3183E5603114CB09FBFE5F2D0EB638B6DE103D990F5B2167"
}]
```

- Put the raw key data into a file and convert it to binary, using: `xxd -p -r keyfile.hex > keyfile.bin`

- Convert the key from binary to pem format with openssl, using: `openssl pkey -pubin -outform PEM -in keyfile.bin > keyfile.pem`
```
-----BEGIN PUBLIC KEY-----
MFkwEwYHKoZIzj0CAQYIKoZIzj0DAQcDQgAEkoO4TvzsJ10xC6lX/mtu+s9q4OJH
5qCthV+aBOHA10qnOCwjuz13DDGD5WAxFMsJ+/5fLQ62OLbeED2ZD1shZw==
-----END PUBLIC KEY-----
```

- Boot the network using psibase and specifying the signing key, using: `psibase -a http://psibase.127.0.0.1.sslip.io:8079 boot -p myproducer -k keyfile.pem`. This worked, but then I couldn't sign transactions with any accounts on the booted network.

swatanabe:
>     * Though it possible to set the pkcs11-module within the config file, it does not automatically get set if you use the command line flag
> 

That applies to all options. Passing an option on the command line never updates the config file.

swatanabe:
>     * I can unlock a device by specifying the device with only the serial number, for example:
>       `curl -d '{"device":"pkcs11:serial=d74fc56b9ce782c3","pin":"1215"}' -H 'Content-Type: application/json' http://admin-sys.psibase.127.0.0.1.sslip.io:8079/native/admin/keys/unlock`
>       But if i use only the serial number when posting to `/native/admin/keys` to generate a key, I get the error, "No token matches URI". Works if I use the entire device string returned by the `/native/admin/keys/devices` endpoint, such as: `"device":"pkcs11:token=psibase%20SoftHSM;manufacturer=SoftHSM%20project;model=SoftHSM%20v2;serial=d74fc56b9ce782c3"`.

This doesn't make sense to me. unlock and keys use the exact same function to look up the token.

James-Mart (approved):

PR #516, opened 2023-11-22
Native OS X Build Fixes

bytemaster:
Requires one update to EOS VM, merging osx branch.

Support for clang17 via homebrew to compile native ARM on OS X

Please verify that these changes do not break other platforms, then merge.

See OSX_README.md for what I had to do to configure my environment.


James-Mart:
```
/home/runner/work/psibase/psibase/libraries/psibase/native/src/ExecutionContext.cpp:1:10: fatal error: psibase/NativeFunctions.hpp: No such file or directory
    1 | #include <psibase/NativeFunctions.hpp>
```

libraries/triedent/src/gc_queue.cpp
```diff
@@ -108,7 +108,7 @@ namespace triedent
 
    void gc_queue::notify_run()
    {
-      std::lock_guard{_queue_mutex};
+      std::lock_guard g{_queue_mutex};
```
swatanabe, line 111: This changes the behavior slightly. The original code is correct and it is intentional.

programs/psinode/main.cpp
```diff
@@ -87,7 +87,7 @@ std::string to_string(const native_service& obj)
 struct autoconnect_t
 {
    static constexpr std::size_t max   = std::numeric_limits<std::size_t>::max();
-   std::size_t                  value = max;
+   std::uint64_t value = max;
```
swatanabe, line 90: `max` and `value` should have the same type

programs/psinode/main.cpp
```diff
@@ -787,33 +787,33 @@ struct ThreadInfo
...
-   std::int64_t user;
-   std::int64_t system;
-   std::int64_t pageFaults;
-   std::int64_t read;
-   std::int64_t written;
+   int64_t user;
+   int64_t system;
+   int64_t pageFaults;
+   int64_t read;
+   int64_t written;
```
swatanabe, line 794: I don't understand the reason for this change.

programs/psinode/main.cpp
```diff
@@ -787,33 +787,33 @@ struct ThreadInfo
...
-   std::size_t data;
-   std::size_t wasmMemory;
-   std::size_t wasmCode;
-   std::size_t unclassified;
+   int64_t database;
+   int64_t code;
+   int64_t data;
+   int64_t wasmMemory;
+   int64_t wasmCode;
+   int64_t unclassified;
```
swatanabe, line 805: `size_t` is unsigned. Is there a reason for changing to signed?

programs/psinode/main.cpp
```diff
@@ -787,33 +787,33 @@ struct ThreadInfo
...
-   std::uint64_t unprocessed;
-   std::uint64_t total;
-   std::uint64_t failed;
-   std::uint64_t succeeded;
-   std::uint64_t skipped;
+   uint64_t unprocessed;
+   uint64_t total;
+   uint64_t failed;
+   uint64_t succeeded;
+   uint64_t skipped;
```
swatanabe, line 816: I don't understand why this is needed

programs/psitest/main.cpp
```diff
@@ -559,9 +560,11 @@ struct callbacks
          result.filetype = wasi_filetype_regular_file;
          result.nlink    = stat.st_nlink;
          result.size     = stat.st_size;
+#ifndef __APPLE__
          result.atim     = to_nsec(stat.st_atim);
          result.mtim     = to_nsec(stat.st_mtim);
          result.ctim     = to_nsec(stat.st_ctim);
+#endif  // apple
```
swatanabe, line 567: Does Apple support `st_atime` (second resolution instead of nanosecond resolution) or is the timespec available under another name (like `st_atimespec`)?

libraries/triedent/src/cache_allocator.cpp
```diff
@@ -21,12 +21,20 @@ namespace triedent
              [this]()
              {
                 thread_name("swap");
+#ifndef __APPLE__
                 pthread_setname_np(pthread_self(), "swap");
+#else // if __APPLE__
+                pthread_setname_np("swap");
+#endif
```
swatanabe, line 28: Can we wrap setting the thread name in a function, so that this #ifdef isn't duplicated all over?

libraries/triedent/src/ring_allocator.cpp
```diff
@@ -50,7 +50,7 @@ namespace triedent
    {
       // The lock here is needed to synchronize-with wait_swap
       // even through no data is accessed while holding the lock.
-      std::lock_guard{_free_mutex};
+      std::lock_guard g{_free_mutex};
```
swatanabe, line 53: The original code is intentional

OSX_README.md
```diff
@@ -0,0 +1,38 @@
...
+3. Install WASI_SDK 20.0 
+
+  https://github.com/WebAssembly/wasi-sdk/releases
+  set WASI_SDK_PREFIX to the root of the wasi-sdk release you downloaded
+
+
+5. Install icu4c 
+
+```
+git clone https://github.com/unicode-org/icu
```
swatanabe, line 31: Does the `brew` version of ICU not work for some reason?

libraries/psibase/native/src/Watchdog.cpp
```diff
@@ -71,10 +71,14 @@ namespace psibase
          elapsedOrStart(CpuClock::now().time_since_epoch()),
          handler(handler)
    {
+#ifdef __APPLE__
+      cpuclock = CLOCK_UPTIME_RAW;
```
swatanabe, line 75: `CpuClock::now` has a default parameter of `CLOCK_THREAD_CPUTIME_ID`. Either change that default to match this, or remove the default and change all uses of `now` to `CpuClock::now(cpuclock)`.

programs/psinode/CMakeLists.txt
```diff
@@ -13,6 +13,7 @@ target_link_libraries(psinode
     debug_eos_vm
     psibase_net
 )
+target_link_directories( psinode PUBLIC ${ICU_LIBRARY_DIR} )
```
swatanabe, line 16: Since the dependency on ICU is apparently coming from Boost through `psibase`, it would be better to set this once on the `psibase` library target.

swatanabe (commented):
clang-format

libraries/psibase/native/src/NativeFunctions.cpp
```diff
@@ -1,4 +1,5 @@
-#include <psibase/NativeFunctions.hpp>
+#include <psibase/RawNativeFunctions.hpp>
```
swatanabe, line 1: There's no reason to `#include` this here because the host functions are not available in this form for native code.

programs/psinode/main.cpp
bytemaster: Done.

programs/psinode/main.cpp
bytemaster: Personal preference, since uint64_t is builtin type.

libraries/triedent/src/cache_allocator.cpp
bytemaster: Wrapped in function set_current_thread_name()

OSX_README.md
bytemaster: This may not be necessary now that I have a path variable.

libraries/psibase/native/src/Watchdog.cpp
bytemaster: they now match with a value that works by default on APPLE

programs/psitest/main.cpp
bytemaster: It does, it has been updated for st_atimespec, etc

programs/psinode/main.cpp
```diff
@@ -2193,13 +2208,15 @@ int main(int argc, char* argv[])
          auto db_root     = std::filesystem::path(vm["database"].as<std::string>());
          option_path      = option_path / db_root;
          auto config_path = db_root / "config";
+         std::cout << "expecting config file: " << config_path<<"\n";
```
swatanabe: Remove debug prints.

bytemaster:
Since all tests have passed and review comments addressed I'm merging so I can start a new branch that isn't related to this issue.

CMakeLists.txt
```diff
@@ -235,8 +239,8 @@ if(BUILD_JS)
     DEPENDS AdminSys_js
     COMMAND rm -rf ${pfx}/share/psibase/services/admin-sys
     COMMAND mkdir -p ${pfx}/share/psibase/services/admin-sys/common
-    COMMAND cp -a -T ${CMAKE_CURRENT_SOURCE_DIR}/services/user/AdminSys/ui/dist ${pfx}/share/psibase/services/admin-sys
-    COMMAND cp -a -T ${CMAKE_CURRENT_SOURCE_DIR}/services/user/CommonSys/common/fonts ${pfx}/share/psibase/services/admin-sys/common/fonts
+    COMMAND cp -a ${CMAKE_CURRENT_SOURCE_DIR}/services/user/AdminSys/ui/dist ${pfx}/share/psibase/services/admin-sys
+    COMMAND cp -a ${CMAKE_CURRENT_SOURCE_DIR}/services/user/CommonSys/common/fonts ${pfx}/share/psibase/services/admin-sys/common/fonts
```
swatanabe, line 243: This change breaks the admin page

PR #532, opened 2023-12-07
add docs for url-encoding, database requirements, and new account nam…

James-Mart:
This is a proposal to initiate further discussion.


James-Mart:
The spec for URL encoding in this PR won't work. `app.branch.domain/path` would mean the node needs a separate TLS cert for every `branch` value. 

So alternatively, what if it was `app.domain/branch/path`. All client-side storage is then kept in the supervisor namespace, and supervisor manages proper isolation between branches? CC @bytemaster and @swatanabe

James-Mart:
Closing in favor of #707. Database doc updates removed for now

PR #534, opened 2023-12-11
Update cargo-psibase

James-Mart:
This reverts commit 2550f84d0475d3cf5ea3acb8328a5ccd42a774bc which unified all rust module version numbers under the workspace. This is much nicer for version management, however it precludes being able to publish updated versions to crates.io. See [this github issue](https://github.com/rust-lang/cargo/issues/11133) for more details.

### DIFF
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -14,7 +14,7 @@ dependencies = [
 
 [[package]]
 name = "account_sys"
-version = "0.5.0"
+version = "0.1.0"
 dependencies = [
  "fracpack",
  "psibase",
@@ -386,7 +386,7 @@ checksum = "a3e2c3daef883ecc1b5d58c15adae93470a91d425f3532ba1695849656af3fc1"
 
 [[package]]
 name = "burst-transfer"
-version = "0.5.0"
+version = "0.1.0"
 dependencies = [
  "anyhow",
  "chrono",
@@ -442,7 +442,7 @@ dependencies = [
 
 [[package]]
 name = "cargo-psibase"
-version = "0.5.0"
+version = "0.2.0"
 dependencies = [
  "anyhow",
  "binaryen",
@@ -830,7 +830,7 @@ dependencies = [
 
 [[package]]
 name = "elections"
-version = "0.5.0"
+version = "0.1.0"
 dependencies = [
  "fracpack",
  "psibase",
@@ -936,7 +936,7 @@ dependencies = [
 
 [[package]]
 name = "fracpack"
-version = "0.5.0"
+version = "0.2.0"
 dependencies = [
  "custom_error",
  "psibase_macros",
@@ -1052,7 +1052,7 @@ dependencies = [
 
 [[package]]
 name = "gen-cpp-doc"
-version = "0.5.0"
+version = "0.1.0"
 dependencies = [
  "anyhow",
  "clang",
@@ -1663,7 +1663,7 @@ dependencies = [
 
 [[package]]
 name = "md2man"
-version = "0.5.0"
+version = "0.1.0"
 dependencies = [
  "clap 3.2.25",
  "pulldown-cmark",
@@ -2232,7 +2232,7 @@ dependencies = [
 
 [[package]]
 name = "psibase"
-version = "0.5.0"
+version = "0.2.0"
 dependencies = [
  "anyhow",
  "async-graphql",
@@ -2268,7 +2268,7 @@ dependencies = [
 
 [[package]]
 name = "psibase_macros"
-version = "0.5.0"
+version = "0.2.0"
 dependencies = [
  "darling",
  "proc-macro-error",
@@ -2280,14 +2280,14 @@ dependencies = [
 
 [[package]]
 name = "psibase_names"
-version = "0.5.0"
+version = "0.2.0"
 dependencies = [
  "seahash",
 ]
 
 [[package]]
 name = "psispace-sys"
-version = "0.5.0"
+version = "0.1.0"
 dependencies = [
  "async-graphql",
  "fracpack",
@@ -2984,7 +2984,7 @@ dependencies = [
 
 [[package]]
 name = "test_contract"
-version = "0.5.0"
+version = "0.1.0"
 dependencies = [
  "psibase",
  "serde",
@@ -2993,7 +2993,7 @@ dependencies = [
 
 [[package]]
 name = "test_fracpack"
-version = "0.5.0"
+version = "0.1.0"
 dependencies = [
  "cxx",
  "cxx-build",
--- a/rust/psibase/Cargo.toml
+++ b/rust/psibase/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "psibase"
-version.workspace = true
+version = "0.2.0"
 rust-version.workspace = true
 repository.workspace = true
 homepage.workspace = true
@@ -16,14 +16,14 @@ anyhow = "1.0"
 async-graphql = "4.0"
 chrono = "0.4"
 custom_error = "1.9"
-fracpack = { path = "../fracpack" }
+fracpack = { version = "0.2.0", path = "../fracpack" }
 futures = "0.3"
 getrandom = { version = "0.2", features = ["js"] }
 include_dir = "0.7.3"
 sha2 = "0.10"
 mime_guess = "2.0"
-psibase_macros = { path = "../psibase_macros" }
-psibase_names = { path = "../psibase_names" }
+psibase_macros = { version = "0.2.0", path = "../psibase_macros" }
+psibase_names = { version = "0.2.0", path = "../psibase_names" }
 ripemd = "0.1"
 serde = { version = "1.0", features = ["derive", "rc"] }
 serde_bytes = "0.11.12"
--- a/rust/Cargo.toml
+++ b/rust/Cargo.toml
@@ -16,7 +16,7 @@ members = [
 ]
 
 [workspace.package]
-version = "0.5.0"
+version = "0.1.0"
 rust-version = "1.64"
 repository = "https://github.com/gofractally/psibase"
 homepage = "https://psibase.io"
--- a/rust/account_sys/Cargo.toml
+++ b/rust/account_sys/Cargo.toml
@@ -11,7 +11,7 @@ publish = false
 crate-type = ["cdylib", "staticlib"]
 
 [dependencies]
-fracpack = { path = "../fracpack" }
-psibase = { path = "../psibase" }
+fracpack = { version = "0.2.0", path = "../fracpack" }
+psibase = { version = "0.2.0", path = "../psibase" }
 serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0"
--- a/rust/burst-transfer/Cargo.toml
+++ b/rust/burst-transfer/Cargo.toml
@@ -14,7 +14,7 @@ anyhow = "1.0"
 chrono = "0.4"
 clap = {version = "3.1", features = ["derive"]}
 futures = "0.3"
-psibase = { path = "../psibase" }
+psibase = { version = "0.2.0", path = "../psibase" }
 rand = "0.8.5"
 reqwest = { version = "0.11", default-features = false, features = ["json","rustls-tls"] }
 serde = { version = "1.0", features = ["derive"] }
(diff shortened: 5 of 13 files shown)


rust/cargo-psibase/src/link.rs
```diff
@@ -243,221 +51,326 @@ pub fn link(filename: &Path, code: &[u8], polyfill: &[u8]) -> Result<Vec<u8>, an
...
-    if module.start_section().is_some() {
-        return Err(anyhow!(
-            "{} has unsupported start section",
-            filename.to_string_lossy()
-        ))?;
+    Ok(())
+}
+
+/// The module is not a supported target for polyfilling if it has the following:
+/// * A start module
```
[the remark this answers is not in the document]
James-Mart: Yep

rust/cargo-psibase/src/link.rs
```diff
@@ -243,221 +51,326 @@ pub fn link(filename: &Path, code: &[u8], polyfill: &[u8]) -> Result<Vec<u8>, an
...
-    {
-        if let OldToNewFn::Import(import) = item {
-            if let Some(index) = import_map.get(import) {
-                *item = OldToNewFn::ResolvedImport(*index);
+    return Ok(Some(targets));
+}
+
+/// Gets the FunctionID if it exists for a function in a `wasm_module` based on a 
+/// `function_module` and `name`.
+fn get_import_fid(function_module : &String, name : &String, wasm_module : &Module) -> Result<Option<FunctionId>, anyhow::Error> {
```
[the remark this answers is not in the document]
James-Mart: Fixed

rust/cargo-psibase/src/link.rs
```diff
@@ -243,221 +51,326 @@ pub fn link(filename: &Path, code: &[u8], polyfill: &[u8]) -> Result<Vec<u8>, an
...
+    };
+    Ok(new_fid)
+}
+
+/// Copies one instruction sequence from a function in a `source` module to a sequence 
+/// in a `dest` module.
+fn copy_instr_block(source_block_id : InstrSeqId, source_function : &LocalFunction, source : &Module, dest_seq : &mut InstrSeqBuilder, dest : &mut Module) -> Result<(), anyhow::Error> {
+
+    // Get the instructions from the specified source block
+    let instrs : Vec<Instr> = source_function.block(source_block_id).instrs.iter().map(|(instr, _)|{instr.clone()}).collect();
```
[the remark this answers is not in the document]
James-Mart: It neither needs to be collected into a vector, nor is the double clone required. Both have been fixed

rust/cargo-psibase/src/link.rs
```diff
@@ -243,221 +51,326 @@ pub fn link(filename: &Path, code: &[u8], polyfill: &[u8]) -> Result<Vec<u8>, an
...
+             | Instr::TableFill(_)
+             | Instr::RefNull(_)
+             | Instr::RefIsNull(_)
+             | Instr::V128Bitselect(_)
+             | Instr::I8x16Swizzle(_)
+             | Instr::I8x16Shuffle(_)
+             | Instr::LoadSimd(_)
+             | Instr::TableInit(_)
+             | Instr::ElemDrop(_)
+             | Instr::TableCopy(_) => {/* No-op */}
```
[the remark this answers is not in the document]
James-Mart: Yes, just hadn't added it yet

rust/cargo-psibase/src/link.rs
```diff
@@ -291,6 +297,26 @@ fn provide_dest_seq_id(source_block_id : InstrSeqId, source_function : &LocalFun
                 dest_seq.instr(i);
             }
 
+            // Local reference updates
+            Instr::LocalGet(local_ref) => {
+                let mut i = local_ref.clone();
+                let ty = source.locals.get(local_ref.local).ty();
+                i.local = dest.locals.add(ty);
```
swatanabe, line 304: Are you sure that this is right? It looks to me like it's creating a new variable for every instruction

James-Mart: Def would not say I'm sure it's right. But I do know that adding the local will only actually add something if it's a new type. And also the locals object here is internal to walrus, once you actually emit the wasm there's no actual "locals" table

swatanabe: That would also be wrong as it would merge all locals of the same type. According to https://docs.rs/walrus/latest/walrus/struct.ModuleLocals.html#method.add, it always returns a new local id. Also, I don't quite understand your comment: it's the ID of the local, not the ID of a type.

James-Mart: Meant to say that adding a `type` would only add one if it's a new type - regardless I guess not really relevant. 

Okay I updated this but I'm still not sure it's correct. In the latest, it should create a new local only if it's the first time a local with a given ID is referenced while parsing instructions from the source module. Because I'm not seeing an instruction that explicitly creates a local?

swatanabe: There isn't an instruction that creates locals. Locals are defined in the function prologue.  Function parameters are accessed as locals, so they also need to be translated.

rust/cargo-psibase/src/link.rs
```diff
@@ -535,17 +535,115 @@ pub fn link_module(source : &Module, dest : &mut Module) -> Result<(), anyhow::E
...
+        new_validator().validate_all(&filled.emit_wasm()).map_err(|e|anyhow!("[Validating filled, simple] {}", e))?;
+        Ok(())
+    }
+
+    #[test]
+    fn valid_wasm_produced_2() -> Result<(), Error> {
+        let (mut original, _, mut filled) = setup(INTERMEDIATE_WASM)?;
+
+        new_validator().validate_all(&original.emit_wasm()).map_err(|e|anyhow!("[Validating original, intermediate] {}", e))?;
+        new_validator().validate_all(&filled.emit_wasm()).map_err(|e|anyhow!("[Validating filled, intermediate] {}", e))?;
```
James-Mart, line 646: This line fails with this error:
https://hotg-ai.github.io/rune/crate-docs/src/wasmparser/operators_validator.rs.html#323

rust/psibase/Cargo.toml
```diff
@@ -1,12 +1,13 @@
...
 name = "psibase"
-version.workspace = true
+version = "0.2.0"
 rust-version.workspace = true
 repository.workspace = true
 homepage.workspace = true
 edition = "2021"
 description = "Library and command-line tool for interacting with psibase blockchains"
 license = "MIT"
+include = ["src/**/*", "boot-image/**/*", "Cargo.toml"]
```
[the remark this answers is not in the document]
James-Mart: Removed this

rust/publish.sh
```diff
@@ -0,0 +1,40 @@
...
+for dir in "${dirs[@]}"; do
+    cd "$dir"
+    echo "Publishing $dir..."
+
+    # psibase crate needs `--allow-dirty` to be published because the boot-image directory
+    #   is copied into the project directory but is not added to git.
+    extra_flags=""
+    if [ "$dir" = "psibase" ]; then
+        extra_flags="--allow-dirty"
+    fi
```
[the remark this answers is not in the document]
James-Mart: Removed

rust/cargo-psibase/src/link.rs
James-Mart: Fixed. Structured control instructions (if, block, etc) were using the wrong type

rust/cargo-psibase/src/link.rs
```diff
@@ -243,221 +51,326 @@ pub fn link(filename: &Path, code: &[u8], polyfill: &[u8]) -> Result<Vec<u8>, an
...
-    {
-        return Ok(code.to_owned());
+    Ok(())
+}
+
+/// A module is not a candidate for polyfilling if it has any of the following:
+/// * No types
+/// * No imports
+/// * No functions
+/// * No function bodies
```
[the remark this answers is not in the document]
James-Mart: Restriction removed

rust/cargo-psibase/src/link.rs
```diff
@@ -1,463 +1,704 @@
...
+                let (params, results) = (ty.params().to_vec(), ty.results().to_vec());
+                let dest_type = dest.types.add(&params, &results);
+                dest.add_import_func(module, name, dest_type).0
             }
-        })
-        .collect()
+        },
+        Local(_) => {
+            let new_local_func = copy_func(source_fid, source, dest)?;
+            dest.funcs.add_local(new_local_func)
```
swatanabe, line 185: I don't see anything protecting against duplicating functions from the polyfill.

rust/cargo-psibase/src/link.rs
swatanabe, line 1: Did you run `cargo fmt`?

rust/scripts/set-version.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+# Run with 
+#   `./scripts/set-version.sh x.y.z`
+# Can also dry-run with:
+#   `./scripts/set-version.sh "x.y.z --dry-run"`
```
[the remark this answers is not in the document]
James-Mart: Done

rust/cargo-psibase/src/link.rs
```diff
@@ -1,463 +1,704 @@
...
+                // The import already exists in the dest
+                if dest.func_type(dest_fid) != source.func_type(source_fid) {
+                    return Err(anyhow!("Type mismatch between imports"));
+                }
+                dest_fid
             } else {
-                None
+                // The import needs to be added to the dest. Update the called fid.
+                let ty = source.types.get(source_func_ty);
+                let (params, results) = (ty.params().to_vec(), ty.results().to_vec());
```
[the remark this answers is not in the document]
James-Mart: Removed

rust/cargo-psibase/src/link.rs
James-Mart: Fixed

rust/cargo-psibase/test-data/intermediate.wasm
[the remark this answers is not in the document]
James-Mart: Removed - I actually added validation to every wasm produced by this linker. It's only slightly increases build time and it seems like a good sanity check to have. This also means future services that automatically build using cargo-psibase in our build system will effectively reproduce the removed unit test.

rust/cargo-psibase/src/link.rs
James-Mart: I had not. I now have, and also updated the default `settings.json` to make sure `cargo fmt` runs on save.

James-Mart:
CI/CD failure is:
```
/usr/bin/ar: unable to copy file 'lib/libbinaryen.a'; reason: No space left on device
2024-03-03T18:40:01.7020452Z   gmake[2]: *** [CMakeFiles/binaryen.dir/build.make:438: lib/libbinaryen.a] Error 1
```
Happens since trying to add a `cargo test` step to ctest.

rust/Cargo.lock
swatanabe, line 1: It looks like this increases the minimum rust version to build. The README will need to be updated
```
bumpalo v3.15.3` cannot be built because it requires rustc 1.73.0 or newer
```

swatanabe:
> CI/CD failure is:
> 
> ```
> /usr/bin/ar: unable to copy file 'lib/libbinaryen.a'; reason: No space left on device
> 2024-03-03T18:40:01.7020452Z   gmake[2]: *** [CMakeFiles/binaryen.dir/build.make:438: lib/libbinaryen.a] Error 1
> ```
> 

binaryen is huge.  It's using >4GiB in the build directory. It looks like it's only used in build.rs. Maybe replace that with wasm-opt as well?

James-Mart:
> > CI/CD failure is:
> > ```
> > /usr/bin/ar: unable to copy file 'lib/libbinaryen.a'; reason: No space left on device
> > 2024-03-03T18:40:01.7020452Z   gmake[2]: *** [CMakeFiles/binaryen.dir/build.make:438: lib/libbinaryen.a] Error 1
> > ```
> 
> binaryen is huge. It's using >4GiB in the build directory. It looks like it's only used in build.rs. Maybe replace that with wasm-opt as well?

Oh yeah, I should do that anyway

rust/Cargo.lock
James-Mart: Done

swatanabe (approved):

PR #535, opened 2023-12-11
Packaging for apps

swatanabe:
Package Format
===

A psibase package is a zip archive which contains the following files:

- meta.json
- service/&lt;service&gt;.wasm
- service/&lt;service&gt;.json
- data/&lt;service&gt;/*
- script/postinstall.json

All files except meta.json are optional. A package may contain more than one service.

I chose zip over tar for 2 reasons:
- jszip appears more complete and better maintained than any of javascript tar readers
- zip compresses files separately, which makes it easier to extract just meta.json for the UI

Features in this PR
===
- [x] Build packages with cmake
- [x] Packages are installed in the correct location in the build directory
- [x] Packages are installed in the correct location in the install tree
- [x] `psibase boot <packages>` uses packaged services from `<prefix>/share/psibase/services` 
- [x] Meta-package Default that has no content but depends on all services that should be installed by default
- [x] psibase boot with no packages specified uses Default rather than trying to boot with nothing
- [x] Packages available from admin-sys
- [x] Admin panel boot loads services over HTTP
- [ ] Admin panel boot can upload packages from the local filesystem
- [ ] psibase option to specify one or more package repositories
- [ ] psibase can take package files on the command line
- [ ] psibase deploy (or perhaps a new subcommand) can accept a package to install or update a package on a running chain
- [x] Add scripts to build and submit arbitrary transactions, so that the initial token users can be defined in a package
- [ ] C++ psidk can build packages
- [ ] cargo psibase can build packages


CMakeLists.txt
```diff
@@ -256,83 +288,328 @@ if(BUILD_JS)
...
     install(FILES services/user/CommonSys/common/rpc.mjs services/user/CommonSys/common/keyConversions.mjs DESTINATION ${CMAKE_INSTALL_DATADIR}/psibase/services/admin-sys/common COMPONENT ServerData)
+    install(FILES ${CMAKE_CURRENT_BINARY_DIR}/share/psibase/services/admin-sys/packages DESTINATION ${CMAKE_INSTALL_DATADIR}/psibase/services/admin-sys COMPONENT ServerData)
 endif()
 
+# Build services
+include(pack_service.cmake)
+
+set(SERVICE_DIR ${CMAKE_CURRENT_BINARY_DIR}/share/psibase/packages)
+
+psibase_package(
```
James-Mart, line 299: Could each of these be added into each project directory CMakeLists?

pack_service.cmake
```diff
@@ -0,0 +1,252 @@
...
+        string(APPEND result true)
+    endif()
+    string(APPEND result "}")
+    file(GENERATE OUTPUT ${_OUTPUT} CONTENT ${result})
+endfunction()
+
+# NAME <name>               - The name of the package
+# DESCRIPTION <text>        - The package description
+# OUTPUT <filename>         - The package file. Defaults to ${NAME}.psi
+# ACCONTS <name>...         - Additional non-service accounts to create
```
James-Mart, line 68: ACCOUNTS*

pack_service.cmake
```diff
@@ -0,0 +1,252 @@
...
+#   DATA GLOB <path>... <dir> - Uploads multiple files to a directory
+#   INIT                    - The service has an init action that should be run with no arguments
+function(psibase_package)
+    set(keywords NAME DESCRIPTION OUTPUT PACKAGE_DEPENDS DEPENDS ACCOUNTS SERVICE DATA WASM FLAGS SERVER INIT POSTINSTALL)
+    foreach(keyword IN LISTS keywords)
+        set(_${keyword})
+    endforeach()
+    set(_SERVICES)
+    math(EXPR end ${ARGC}-1)
+    foreach(n RANGE ${end}})
```
James-Mart, line 85: Extra closing curly brace?

pack_service.cmake
```diff
@@ -0,0 +1,252 @@
...
+        set(my_keyword)
+        foreach(keyword IN LISTS keywords)
+            if (ARGV${n} STREQUAL keyword)
+                set(my_keyword ${keyword})
+            endif()
+        endforeach()
+        if(my_keyword)
+            if(current_keyword STREQUAL "DATA")
+                list(APPEND _DATA_${_SERVICE} DATA)
```
James-Mart, line 94: Is this meant to append the string "DATA"?

rust/psibase/src/boot.rs
```diff
@@ -490,20 +182,16 @@ pub fn get_initial_actions(
 /// create your own.
```
James-Mart, line 182: Probably should update the doc comment

services/user/TokenSys/src/TokenUsers.json
James-Mart, line 1: I guess this is more than just setting token users. Maybe it should just be called, "post-install.json".

James-Mart:
If there are any actions that need to be called on or by a service after installation, then an init action should be used. Therefore at first I was thinking that a post-install script is redundant. But I think a post-install script could still be useful for an optional post-install procedure? So different people could use a different post-install procedure for the same package, and you could know (via comparing the hash of the service in each case) that the service was identical and only the post-install script has changed... I guess that's useful? Am I missing another reason why post-install is helpful?

James-Mart:
Also, I think that you should add something to the specification docs as part of this PR

pack_service.cmake
swatanabe: Yes

swatanabe:
> If there are any actions that need to be called on or by a service after installation, then an init action should be used. Therefore at first I was thinking that a post-install script is redundant. But I think a post-install script could still be useful for an optional post-install procedure? So different people could use a different post-install procedure for the same package, and you could know (via comparing the hash of the service in each case) that the service was identical and only the post-install script has changed... I guess that's useful? Am I missing another reason why post-install is helpful?

There are a few reasons why an init action is insufficient:
- Additional authorization beyond the service account may be needed
- The account might not have a wasm at all

CMakeLists.txt
swatanabe: It's possible, but
- The js build would also need to be moved which increases build system recursion depth
- Some of the packages (DocSys, Default) don't have services
- In some cases, several services should be included in a single package, e.g. TokenSys and SymbolSys probably belong in the same package because they're mutually dependent.

services/user/TokenSys/src/TokenUsers.json
swatanabe: This isn't part of the TokenSys package. It's a separate package called TokenUsers. Besides creating the token users all it does is some configuration of the token.  This was all originally behind `install_token_users`.

James-Mart (approved):

PR #537, opened 2023-12-19
Add supervisor

Velua:


services/user/App1/src/App.tsx
```diff
@@ -0,0 +1,27 @@
+import "./App.css";
+import { connect } from "@psibase/plugin";
```
James-Mart, line 2: This should come from `@psibase/common-lib` or `/common`

libraries/psibase/tester/src/DefaultTestChain.cpp
James-Mart, line 1: This is the file that is used for setting up the chain that the tester uses. The tester is what runs the C++ service unit tests. I don't think we need to update the tester chain with any of our POC code.

services/system/AccountSys/ui/src/App.tsx
```diff
@@ -39,6 +43,23 @@ export interface AccountWithKey extends AccountWithAuth {
     privateKey: KeyPairWithAccounts["privateKey"];
 }
 
+/// Problem: All these exisitng system / user apps are using the old operations / mutations or whatever
+// Solution: Migrate them all to Plugin WASMs, and figure out
+// Importables! -> Providing a JS function to the WASM for it to call, to add actions to the actions array,
+// We don't solely operate on the return values of these WASM functions, we also require to monitor the functions that are called (that we give it via importables)
```
James-Mart, line 49: Idk what you mean we are required to monitor the functions that are called

services/system/AccountSys/ui/src/App.tsx
```diff
@@ -39,6 +43,23 @@ export interface AccountWithKey extends AccountWithAuth {
...
+//
+// Do we have operations that do network requests prior to providing actions to submit to chain?
+// JS Operations -> Can do network requests
+// WASM operations / functions -> Can't.
+//
+// Solution 1: Provide Rust with the option of network calls via importables
+// Solution 2: Figger it out in the app rather than operations... and dumb it down for the WASM
+//
+// In the future, we want these WASM functions to know context, e.g. who called this? and what app they were in when it was called?
```
James-Mart, line 58: This will be done via an importable like, "getSender". The loader knows who called it because that information was sent by the supervisor, so it can fill in the getSender importable with the name of the caller service.

services/user/App1/src/App.tsx
Velua: Why can't we use the external package if other app developers are gonna use it?

James-Mart: I just meant that this API is part of the common libraries. And since supervisor does more than just plugin communication, I think it should be called "common-lib" or "common" rather than "plugin"

services/user/App1/src/App.tsx
```diff
@@ -0,0 +1,27 @@
...
+
+function App() {
+  const run = async () => {
+    const res = await connect();
+
+    console.log(res, "came back");
+    const back = await res.functionCall({
+      service: "app2",
```
cool-ant, line 10: `plugin` instead of `service`? so that `method` doesn't sound like `action`?

cool-ant: `account`?
`plugin-method`?

services/user/TokenSys/wasm/src/lib.rs
```diff
@@ -0,0 +1,32 @@
...
+impl Guest for Component {
+    /// Say hello!
+    fn transfer(to: String, amount: u32) -> String {
+        let logged_in_user = imports::get_logged_in_user();
+
+        println!("logged_in_user x, 'is a thing.. {:?}", logged_in_user);
+
+        let test_params = Funccallparams {
+            service: "test_service".to_string(),
+            method: "test_method".to_string(),
```
cool-ant, line 20: should this be `action`, not `method`?

services/user/App1/src/App.tsx
cool-ant: and perhaps `functionCall` --> `pluginFnCall`?

PR #542, opened 2024-01-07
Apps use root window

James-Mart:
This PR:
* Eliminates the /applet/ routing
* Deletes old vanilla UIs (idk why they were even still around)
* Stops using initializeApplet for DocSys and AccountSys apps
  * For the other apps that do still use initializeApplet, they at least don't try to redirect because I removed the auto-redirection from initializeApplet
* Deletes the `generators/` project. The generated apps were using old architecture, we can remake generators later once everything stabilizes


James-Mart:
Oh and I'm using John's `add-supervisor` branch as our feature branch

James-Mart:
This also breaks all apps that required the parent window for messaging. Those apps will either need to fixed or deleted

services/user/CommonSys/ui/src/views/dashboard.tsx
```diff
@@ -1,16 +1,34 @@
...
+            const newHrefs: { [key: string]: string } = {};
+
+            for (const applet of applets) {
+                newHrefs[applet.service] = await siblingUrl(
+                    null,
+                    applet.service
+                );
+            }
+
+            setHrefs(newHrefs);
```
Velua, line 23: Could probably get away with a `Promise.all` here rather than one at a time, but no biggy, not worth holding this up. :+1:

PR #543, opened 2024-01-08
Update CI/CD builder image

James-Mart:


.github/workflows/ubuntu-builder.yml
```diff
@@ -11,9 +11,9 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - builder-image: "ghcr.io/${{ github.repository_owner }}/psibase-builder-ubuntu-2004:9e5daca8d7a08cf877ac4c40a9d3c12cdb44e534"
+          - builder-image: "ghcr.io/${{ github.repository_owner }}/psibase-builder-ubuntu-2004:f202fea25d15c43afc8361551313e981ce2a5f87"
```
swatanabe, line 14: I can't figure out what this image is. I can see it in the packages list, but there's no such revision in the image-builders repo.

James-Mart: Good catch, this made me realize artifacts are being uploaded on PR, instead of just on merge.

James-Mart: I ended up reworking the CI/CD logic in the `image-builders` repo. See [this PR](https://github.com/gofractally/image-builders/pull/18).

James-Mart:
I updated the version here to one that is tied to an actual commit ID in the builders repo.
I thought about updating this PR with some code that would automatically pull the latest builder image instead of using a hardcoded one, but I will move that into a separate PR to fast-track this.

swatanabe (approved):

PR #545, opened 2024-01-10
Try to fix watchdog tests again.

swatanabe:
I have never been able to reproduce this locally. Maybe the control thread is getting starved.


swatanabe:
This doesn't seem to help

PR #546, opened 2024-01-11
Add support for users to build packages

swatanabe:


libraries/psibase/sdk/pack_service.cmake
```diff
@@ -63,6 +63,7 @@ endfunction()
 # DEPENDS <targets>...      - Targets that this target depends on
 # PACKAGE_DEPENDS <name>... - Other psibase packages that the package depends on
 # SERVICE <name>            - The account name of a service
+#   TARGET <target>         - The wasm file that will be deployed
 #   WASM <filename>         - The wasm file that will be deployed
```
James-Mart, line 67: Unclear from this comment what the difference is between TARGET and WASM

swatanabe: One takes the name of a target, the other takes a path. I didn't want the comment to be too long here. There's a more detailed description in the docs.

James-Mart (approved):

PR #548, opened 2024-01-12
rework psibase upload

swatanabe:
- Add `-r` which subsumes the functionality of `upload-tree`
- Make content-type an optional argument and default to deducing it from the file name
- Reorder the source and dest to be more consistent with unix conventions
- Make dest optional. With `-r`, the dest defaults to `/` which is almost always going to be the right thing. Without `-r`, the destination is derived from the source file name. i.e. `path/to/index.html` will be uploaded to `/index.html`
- If the dest does not have a leading `/`, add one automatically.


James-Mart:
Awesome

James-Mart (approved):

PR #554, opened 2024-01-30
Add account root that owns all accounts created by packages

swatanabe:


services/system/AuthDelegateSys/include/services/system/AuthDelegateSys.hpp
```diff
@@ -0,0 +1,70 @@
...
+      /// Verifies that a particular user is allowed to use a particular auth service.
+      ///
+      /// This action allows any user who has already set an owning account with `AuthDelegateSys::setOwner`.
+      void canAuthUserSys(psibase::AccountNumber user);
+
+      /// Set the sender's public key
+      ///
+      /// This is the public key that must be claimed by the transaction whenever a sender using this auth service
+      /// submits a transaction.
+      void setOwner(psibase::AccountNumber owner);
```
James-Mart, line 58: Comment needs to be updated

services/system/ProducerSys/include/services/system/ProducerSys.hpp
```diff
@@ -14,6 +14,13 @@ namespace SystemService
       static constexpr auto service      = psibase::AccountNumber("producer-sys");
       static constexpr auto serviceFlags = psibase::CodeRow::allowWriteNative;
 
+      // An action authorized by this account must be authorized by at
+      // least one honest producer
+      static constexpr auto producerAccountWeak = psibase::AccountNumber("prods-weak");
+      // If two actions are authorized by this account, there is at least
+      // one honest producer that authorized both actions.
+      static constexpr auto producerAccountStrong = psibase::AccountNumber("prods-strong");
```
James-Mart, line 22: I don't know what you mean by "honest" in these comments.

swatanabe: An honest producer is one that follows the rules, written or unwritten, for the chain and the actions being authorized. What that means is going to be dependent on the specific actions.  For system services it mostly means not breaking the chain.

James-Mart: I think you're using the assumption that there are some minimum number of honest producers according to the consensus rules, but I think most people will read this and just think, "how do you know they're honest?". 

Also, the comment over checkAuthSys should be updated since it's no longer just ">2/3rds active producers"

services/system/ProducerSys/src/ProducerSys.cpp
```diff
@@ -74,14 +74,20 @@ namespace SystemService
       psibase::kvPut(StatusRow::db, StatusRow::key(), *status);
    }
 
-   std::size_t getThreshold(const CftConsensus& bft)
+   std::size_t getThreshold(const CftConsensus& bft, AccountNumber account)
```
James-Mart, line 77: Shouldn't this variable name be cft?

James-Mart:
PR title suggests that all accounts created by packages are owned by root, but actually they are owned by whatever sender installs them, and system packages are installed by root.

swatanabe:
> PR title suggests that all accounts created by packages are owned by root, but actually they are owned by whatever sender installs them, and system packages are installed by root.

Non-root doesn't quite work due to the restrictions on account creation

James-Mart (approved):

PR #560, opened 2024-02-01
Fix timeout for http connections

swatanabe:
- The timer should be paused while processing transactions and queries
- Add an option to control the timeout


doc/psidk/src/run-infrastructure/cli/psinode.md
```diff
@@ -77,6 +77,10 @@ psinode - The psibase blockchain server
   - `bearer`:*key*: Allows access with a bearer token, which must be sent in the HTTP `Authorization` header. The key is an arbitrary string which will be used to verify the tokens. Tokens can be generated by `psibase create-token` or by the `/native/admin/login` endpoint.
   This option may be specified more than once. A client can access the admin API if it satisfies any of the conditions.
 
+- `--http-timeout` *seconds*
+
+  tells psinode how long to wait before closing an idle connection.
```
James-Mart, line 82: I think here would be the right place to mention that the units can be specified as: nothing (seconds), s, ms, us/μs, or ns

programs/psinode/main.cpp
```diff
@@ -1427,6 +1541,11 @@ void to_config(const PsinodeConfig& config, ConfigFile& file)
           "", "admin-authz", admin_authz, [](std::string_view text) { return std::string(text); },
           "Authorization for admin access");
    }
+   if (config.http_timeout != Timeout::none())
+   {
+      file.set("", "http-timeout", to_string(config.http_timeout),
+               "The maximum time for HTTP clients to send or receive a message");
```
James-Mart, line 1547: I think something like "The timeout before the server closes idle HTTP connections" is a clearer description.

programs/psinode/main.cpp
```diff
@@ -2460,6 +2584,8 @@ int main(int argc, char* argv[])
    // specify default token/service
    opt("leeway", po::value<uint32_t>(&leeway_us)->default_value(200000),
        "Transaction leeway, in µs.");
+   opt("http-timeout", po::value(&http_timeout)->default_value({}, "")->value_name("seconds"),
+       "The maximum time for HTTP clients to send or receive a message");
```
James-Mart, line 2588: I think something like "The timeout before the server closes idle HTTP connections" is a clearer description.

James-Mart:
I couldn't seem to force a timeout on boot even setting `http-timeout` to 1us

swatanabe:
> I couldn't seem to force a timeout on boot even setting `http-timeout` to 1us

Works for me:
```
$ ./psinode ./tmpdba -l 8080 --host psibase.127.0.0.1.sslip.io -p a --http-timeout=1us
$ rust/release/psibase boot -p a
Error: Failed to boot

Caused by:
    0: error sending request for url (http://psibase.127.0.0.1.sslip.io:8080/native/push_boot): connection error: Connection reset by peer (os error 104)
    1: connection error: Connection reset by peer (os error 104)
    2: Connection reset by peer (os error 104)
```

Note that 1ns won't work, because the representation of no timeout is currently wrong.

James-Mart:
```
$ psinode ./tempdb -l 8089 --host psibase.127.0.0.1.sslip.io -p a --http-timeout=1us
[2024-02-05T14:59:33.873Z] [warning]: unable to lock memory for ./tempdb. Try upgrading your shell's limits using "sudo prlimit --memlock=-1 --pid $$". If that doesn't work, try running psinode with "sudo".
[2024-02-05T14:59:33.874Z] [info]: Active producers: CFT:[]
[2024-02-05T14:59:33.875Z] [info]: Node is active producer
[2024-02-05T14:59:33.875Z] [info]: Starting block production for term 1 as a
[2024-02-05T14:59:33.875Z] [info]: Listening on 0.0.0.0:8089
[2024-02-05T14:59:33.975Z] [notice]: Need genesis block; use 'psibase boot' to boot chain
```
```
$ psibase -a http://psibase.127.0.0.1.sslip.io:8089/ boot -p a
███████████████████████████████████████████████████████████████████████ 10/10
Ok
```

swatanabe:
Can you check the value of the timeout in the config: `curl http://localhost:8089/native/admin/config | jq`? If you connect with the browser or another client that keeps persistent connections, does the timeout trigger?

James-Mart:
> Can you check the value of the timeout in the config

Interesting...  
```
psinode ./tempdb -l 8089 --host psibase.127.0.0.1.sslip.io -p a --http-timeout=1us
```

```
curl http://localhost:8089/native/admin/config | jq
...
"http_timeout": null,
...
```

programs/psinode/main.cpp
```diff
@@ -264,7 +264,7 @@ void validate(boost::any& v, const std::vector<std::string>& values, Timeout*, i
       {
          divisor *= 1000;
       }
-      else if (s == "us" || s == "µs")
+      else if (s == "us" || s == "µs" || s == "μs")
```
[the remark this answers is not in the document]
swatanabe: U+00B5 (micro sign)
U+03BC (greek small letter mu)

I mostly included these for completeness. I don't expect anyone to actually type unicode in.

James-Mart:
> I see the problem. The value is being divided by 10, which causes it to truncate to 0, which was (before my last round of fixes) treated as infinite.

After your most recent fix, it's working. This is with 1us timeout:
```
$ psibase -a http://psibase.127.0.0.1.sslip.io:8089/ boot -p a
Error: Failed to boot

Caused by:
    0: error sending request for url (http://psibase.127.0.0.1.sslip.io:8089/native/push_boot): connection closed before message completed
    1: connection closed before message completed
```

James-Mart (approved):

programs/psinode/main.cpp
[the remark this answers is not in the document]
cool-ant: Opt-m on a Mac. I do it cuz I can! :)

swatanabe:
> > I see the problem. The value is being divided by 10, which causes it to truncate to 0, which was (before my last round of fixes) treated as infinite.
> 
> After your most recent fix, it's working.

To see the current problem, set the timeout higher and then check the config: The timeout that it shows will be smaller than you specified.

James-Mart (changes requested):
Unapproving for the following reason:


> @swatanabe: To see the current problem, set the timeout higher and then check the config: The timeout that it shows will be smaller than you specified.

James-Mart:
```
psinode ./tempdb -l 8089 --host psibase.127.0.0.1.sslip.io -p a --http-timeout=4s
```
```
$ curl http://localhost:8089/native/admin/config | jq
...
"http_timeout": "400000",
...
```

400`000 us = 0.4 sec

swatanabe:
> 400`000 us = 0.4 sec

I thought I fixed this in 594e43a

James-Mart (approved):

PR #564, opened 2024-02-02
Show service stack trace on transaction failure

swatanabe:
- Change psio json serialization of variant to {type:value} to match serde. It's the same number of characters and the serde version seems more sane.
- Allow rust deserialization of TransactionTrace to handle number-as-string.
- Ensure that the ActionTrace::error field is set
- Add --trace=error|stack|full|json to determine the verbosity

### DIFF
--- a/doc/psidk/src/run-infrastructure/cli/psibase.md
+++ b/doc/psidk/src/run-infrastructure/cli/psibase.md
@@ -37,6 +37,14 @@ psibase - The psibase blockchain command line client
   - A PKCS #11 URI
   - An EOS style base58-encoded private key beginning `PVT_K1_`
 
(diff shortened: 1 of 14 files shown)


James-Mart:
> * Change psio json serialization of variant to {type:value} to match serde. It's the same number of characters and the serde version seems more sane.
> * Add --trace=message|stack|full|json to determine the verbosity

Both of these probably should have associated docs changes

swatanabe:
> > * Change psio json serialization of variant to {type:value} to match serde.
> 
> Both of these probably should have associated docs changes

The data-formats/json.md page needs a complete rewrite, as it isn't proper documentation at all.

James-Mart (approved):

PR #566, opened 2024-02-03
This is the fix for the rollup load issue, but...

cool-ant:
this branch isn't working. Be back to this after my original work is done.


Velua (approved):
This is already an improvement over what's existing, lets merge.

PR #568, opened 2024-02-06
Fix builds without documentation

swatanabe:
BUILD_DOC is off by default, but several pieces of CMakeLists.txt referenced it unconditionally.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -420,6 +420,8 @@ psibase_package(
     DEPENDS wasm
 )
 
+set(DOC_SYS)
+
 if(BUILD_DOC)
     psibase_package(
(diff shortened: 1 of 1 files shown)


James-Mart:
I think the optionality of docs was originally just because it didn't fit in boot. We could just remove the build flag altogether and leave it up to package selection at boot if a user doesn't want to install it.

swatanabe:
We also have `BUILD_NATIVE`, `BUILD_RUST`, `BUILD_JS`, `BUILD_EXAMPLES`, `BUILD_DEBUG_WASM`, and `BUILD_RELEASE_WASM`, at least half of which don't work, because we're not testing with them disabled.

PR #569, opened 2024-02-06
Remove cmake variables for partial builds

swatanabe:
Remove BUILD_DOC, BUILD_JS, BUILD_NATIVE, BUILD_RELEASE_WASM, BUILD_RUST, and CLIO_ONLY. We're not testing these options, and the interdependencies cause them to break on a regular basis.


.vscode/tasks.json.sample
```diff
@@ -18,7 +18,7 @@
         {
             "label": "Build Services",
             "type": "shell",
-            "command": "mkdir -p build && cd build && cmake -DCMAKE_EXPORT_COMPILE_COMMANDS=ON -DCMAKE_BUILD_TYPE=Debug -DBUILD_DEBUG_WASM=ON -DBUILD_RUST=yes -DBUILD_DOC=no -DBUILD_JS=no -DBUILD_NATIVE=no -DCMAKE_CXX_COMPILER_LAUNCHER=ccache -DCMAKE_C_COMPILER_LAUNCHER=ccache -DCMAKE_INSTALL_PREFIX=\"psidk\" .. && make install",
+            "command": "mkdir -p build && cd build && cmake -DCMAKE_EXPORT_COMPILE_COMMANDS=ON -DCMAKE_BUILD_TYPE=Debug -DBUILD_DEBUG_WASM=ON -DCMAKE_CXX_COMPILER_LAUNCHER=ccache -DCMAKE_C_COMPILER_LAUNCHER=ccache -DCMAKE_INSTALL_PREFIX=\"psidk\" .. && make install",
```
swatanabe, line 21: This isn't different from the full build anymore.

James-Mart: Oh, good point. Technically it differs in that the full build uses CMAKE_BUILD_TYPE=Release, but I think I'll just delete Build Services anyways.

James-Mart (approved):

James-Mart:
FYI @brandonfancher @cool-ant @Velua 
After you update your branch with the changes in this PR, you will need to update your `.vscode/tasks.json` file because the one in your repo will be using these old flags.

You can just delete `.vscode/tasks.json`, and then copy and rename `.vscode/tasks.json.sample` (which is updated in this PR) to the new `.vscode/tasks.json`.

PR #589, opened 2024-02-15
Unify dueling RPC files (mjs vs. ts) and generate targets from TS

brandonfancher:


services/system/AccountSys/ui/tsconfig.json
```diff
@@ -2,15 +2,12 @@
...
-            "DOM",
-            "DOM.Iterable",
-            "ESNext"
-        ],
+        "lib": ["DOM", "DOM.Iterable", "ESNext"],
         "allowJs": false,
         "paths": {
-            "/common": [
-                "../../../user/CommonSys/common"
+            "/common": ["../../../user/CommonSys/common"],
```
cool-ant, line 8: This doesn't feel right. Is this for static assets like fonts? Wouldn't that be a common/public folder or something? given loader and rpc code are both in packages/ subdir?

brandonfancher: This was just a formatting change. Instead of digging into every app to see what _else_ they're consuming from the `common` directory, I focused more narrowly--on making sure they're not trying to consume the old `rpc.mjs` and `keyConversions.mjs`... and that they're instead consuming the new `rpc.ts`, `keyConversions.ts` (in dev) and `common-lib.js` (at runtime).

These other, less specific references to `common` will be cleaned up when we start getting these apps working again.

services/user/PsiSpaceSys/ui/vite.config-default-profile.ts
```diff
@@ -22,7 +22,9 @@ const psibase = (appletContract: string, isServing?: boolean) => {
     if (isServing) {
         buildAliases.push({
             find: "@psibase/common-lib",
-            replacement: path.resolve("../../CommonSys/common-lib/src"),
+            replacement: path.resolve(
+                "../../CommonSys/common/packages/rpc/src"
+            ),
         });
```
cool-ant, line 28: None of the other apps have this file, right? Did you ever figure out why this default-profile version exists? If not, I'd strip it out.

cool-ant:
So much solid cleanup! Thanks for being so thorough.
Deleted code is tested code! ;)

services/user/PsiSpaceSys/ui/vite.config-default-profile.ts
brandonfancher: That's in the territory of app cleanup. And I think that should happen when we work on getting these apps working again.

I'd like to keep this PR as narrowly focused on the RPC/keyConversions stuff as possible.

cool-ant (approved):

PR #591, opened 2024-02-16
Adding no-ancestor http header to apps (so they're guaranteed to be at _top)

cool-ant:


libraries/psibase/common/include/psibase/Rpc.hpp
```diff
@@ -11,7 +11,13 @@ namespace psibase
    {
       std::string name;
       std::string value;
+
+      friend bool operator==(const HttpHeader& lhs, const HttpHeader& rhs)
```
swatanabe, line 15: Use `= default`

services/system/ProxySys/src/ProxySys.cpp
```diff
@@ -79,10 +79,23 @@ namespace SystemService
       // TODO: avoid repacking (both directions)
       psibase::Actor<ServerInterface> iface(act.service, service);
 
+      auto reqTarget = req.target;
+
       auto result = iface.serveSys(std::move(req));
       if (result && !result->headers.empty() && serviceName != "common-sys")
          abortMessage("service " + service.str() + " attempted to set an http header");
 
+      if (reqTarget.size() == 0 || reqTarget.starts_with("/index.html"))
```
swatanabe, line 88: Do you really mean `starts_with`? `ends_with` or `==` would make more sense to me.

services/system/ProxySys/src/ProxySys.cpp
```diff
@@ -79,10 +79,23 @@ namespace SystemService
...
+      {
+         // Check if the target header is already in the vector; if the header was not found, add it
+         HttpHeader frameAncestorHeader = {"Content-Security-Policy", "frame-ancestors 'none';"};
+         auto pos = std::find(result->headers.begin(), result->headers.end(), frameAncestorHeader);
```
swatanabe, line 92: This will miss many valid ways to spell the header.
Also, if the service has already provided a different value of `frame-ancestors`, do we still want to add `frame-ancestors 'none'`?

James-Mart:
No longer needed.

PR #592, opened 2024-02-19
Dev/wit parsing

Velua:
Adds a new transpiled wasm we use to parse other WASM files to determine their imports / exports.


Velua:
Everything seems to work, we just need to get the `WASM` deployable to chain, where should we go from here? @James-Mart

James-Mart:
Let's plan to deploy to the chain the component-parser *component*. Meaning, we don't need CMake to run `jco transpile`, because we only have to upload the one `.wasm` file and then we transpile it in-browser after fetching it from the chain.

services/user/CommonSys/common-lib/packages/wasm-loader/src/main.ts
```diff
@@ -198,25 +160,37 @@ const parseFunctions = (data: string): ServiceMethodIndex => {
...
-                )
-                .join("\n")
-        };
-    });
-};
+    id: string,
+    functionsResult: FunctionResult[]
+): { [key: string]: string }[] =>
+    Object.entries(serviceMethodIndex).map(([key, methodNames]) => ({
+        [`component:${service}/${key}`]: methodNames
```
James-Mart: The component-parser component I made was "psibase:component-parser".
An example wasi component looks like:  "wasi:io". And it's possible to only import parts of a component by using slash notation, such as `wasi:io/error` if I only want the `error` object out of the `wasi:io` component.

In WIT, they refer to these fields in the following way:
`namespace:package@semver/name`

To generalize,
* `namespace` = service
* `package` = which plugin
* `@semver` is optional, possibly forbidden for psibase packages
* `/name` = name of the item being imported from that component. If left out, the entire component is imported.

PR #595, opened 2024-02-19
add component parser

James-Mart:
This is a stripped-down version of the graph component that simply parses components into component objects and wit files.
It should be able to be transpiled using `jco` into either a core wasm or js in order to use it in the browser.


James-Mart:
> This CI/CD build should succeed if we re-trigger it after [this PR](https://github.com/gofractally/psibase/pull/543) gets merged.

It was merged, retriggering CI/CD now.

PR #596, opened 2024-02-20
improve service out of memory error message

James-Mart:


libraries/psibase/service/src/simple_malloc.cpp
```diff
@@ -54,7 +54,7 @@ namespace eosio
          {
             size_t new_next_page = align(next_addr, wasm_page_size);
             if (GROW_MEMORY((new_next_page - next_page) >> 16) == -1)
-               psibase::raw::abortMessage("failed to allocate pages", 24);
+               psibase::raw::abortMessage("service exceeded its memory limit", 24);
```
swatanabe, line 57: This will print "service exceeded its mem"

James-Mart: slightly ironic

James-Mart: fixed

swatanabe (approved):

PR #597, opened 2024-02-20
Change checkAuthSys signature to avoid passing the potentially large data, which is never actually used.

swatanabe:
This should increase the effective maximum action size, though it doesn't fully resolve it.

### DIFF
--- a/services/system/ProducerSys/include/services/system/ProducerSys.hpp
+++ b/services/system/ProducerSys/include/services/system/ProducerSys.hpp
@@ -47,6 +47,6 @@ namespace SystemService
    PSIO_REFLECT(ProducerSys,
                 method(setConsensus, consensus),
                 method(setProducers, producers),
-                method(checkAuthSys, flags, requester, action, allowedActions, claims),
+                method(checkAuthSys, flags, requester, sender, action, allowedActions, claims),
                 method(canAuthUserSys, user))
 }  // namespace SystemService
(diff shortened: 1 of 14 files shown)


services/system/ProducerSys/include/services/system/ProducerSys.hpp
```diff
@@ -38,7 +38,8 @@ namespace SystemService
       // Allows this service to be used as an auth service for `prods-weak` and `prods-strong`.
       void checkAuthSys(uint32_t                    flags,
                         psibase::AccountNumber      requester,
-                        psibase::Action             action,
+                        psibase::AccountNumber      sender,
+                        ServiceMethod               action,
```
James-Mart, line 42: Missing the corresponding reflection update

services/system/TransactionSys/include/services/system/TransactionSys.hpp
James-Mart, line 1: Are you planning to provide a `getAction` action in case an auth service needs the data?

James-Mart (changes requested):

services/system/TransactionSys/include/services/system/TransactionSys.hpp
swatanabe: I couldn't decide what to call it (the name `getAction` sounds like it should be behave like `psibase::getCurrentAction`), and I'm not sure that it will ever be needed, so I deferred it for now.

James-Mart:
Haha
```
/home/runner/work/psibase/psibase/libraries/psibase/native/tests/WatchdogTests.cpp:134: FAILED:
  CHECK( failed <= allowed_failures )
with expansion:
  6 <= 5
```

James-Mart (approved):

PR #601, opened 2024-02-21
Deploy component-parser with CommonSys.psi

James-Mart:
This moves the location of the generated component-parser component to `wasm-loader/dist/component_parser.wasm`. The `dist/*` directory is already included in the CommonSys.psi package, and therefore the component is deployed to the chain at boot.


James-Mart:
Probably makes sense to merge this, and then you can fix the issue in main.ts on the `dev/wit-parsing` branch

PR #603, opened 2024-02-21
Dev/plugins

James-Mart:


James-Mart:
This PR got us past the build environment's lack of `cargo component`, but it exposed another issue with the build environment. 
```
error[E0463]: can't find crate for `core`
2024-02-21T16:28:18.3446507Z   |
2024-02-21T16:28:18.3447450Z   = note: the `wasm32-unknown-unknown` target may not be installed
2024-02-21T16:28:18.3448853Z   = help: consider downloading the target with `rustup target add wasm32-unknown-unknown`
```

The `wasm32-unknown-unknown` target needs to be installed in order to build components that don't need wasi interfaces.
I've added a PR to create yet another builder image, but it's going to take a little while for this change to make it's way to `dev/plugins`.

James-Mart:
Will reopen this based on the new dev/plugins after [this PR](https://github.com/gofractally/psibase/pull/604) merges

PR #605, opened 2024-02-21
Get the latest from dev/plugins

James-Mart:


James-Mart:
tl;dr:

It's okay to merge this even with the CI/CD error. We will not merge `wit-parsing` back into `dev/plugins` until the ci/cd error is fixed.

--------



This PR exposed another issue with the build environment. 
```
error[E0463]: can't find crate for `core`
2024-02-21T16:28:18.3446507Z   |
2024-02-21T16:28:18.3447450Z   = note: the `wasm32-unknown-unknown` target may not be installed
2024-02-21T16:28:18.3448853Z   = help: consider downloading the target with `rustup target add wasm32-unknown-unknown`
```

The `wasm32-unknown-unknown` target needs to be installed in order to build components that don't need wasi interfaces.
I've [added a PR](https://github.com/gofractally/image-builders/pull/25) to create yet another builder image, but it's going to take a little while for this change to make it's way to `dev/plugins`.

PR #607, opened 2024-02-21
Move loader code into common/packages/wasm-loader

brandonfancher:


cool-ant (commented):
I'd suggest moving current denizens of the common/ folder to something like common/public so that common/ isn't mixing static resources like fonts with built resources like the contents of common/packages.

brandonfancher:
> I'd suggest moving current denizens of the common/ folder to something like common/public so that common/ isn't mixing static resources like fonts with built resources like the contents of common/packages.

Sounds good. As discussed, in the directory tree, we'll have `/common/packages` and `/common/public`.

cool-ant (approved):

PR #610, opened 2024-02-22
Common Cleanup: Messaging Lib

brandonfancher:
- Moves the messaging code (currently in flux) into the `common-lib library`.
- Apps can now import from `@psibase/common-lib` or `@psibase/common-lib/(rpc|keyConversions|messaging)`. This is for convenience and code organization more than anything. On chain, apps will reference common-lib from the version deployed by CommonSys on chain. (And most are currently set up that way already.)


brandonfancher:
This is another case wherein the build is failing here but not locally. 🤷‍♂️ I'll investigate further tomorrow and merge, assuming approval in the meantime.

Velua (approved):
LGTM! :fire:

PR #615, opened 2024-02-26
Add logging of transaction and query console output

swatanabe:
The default format shows it indented 4 spaces. The output is controlled by the service author and is not sanitized. Beware if sending it to an automated consumer.


James-Mart:
> The output is controlled by the service author and is not sanitized. Beware if sending it to an automated consumer.

Seems like this warning should be in the `logging.md` docs

James-Mart (approved):

PR #628, opened 2024-03-01
Add WIT parsing

Velua:
Allows user to define an external plugin in a plugins WIT File 

E.g. DemoApp1 
```
interface demoapp2 {
    callintoplugin: func() -> string;
}


    fn helloworld() -> String {
        let barry = demoapp2::callintoplugin();
        format!("barry is {}", barry)
    }
```

The supervisor will then parse this WIT file, recognize the mention of an external service and it's function, mock it and run the plugin, should the plugin call the mocked function, the supervisor will initialize the foreign service and run it again. 

Previous implementation of this required on hardcoded values, this is now dynamic.


services/user/CommonSys/common/packages/wasm-loader/src/main.ts
```diff
@@ -24,68 +21,53 @@ interface Importables {
...
 
 const functionCall = async ({
     id,
     args,
     precomputedResults
 }: PluginCallPayload) => {
-    const url = "/plugin.wasm";
+    const [pluginBytes, parserBytes, wasmComponent] = await Promise.all([
+        wasmUrlToIntArray("/plugin.wasm"),
+        wasmUrlToIntArray("/common/wasm-loader/component_parser.wasm"),
```
brandonfancher, line 36: Where does this file come from? When I build and run, I'm getting a 404 on this file.

brandonfancher (commented):
Can I get a walkthrough of all of this from you...start shoring up my understanding of all of this?

services/user/DemoApp1/ui/src/App.tsx
```diff
@@ -21,8 +21,8 @@ function App() {
 
     console.log("calling demoapp2.callintoplugin");
     const res = await supervisor.functionCall({
-      service: "demoapp2",
-      method: "callintoplugin",
+      service: "demoapp1",
```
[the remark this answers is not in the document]
Velua: Because we're demonstrating the ability that we can call `demoapp1` which then calls `demoapp2` from Rust / plugin

services/user/CommonSys/common/packages/wasm-loader/src/main.ts
James-Mart: It's supposed to be a build artifact, but @Velua you somehow lost the project that generates the wasm when you migrated into this new branch. I [still see it](https://github.com/gofractally/psibase/pull/592/files#diff-8b1fea070d00786d2ef9c890d5b628b4e657ab912bff0675d8436188ad9037d8) in the branch in your other open PR

James-Mart: @brandonfancher and @Velua.
I've added the component-parser in [a separate PR](https://github.com/gofractally/psibase/pull/638). Once merged, it will be available on chain by default.

@Velua it implies two changes you need to make this pr:
1. Undo the CMakeLists.txt changes
2. Anything that attempts to load the component_parser from chain should look for it at: `/common/component_parser.wasm`. It's not actually coupled to the loader, hence the separation.

Velua:
I've messed up my rebasing on this branch, closing this PR in favour of #639

Issue #627, opened 2024-02-29
CICD build race condition building wasm-loader ui files


brandonfancher:
There is a race condition intermittently impacting only Ubuntu 2004 builds on CICD on the `dev/plugins` branch and any branch off of that. I'm not able to reproduce this locally.

<img width="1028" alt="image" src="https://github.com/gofractally/psibase/assets/7911424/fb0526aa-e52c-48c6-8a9c-d2651c368097">

brandonfancher:
Discovered in #626

swatanabe:
Does `common-lib` need to build before `wasm-loader`?

brandonfancher:
@swatanabe Yes, that's exactly the problem.

brandonfancher:
Need to tackle this. Tagging #561 too.

brandonfancher:
Soon, we'll make `wasm-loader` depend on `common-lib` as an external ES Module resource (already being served by CommonSys.) But for right now, as we're in active development of `wasm-loader`, it makes more sense to bundle it. Hence `wasm-loader`'s dependency on `common-lib`.

brandonfancher:
#634 too.

James-Mart:
This should be fixed [here](https://github.com/gofractally/psibase/pull/634/commits/1558448e7dad5fa5e9578185e1fceb7236db5cd4).

PR #634, opened 2024-03-01
cmakelists fixes and clarity improvements

James-Mart:


CMakeLists.txt
```diff
@@ -210,39 +210,36 @@ set(ADMIN_DIR services/user/AdminSys/ui:AdminSys_js)
 set(COMMON_LIB_DIR services/user/CommonSys/common/packages/common-lib/:CommonSysCommonLib_js)
 set(COMMON_LOADER_DIR services/user/CommonSys/common/packages/wasm-loader/:CommonSysLoaderLib_js)
 
-function(add_js_target ENTRY OUTPUT_FILE)
-    string(REGEX REPLACE "^([^:]+):([^:]+)$" \\1 PATH ${ENTRY})
-    string(REGEX REPLACE "^([^:]+):([^:]+)$" \\2 DEP ${ENTRY})
+function(add_js_target ROW OUTPUT_FILE)
+    string(REGEX REPLACE "^([^:]+):([^:]+)$" \\1 PATH ${ROW})
+    string(REGEX REPLACE "^([^:]+):([^:]+)$" \\2 TargetName ${ROW})
```
brandonfancher, line 215: Why `TargetName` and not `TARGET_NAME` like other variable names? Is it just a hint that the values are in PascalCase? (`CommonSys_js`, `TokenSys_js`, etc.)?

CMakeLists.txt
```diff
@@ -210,39 +210,36 @@ set(ADMIN_DIR services/user/AdminSys/ui:AdminSys_js)
...
+    set(OUTPUT_FILEPATH ${CMAKE_CURRENT_SOURCE_DIR}/${PATH}/dist/${OUTPUT_FILE})
```
brandonfancher, line 216: What did we decide on the merits of hard-coding `/dist` in this?

CMakeLists.txt
```diff
@@ -210,39 +210,36 @@ set(ADMIN_DIR services/user/AdminSys/ui:AdminSys_js)
...
 endfunction()
 
-list(TRANSFORM JS_DIRS REPLACE "^[^:]+:([^:]+)$" \\1 OUTPUT_VARIABLE JS_DEPS)
-file(GLOB common LIST_DIRECTORIES false ${CMAKE_CURRENT_SOURCE_DIR}/services/user/CommonSys/common/resources/*)
+# Static (not built) resource dependencies
+file(GLOB common-misc-resources LIST_DIRECTORIES false ${CMAKE_CURRENT_SOURCE_DIR}/services/user/CommonSys/common/resources/*)
 file(GLOB common-fonts LIST_DIRECTORIES false ${CMAKE_CURRENT_SOURCE_DIR}/services/user/CommonSys/common/resources/fonts/*)
 
-# TODO: Can this be a dist folder, i.e., a build product?
-file(GLOB common-loader LIST_DIRECTORIES false ${CMAKE_CURRENT_SOURCE_DIR}/services/user/CommonSys/common/packages/wasm-loader/dist/*)
```
brandonfancher, line 236: I take it we have confirmation that loader builds without this superfluous line, eh?

brandonfancher (approved):

CMakeLists.txt
James-Mart: Yeah it's sort of a convention that target names are PascalCase. My thinking was that, even though it's a variable, it contains the name of a target and so I wanted to indicate that by giving the variable PascalCase.

But... upon reflection, I don't actually like this. It's a variable, should be caps.

CMakeLists.txt
James-Mart: If anything the intention here was that other things would correctly be dependent on the loader, but it's being done incorrectly. I used the _DEP variable to ensure things properly depend on the loader.

CMakeLists.txt
James-Mart: Works for now until we have output files that don't live in a `dist` directory. At which point we can change the parameters to OUTPUT_FILE_SUBPATH and pass in `dist/index.html` etc.

CMakeLists.txt
brandonfancher: I see. And it's a few lines later where we build the loader (240/237). Cool.

CMakeLists.txt
James-Mart: fixed

James-Mart:
Closes https://github.com/gofractally/psibase/issues/627

PR #638, opened 2024-03-03
Adds the component parser to CommonSys.psi package

James-Mart:
Retrievable on a default chain at `/common/component_parser.wasm`.

In order for this to build, the version of cargo-component must be >= 0.9.0. Anyone who already has a psibase build environment probably already has cargo-component, and it won't update automatically, **so you have to run `cargo install cargo-component@0.9.0`.**

### CICD fail note

CICD will fail because when the image for CICD was built, version 0.9.0 of `cargo-component` had not come out yet. I propose that we merge this instead of waiting for an updated CICD. Local builds should work as long as you followed the manual update steps above.

If we want to wait, the new version of the CICD image is already PRed to main in [this PR](https://github.com/gofractally/psibase/pull/637). After that gets merged, we need to merge `main`->`dev/plugins` again.


Velua (approved):

James-Mart:
I'm going to go ahead and merge this, hopefully @brandonfancher is okay with my temporarily breaking CI/CD in dev/plugins. The PRs that fix it again will likely go through Monday morning.

PR #652, opened 2024-03-06
Removing FractalSys

cool-ant:
FractalSys will need rebuliding, and it's now 2nd priority to core psibase work.
Also, a big part of the work now lives in InviteSys.
So we're removing it for now until we rebuild it.


cool-ant:
@James-Mart and I discussed this, and he'd like this left in until he rewrites it in Rust

Issue #651, opened 2024-03-06
SessionID tracking in supervisor infra


James-Mart:
Unless I'm missing something, I'm not seeing anything in `onPluginCallResponse` and `onPluginCallFailure` that checks for any kind of session ID from the sender.
Checking the origin of a response message is sufficient for security, but checking a session ID is needed to disambiguate between messages posted to the window in the case that the app is open in multiple tabs.

James-Mart:
More thoughts on this feature:

Any time a parent window creates a child window, the ID of the session cannot be passed in a message from parent->child. This is because, from the perspective of the child, it can validate any message it receives by domain, but to my knowledge it cannot know that the message is from its parent. 

A window does know its parent via [window.parent](https://developer.mozilla.org/en-US/docs/Web/API/Window/parent).

...We need a general strategy for synchronizing parent and child windows.

Velua:
Currently investigating window behaviour when it comes to window post messages, @brandonfancher to provide a 2nd opinion. So far we're seeing messages remain private between windows.

brandonfancher:
Second opinion here:

Both my research (documentation, stack overflow, etc.) and my own testing confirms that a window may only send a message to a target it already has a reference to. As described [here](https://developer.mozilla.org/en-US/docs/Web/API/Window/postMessage):

> Broadly, one window may obtain a reference to another (e.g., via `targetWindow = window.opener`), and then dispatch a [MessageEvent](https://developer.mozilla.org/en-US/docs/Web/API/MessageEvent) on it with `targetWindow.postMessage()`. The receiving window is then free to [handle this event](https://developer.mozilla.org/en-US/docs/Web/Events/Event_handlers) as needed.

Velua:
Thanks @brandonfancher I'll close this now https://github.com/gofractally/psibase/pull/654

PR #654, opened 2024-03-08
check message ids on supervisor inbound messages

Velua:
In response to #651 this adds an early return when handling plugin call success and error responses making an early return if not found.

### DIFF
--- a/services/user/SupervisorSys/ui/src/main.ts
+++ b/services/user/SupervisorSys/ui/src/main.ts
@@ -178,6 +178,12 @@ const checkForResolution = () => {
 
 const onPluginCallResponse = (message: PluginCallResponse) => {
   const id = message.payload.id;
+  const isIdTracked = !!findFunctionCallById(pendingFunctionCalls, id)
+  if (!isIdTracked) {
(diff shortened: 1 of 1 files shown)


James-Mart:
Taking another look at this feature, I don't think this code is sufficient, actually. Any time a parent window creates a child window, the ID of the session cannot be passed in a message from parent->child. This is because, from the perspective of the child, it can validate any message it receives by domain, but to my knowledge it cannot know that the message is from its parent. I believe when a child window is created, the first thing it should do is ask its parent (which it knows via [window.parent](https://developer.mozilla.org/en-US/docs/Web/API/Window/parent)) for the session ID. Only after it knows this sessionID can it listen for messages sent to it with Window.PostMessage.

Unless I'm missing something, the current loader code also does not check any ID right now when it gets a function call. It too must get the session ID from its parent supervisor upon creation, and then only listen for function calls from its parent supervisor.

Velua:
@James-Mart on it.

James-Mart:
Might still be good to merge this for general error handling, even though it can't happen during happy path execution

PR #672, opened 2024-03-28
Additional account name restrictions

James-Mart:
Hyphens cannot be in the first or last positions of an account.
* Cannot be in the first or last position because account names are intended to be subdomains, and that is part of the DNS rules.

Additionally, an account name may not start with `x-`. This prefix is reserved for use by the infra providers, intended so that there is a namespace at which subjective apps can be served from the node without fear of shadowing real user accounts/apps.


services/system/AccountSys/src/AccountSys.cpp
```diff
@@ -55,20 +55,26 @@ namespace SystemService
...
          writeConsole("new acc: ");
-         writeConsole(name.str());
+         writeConsole(strName);
          writeConsole("auth con: ");
          writeConsole(authService.str());
       }
 
       check(name.value, "invalid account name");
+      check(strName.back() != '-', "account name must not end in a hyphen");
+      if (strName.size() > 2 && sender != service)
```
swatanabe, line 71: `starts_with` checks the size.

CMakeLists.txt
```diff
@@ -359,9 +359,9 @@ psibase_package(
     DESCRIPTION "This service facilitates the creation of new accounts"
     SERVICE account-sys
         WASM ${CMAKE_CURRENT_BINARY_DIR}/AccountSys.wasm
-        SERVER r-account-sys
+        SERVER raccount-sys
```
swatanabe, line 362: Renaming shouldn't be needed anymore, right?

services/system/AccountSys/src/AccountSys.cpp
```diff
@@ -55,20 +55,26 @@ namespace SystemService
...
+      {
+         check(!strName.starts_with("x-"), "'x-' is a reserved prefix");
```
swatanabe, line 73: Maybe add who it is reserved for, for clarity.

swatanabe (approved):

services/system/AccountSys/src/AccountSys.cpp
James-Mart: As in, for subjective services? Technically it could really be for anything agreed upon by prods, though.

swatanabe (approved):

PR #674, opened 2024-03-28
Remove applet related libraries

James-Mart:
Applets in iframes are no longer needed with the new psibase plugin architecture. Basically all of our old apps were based on this model. This PR removes references to applets and iframeresizer. To fully remove applet-management-related items, I had to effectively wipe the entire home page.


brandonfancher (approved):
LGTM! Glad you're removing the UI components as part of this PR. I have them removed in the draft NftSys PR, but makes more sense removing them as part of a cleanup PR.

PR #678, opened 2024-04-02
Remove outdated flags from build script

James-Mart:
Sparky pointed out that the build script wasn't updated since
https://github.com/gofractally/psibase/pull/569


sparkplug0025 (approved):
🎉

PR #683, opened 2024-04-05
CMake fixes

swatanabe:
- The dist directory is marked as an input for other rules, so ninja needs to know that it is created.
- Make the component copy conditional, to avoid triggering spurious rebuilds
- `BUILD_BYPRODUCTS` refers to the build step, not the install step.


James-Mart (approved):

James-Mart:
> Added in 3.26

Should we just update minimum cmake version?

swatanabe:
> > Added in 3.26
> 
> Should we just update minimum cmake version?

22.04 has 3.22. I don't think we should increase past that yet.

James-Mart (approved):

PR #685, opened 2024-04-05
increase test boot timeout

swatanabe:
I've seen some random failures in CICD.


swatanabe:
Nevermind, that's not the problem:
```
service 'transact' aborted with message: service exceeded its memory limit
```

PR #686, opened 2024-04-08
Reduce memory usage of processTransaction

swatanabe:


libraries/psibase/common/include/psibase/dispatch.hpp
```diff
@@ -1,16 +1,53 @@
...
    template <typename R, typename T, typename MemberPtr, typename... Args>
    void callMethod(T& service, MemberPtr method, Args&&... args)
    {
-      psio::shared_view_ptr<R> p((service.*method)(std::forward<decltype(args)>(args)...));
-      raw::setRetval(p.data(), p.size());
+      if constexpr (is_view_v<R>)
+      {
+         R    v((service.*method)(std::forward<decltype(args)>(args)...));
+         auto s = find_view_span(v);
+         check(s.data() != nullptr, "Cannot handle extensions in returned view");
```
James-Mart, line 43: Does this imply that objects that are return values from synchronous action calls cannot be extended?

James-Mart (approved):

libraries/psibase/common/include/psibase/dispatch.hpp
swatanabe: The service can safely be modified to return a new type that extends the original type. It cannot return fields that it is unaware of using this API.

PR #689, opened 2024-04-17
Add test service

Velua:


rust/psibase/src/services/demoapp1.rs
James-Mart, line 1: This should be deleted: we don't want `demoapp1` to show up in the `psibase::services` interface for everyone

dev/DemoApp1/plugin/src/lib.rs
```diff
@@ -1,11 +1,14 @@
...
 mod bindings;
 
-use bindings::common::plugin::types as CommonTypes;
+use bindings::common::plugin::{server, types as CommonTypes};
 use bindings::exports::demoapp1::example::intf::Guest as Intf;
 use bindings::invite;
 use bindings::Guest as MainInterface;
 
+use psibase::fracpack::Pack;
+use psibase::services::demoapp1 as app_1;
```
James-Mart, line 10: demoapp1 should just be a separate service, not taken from psibase::services

dev/DemoApp1/plugin/src/lib.rs
```diff
@@ -18,6 +21,15 @@ impl Intf for Component {
     fn helloworld2() -> Result<String, CommonTypes::Error> {
         Ok(invite::plugin::inviter::generate_invite("/subpath")?)
     }
+
+    fn multipli(a: u32, b: u32) -> String {
```
James-Mart, line 25: Why not `multiply`?

dev/DemoApp1/plugin/src/lib.rs
```diff
@@ -18,6 +21,15 @@ impl Intf for Component {
     fn helloworld2() -> Result<String, CommonTypes::Error> {
```
James-Mart, line 21: you could also delete helloworld and helloworld2 if you want, just to clean up the example app

dev/DemoApp1/plugin/wit/world.wit
```diff
@@ -3,11 +3,16 @@ package demoapp1:example;
 interface intf {
     use common:plugin/types.{error};
     helloworld2: func() -> result<string, error>;
+    multipli: func(a: u32, b: u32) -> string;
```
James-Mart, line 6: I think you should return `result<string, error>`. I expect almost every function in every plugin to return `result<something, error>`.

dev/demoapp1-build-and-deploy.sh
James-Mart, line 1: At some point this should also be deleted. We can just set up demoapp1 in cmake so it's not deployed separately. Should also help make sure we don't break demoapp1 when its dependencies change (if we change common plugin interfaces, and such, for example).

James-Mart (changes requested):

dev/demoapp1-build-and-deploy.sh
Velua: :+1:

Velua:
@James-Mart I've made those changes except the Rust wrapper import, I can't crack that... will try get on a call and sus it.

dev/DemoApp1/plugin/wit/world.wit
```diff
@@ -3,11 +3,16 @@ package demoapp1:example;
 interface intf {
     use common:plugin/types.{error};
     helloworld2: func() -> result<string, error>;
+    multiply: func(a: u32, b: u32) -> result<string, error>;
```
James-Mart, line 6: Bit odd for `multiply` to return a string

James-Mart:
Feel free to merge before you start getting merge conflicts

PR #707, opened 2024-05-14
misc doc updates

James-Mart:


doc/psidk/src/specifications/app-architecture/plugins.md
```diff
@@ -64,11 +64,9 @@ According to the [Window.postMessage documentation](https://developer.mozilla.or
 
 Therefore `postMessage` does not immediately (synchronously) post to the other domain. Instead it schedules a payload to be dispatched after the completion of all remaining execution contexts. In order for cross-domain calls to appear synchronous, the caller should await the asynchronous call.
 
-## Cross-domain security considerations
+## Plugin isolation
 
-Cross-domain messaging can be dangerous if the proper checks are not in place to ensure that the messages are going to/from whoever is intended. 
-
-The supervisor listens for cross-domain messages from its parent window and from the plugins it instantiates, and plugins will only listen for messages that come directly from the supervisor.
+It is important that plugins are sandboxed on the client-side. To communicate with each other, they should use a middleware layer that enforces the identity of the caller is accurately reported to the callee, allowing identity-based permissions. This middleware is called the supervisor. The supervisor listens for cross-domain messages from its parent window and from the plugins it instantiates, and plugins will only listen for messages that come directly from the supervisor.
```
swatanabe, line 69: "...enforces *that* the identity of..."

Also, this paragraph is somewhat redundant with the next paragraph. If we want to keep the separation, this paragraph should be normative rather than descriptive. i.e. "plugins MUST only listen for messages that come directly from the supervisor."

James-Mart: Reworked a bit

swatanabe (approved):

PR #708, opened 2024-05-14
improve misc error messages

James-Mart:


James-Mart:
had this in an old stash, might as well deliver it

libraries/psibase/native/src/ExecutionContext.cpp
```diff
@@ -51,7 +51,7 @@ namespace psibase
       auto codeHash = sha256(code.pos, code.remaining());
 
       auto account = database.kvGet<CodeRow>(CodeRow::db, codeKey(service));
-      check(account.has_value(), "setCode: unknown service account");
+      check(account.has_value(), "setCode: unknown service account: " + service.str());
```
swatanabe, line 54: It shouldn't be possible to trigger this.  This `setCode` is only run by the genesis action, which unconditionally creates the `CodeRow` first.

libraries/psibase/native/src/pkcs11.cpp
```diff
@@ -1173,7 +1173,8 @@ namespace psibase::pkcs11
 
    URI::URI(std::string_view uri)
    {
-      check(parseLiteral(uri, "pkcs11:"), "Expected PKCS #11 URI");
+      check(parseLiteral(uri, "pkcs11:"),
+            "Invalid PKCS #11 URI, expected 'pkcs11:...', got: '" + std::string(uri) + "'");
```
swatanabe, line 1177: The error message is always constructed, even if the check passes.

libraries/psibase/native/src/ExecutionContext.cpp
James-Mart: I remember getting it when I removed the -sys suffixes, and it was from one of the packages having a postinstall script that had a hex-encoded deprecated service name. This was needed to figure out where to look.

swatanabe: That should have caused an error at L196, not here.

James-Mart: I'm not in a position to try to reproduce the error to confirm or disconfirm, would you prefer I
1. Add the `service.str()` output to L196, or
2. Replace this change with one at L196

swatanabe: I don't really care whether there's a change here. It should only be triggered by changes to native, but it doesn't really hurt. I only really care that if you're adding extra info, it should be added in the place that can be reached by client code.

James-Mart: I did option 2

swatanabe (approved):

PR #713, opened 2024-05-20
Improve token service graphql queries

James-Mart:
This adds a new capability to graphql queries in C++: A vector of objects dynamically constructed at query time can be more easily queried using graphql.

The tokens query service uses this to allow for pagination on a dynamically constructed vector of objects, rather than returning the entire vector.

-------

New queries:
1. `userBalances(AccountNumber user, ...pagination)`
2. `userCredits(AccountNumber user, ...pagination)`
3. `userDebits(AccountNumber user, ...pagination)`
4. `tokenDetails(TID tokenId)`

The first three queries manually construct an index that combines the normal balance table row with details about the token (symbolID and precision) in order to return all relevant details within a single query.

The fourth new query `tokenDetails` returns details about a token given its ID, including the current owner of the issuer NFT, and the auth-service set by that owner account.

### DIFF
--- a/libraries/psibase/common/include/psibase/serveGraphQL.hpp
+++ b/libraries/psibase/common/include/psibase/serveGraphQL.hpp
@@ -349,17 +349,14 @@ namespace psibase
       return result;
    }  // makeConnection
 
-   template <typename T, typename Key>
-   concept HasGetKey = requires(T t) {
-      {
-         t.getKey()
-      } -> std::same_as<Key>;
-   };
-
    /// Similar to makeConnection, except that it allows pagination through a virtual table index.
+   ///
    /// A virtual table index is a dynamically constructed vector of objects.
+   ///
+   /// Allows for a user-provided projection function that transforms elements from type T to Key
+   /// before isolating a range within the container.
+   ///
    template <typename Connection, typename T, typename Key>
-      requires HasGetKey<T, Key>
    Connection makeVirtualConnection(const std::vector<T>&             elements,
                                     const std::optional<Key>&         gt,
                                     const std::optional<Key>&         ge,
@@ -368,7 +365,8 @@ namespace psibase
                                     std::optional<uint32_t>           first,
                                     std::optional<uint32_t>           last,
                                     const std::optional<std::string>& before,
-                                    const std::optional<std::string>& after)
+                                    const std::optional<std::string>& after,
+                                    std::function<Key(T)>             projection = {})
    {
       auto indexFromString = [&](const std::string& str, size_t& index) -> bool
       {
@@ -383,21 +381,20 @@ namespace psibase
       };
 
       auto lower_bound = [&](const Key& key)
-      {
-         return std::lower_bound(elements.begin(), elements.end(), key,
-                                 [&](const T& elem, const Key& key)
-                                 { return elem.getKey() < key; });
-      };
+      { return std::ranges::lower_bound(elements, key, {}, projection); };
 
       auto upper_bound = [&](const Key& key)
-      {
-         return std::upper_bound(elements.begin(), elements.end(), key,
-                                 [&](const Key& key, const T& elem)
-                                 { return key < elem.getKey(); });
-      };
+      { return std::ranges::upper_bound(elements, key, {}, projection); };
 
       auto rangeBegin = elements.begin();
       auto rangeEnd   = elements.end();
+      if (ge || gt || le || lt)
+      {
+         if (!projection)
+         {
+            psibase::check(false, "makeVirtualConnection: missing projection function");
+         }
+      }
       if (ge)
          rangeBegin = std::max(rangeBegin, lower_bound(*ge));
       if (gt)
--- a/libraries/psibase/common/include/psibase/serveGraphQL.hpp
+++ b/libraries/psibase/common/include/psibase/serveGraphQL.hpp
@@ -355,7 +355,6 @@ namespace psibase
    ///
    /// Allows for a user-provided projection function that transforms elements from type T to Key
    /// before isolating a range within the container.
-   ///
    template <typename Connection, typename T, typename Key>
    Connection makeVirtualConnection(const std::vector<T>&             elements,
                                     const std::optional<Key>&         gt,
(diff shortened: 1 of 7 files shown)


libraries/psibase/common/include/psibase/serveGraphQL.hpp
```diff
@@ -349,6 +349,120 @@ namespace psibase
...
+         unsigned long result = std::strtoul(str.c_str(), &end, 10);
+         if (end == str.c_str() || *end != '\0' || result == std::numeric_limits<size_t>::max())
+         {
+            return false;
+         }
+         index = static_cast<size_t>(result);
+         return true;
+      };
+
+      auto lower_bound = [&](const Key& key)
```
swatanabe, line 385: If you pass in a proj function (`T -> Key`), instead of relying on `getKey`, then `std::ranges::lower_bound(elements, key, {}, proj)` will work.

services/user/Tokens/src/RTokens.cpp
```diff
@@ -13,64 +15,229 @@ using namespace UserService;
...
+      {
+         auto tokenOpt = tokenTypeIdx.get(balance.key.tokenId);
+         check(tokenOpt.has_value(), "Invalid token type");
+         balances.push_back(TokenBalance{tokenOpt->symbolId, balance.key.tokenId,
+                                         tokenOpt->precision, balance.balance});
+      }
+
+      return makeVirtualConnection<
+          Connection<TokenBalance, "UserBalanceConnection", "UserBalanceEdge">>(
+          balances, balance_gt, balance_ge, balance_lt, balance_le, first, last, before, after);
```
swatanabe, line 127: `gt`, `lt`, etc., only make sense if the data is sorted by balance.

services/user/XAdmin/ui/yarn.lock
James-Mart, line 1: Showed up on build

services/user/Tokens/src/RTokens.cpp
```diff
@@ -13,64 +15,229 @@ using namespace UserService;
...
+namespace TokenQueryTypes
 {
-   AccountNumber user;
-   uint64_t      balance;
-   uint8_t       precision;
-   TID           token;
-   SID           symbol;
-};
-PSIO_REFLECT(UserBalanceRecord, user, balance, precision, token, symbol);
+   auto realBalance = [](uint64_t bal, const Precision& p) -> uint64_t
```
swatanabe, line 21: Do we really want to ignore fractional tokens when comparing?

services/user/Tokens/src/RTokens.cpp
```diff
@@ -13,64 +15,229 @@ using namespace UserService;
...
-      return tokenService.index<BalanceTable, 0>();
+      return tokenService.open<BalanceTable>().getIndex<0>();
    }
-   auto userBalances(AccountNumber user) const
-   {  //
-      vector<UserBalanceRecord> balances;
-      for (auto tokenType : tokenService.index<TokenTable, 0>())
+
+   /// Provides a way to query all balances for a given user.
+   /// Further filtering can be done to restrict the results by balance.
```
swatanabe, line 100: I don't understand what it means to compare the balances of different tokens or how this comparison would be useful.

services/user/Tokens/include/services/user/tokenTables.hpp
```diff
@@ -92,10 +91,13 @@ namespace UserService
       BalanceKey key;
       uint64_t   balance;
 
+      auto byUserToken() const { return key.account; }
```
swatanabe, line 94: Secondary indexes need to be unique. The `Table` implementation does not implicitly append the primary key.

services/user/Tokens/src/RTokens.cpp
James-Mart: A similar function exists on multiple different eos block explorers, for example, because old accounts accumulate small numbers of junk tokens. But tbh it's really not that useful as a query on the server side. I think i'll just delete it. and delete the gt,lt,ge,le flags

services/user/Tokens/include/services/user/tokenTables.hpp
James-Mart: Ahh, I was wondering why we were using tuples elsewhere, I forgot that.

libraries/psibase/common/include/psibase/serveGraphQL.hpp
James-Mart: I like this. There's probably a more clever way to do it, but I just used std::function. That workable?

services/user/Tokens/src/RTokens.cpp
James-Mart: Deleted those filters

services/user/Tokens/src/RTokens.cpp
James-Mart: It was intentional, but... deleted.

services/user/Tokens/src/RTokens.cpp
James-Mart: Deleted

libraries/psibase/common/include/psibase/serveGraphQL.hpp
swatanabe: Yes. Or make it a template parameter and `requires std::is_invokable_r<Key, F, const T&>`

services/user/Tokens/src/RTokens.cpp
```diff
@@ -13,64 +15,212 @@ using namespace UserService;
...
 
+using namespace TokenQueryTypes;
 struct TokenQuery
 {
    auto allBalances() const
    {  //
-      return tokenService.index<BalanceTable, 0>();
+      return tokenService.open<BalanceTable>().getIndex<0>();
    }
-   auto userBalances(AccountNumber user) const
```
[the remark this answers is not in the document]
James-Mart: Even though it's probably still not stable, I'll probably leave the test pointing at this query. I think there's probably zero stable queries right now

libraries/psibase/common/include/psibase/serveGraphQL.hpp
James-Mart: If you're okay with std function, let's proceed as is, otherwise lmk and I'll work on this

services/user/Tokens/src/RTokens.cpp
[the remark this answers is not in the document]
James-Mart: Done

swatanabe (approved):

PR #714, opened 2024-05-20
Allow adding multiple event indices at the same time

James-Mart:


services/user/Events/src/Events.cpp
```diff
@@ -47,6 +47,47 @@ void EventIndex::addIndex(psibase::DbId          db,
...
+   auto secondary = Events::open<SecondaryIndexTable>(secondaryIndexSpecTableNum);
+   auto row       = secondary.getIndex<0>().get(std::tuple(db, service, event));
+   if (!row)
+      row = SecondaryIndexRecord{db, service, event, std::vector{SecondaryIndexInfo{}}};
+
+   for (auto column : columns)
+   {
+      check(column < type->children.size(), "Unknown column");
+
+      // Don't add duplicate indexes
```
swatanabe, line 67: I would use `sort` + `set_union` or at least convert the inner loop to `std::ranges::find` (`contains` in C++ 23).

services/user/Events/include/services/user/Events.hpp
```diff
@@ -62,6 +62,12 @@ namespace UserService
                     psibase::MethodNumber  event,
                     std::uint8_t           column);
 
+      // Same as `addIndex` except that multiple columns can be specified at the same time.
```
swatanabe, line 65: I'm not sure how useful this really is. There's still a lot of boilerplate using it. There's also no performance benefit to batching here, so I'd prefer to keep the service interface minimal.

I think a better approach would be to create a wrapper using reflection:
```
auto indexes = EventIndexCreator<Tokens>{}.history();
indexes.created().addIndices("tokenId", "creator");
```

James-Mart: I like this idea

PR #718, opened 2024-05-23
Reduce system token precision

James-Mart:
* Sets max precision to 8 instead of 16
* Updates the precision class so min/max are enforced when unpacking from frackpack or json
* Updates the precision of the system token to 4
* Updates a bunch of tests to look up the previously hardcoded sys token precision

### DIFF
--- a/services/user/Tokens/include/services/user/tokenTypes.hpp
+++ b/services/user/Tokens/include/services/user/tokenTypes.hpp
@@ -35,6 +35,11 @@ namespace UserService
    };
    PSIO_REFLECT(Precision, value);
 
+   inline bool clio_validate_packable(const Precision& p)
+   {
+      return Precision::validate(p.value);
+   }
+
    void from_json(Precision& p, auto& stream)
    {
       from_json_object(stream,
@@ -53,12 +58,7 @@ namespace UserService
                           }
                        });
 
-      Precision::validate(p.value);
-   }
-
-   inline bool clio_validate_packable(const Precision& p)
-   {
-      return Precision::validate(p.value);
+      clio_validate_packable(p);
    }
 
    struct Quantity
--- a/services/user/Tokens/include/services/user/tokenTypes.hpp
+++ b/services/user/Tokens/include/services/user/tokenTypes.hpp
@@ -40,27 +40,6 @@ namespace UserService
       return Precision::validate(p.value);
    }
 
-   void from_json(Precision& p, auto& stream)
-   {
-      from_json_object(stream,
-                       [&](std::string_view key) -> void
-                       {
-                          bool found = false;
-                          psio::reflect<Precision>::get(key,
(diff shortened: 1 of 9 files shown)


services/user/Tokens/include/services/user/tokenTypes.hpp
```diff
@@ -12,26 +12,55 @@ namespace UserService
...
+                           std::to_string(precisionMin) + " to " + std::to_string(precisionMax));
+         }
+         return true;
       }
 
       friend std::strong_ordering operator<=>(const Precision&, const Precision&) = default;
    };
    PSIO_REFLECT(Precision, value);
 
+   void from_json(Precision& p, auto& stream)
```
James-Mart, line 38: This is essentially a simplified version of the struct unpacking [here](https://github.com/gofractally/psibase/blob/main/libraries/psio/include/psio/from_json.hpp#L745).

services/user/Tokens/include/services/user/tokenTypes.hpp
```diff
@@ -12,26 +12,55 @@ namespace UserService
...
+                                                           from_json(p.*member, stream);
+                                                           found = true;
+                                                        });
+                          if (not found)
+                          {
+                             from_json_skip_value(stream);
+                          }
+                       });
+
+      Precision::validate(p.value);
```
swatanabe, line 56: I think we'd be better off making `from_json` call `clio_validate_packable`, to get unified validation for deserialization.

James-Mart: done

swatanabe: Not quite what I meant. I meant making the generic from_json call `clio_validate_packable` if the call is well-formed.

services/user/Tokens/include/services/user/tokenTypes.hpp
```diff
@@ -12,26 +12,34 @@ namespace UserService
 
    struct Precision
    {
-      static constexpr uint8_t          precisionMin = 0;
-      static constexpr uint8_t          precisionMax = 16;
-      static constexpr std::string_view errorInvalid = "Value exceeds allowed Precision range";
+      static constexpr uint8_t precisionMin = 0;
+      static constexpr uint8_t precisionMax = 8;
```
swatanabe, line 16: I don't understand the reasoning for choosing a particular value of `precisionMax`. Is there some way that 16 causes a problem?

swatanabe (approved):

services/user/Tokens/include/services/user/tokenTypes.hpp
James-Mart: It's just for ergonomics. Can be increased later if there's actually a need for >8 decimals of precision.

PR #719, opened 2024-05-23
Add `SplitKey` type for subindexing in token query service

James-Mart:
* Adds the `SplitKey` type to the token query service to allow subindexing on indices with a struct key
* Eliminates redundant token table indices

### DIFF
--- a/libraries/psibase/common/include/psibase/Table.hpp
+++ b/libraries/psibase/common/include/psibase/Table.hpp
@@ -6,8 +6,8 @@
 #include <concepts>
 #include <cstdint>
 #include <functional>
-#include <psibase/blob.hpp>
 #include <psibase/RawNativeFunctions.hpp>
+#include <psibase/blob.hpp>
 #include <psibase/serviceState.hpp>
 #include <psio/to_key.hpp>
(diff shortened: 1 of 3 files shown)


libraries/psibase/common/include/psibase/Table.hpp
```diff
@@ -271,12 +271,30 @@ namespace psibase
...
+   ///
+   /// For example, a primr
+   /// Warning: No type checking! It's up to the user to ensure that T + Rest
+   ///   is equivalent to the primary key)
+   template <typename Rest, typename T>
+   struct SplitKey
+   {
+      T value;
+   };
+
```
swatanabe, line 285: The reason I suggested `SplitKey` is that it's non-intrusive. If we're adding this to the main `Table.hpp` header, a better solution would be to add an overload of `subindex` that takes an explicit type.
`subindex<Rest>(key)`

James-Mart: I decided to make this non-intrusive rather than try to figure out how to implement your subindex idea

swatanabe (approved):

PR #723, opened 2024-05-24
Add `userTokens` query

James-Mart:
* Adds the `userTokens` query to the tokens query service. This can be used to look up any tokens for which the current user owns the issuer NFT.
* Also, this PR updates the `.ownerDetails` field in the `tokenDetails` query to be an optional value. This field will be empty if, for example, the issuer NFT has been burned.

### DIFF
--- a/services/user/Tokens/src/RTokens.cpp
+++ b/services/user/Tokens/src/RTokens.cpp
@@ -14,6 +14,8 @@
 using namespace UserService;
 using namespace std;
 using namespace psibase;
+using SystemService::Accounts;
+using SystemService::AccountTable;
 
 auto tokenService = Tokens::Tables{Tokens::service};
 namespace TokenQueryTypes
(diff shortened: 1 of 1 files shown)


services/user/Tokens/src/RTokens.cpp
```diff
@@ -164,6 +166,34 @@ struct TokenQuery
...
+                   optional<uint32_t>    last,
+                   optional<std::string> before,
+                   optional<std::string> after) const
+   {
+      std::vector<TokenRecord> tokens;
+
+      auto tokenTypeIdx = tokenService.open<TokenTable>().getIndex<0>();
+      for (auto token : tokenTypeIdx)
+      {
+         token.ownerNft;
```
Velua, line 180:
```suggestion

```

Velua (approved):

PR #726, opened 2024-05-28
Rust Fracpack: Omit trailing optionals

sparkplug0025:
Fracpack fix, plus adjust the `burst-transfer` rust tool just to fix our `cargo build`.

### DIFF
--- a/rust/fracpack/src/fracpack.rs
+++ b/rust/fracpack/src/fracpack.rs
@@ -136,6 +136,11 @@ pub trait Pack {
     }
 
     #[doc(hidden)]
+    fn is_empty_optional(&self) -> bool {
+        false
+    }
+
+    #[doc(hidden)]
     fn embedded_fixed_pack(&self, dest: &mut Vec<u8>) {
         if Self::VARIABLE_SIZE {
             dest.extend_from_slice(&0_u32.to_le_bytes());
@@ -245,6 +250,17 @@ pub trait Unpack<'a>: Sized {
             return Err(Error::BadOffset);
         }
         Self::unpack(src, heap_pos)
+    }
+
+    #[doc(hidden)]
+    fn check_opt_embedded_unpack(
+        src: &'a [u8],
+        fixed_pos: &mut u32,
+        heap_pos: &mut u32,
+        _not_trailing: bool,
+        _has_opt_bytes: bool,
+    ) -> Result<Self> {
+        Self::embedded_unpack(src, fixed_pos, heap_pos)
     }
 
     #[doc(hidden)]
@@ -487,6 +503,10 @@ impl<T: Pack> Pack for Option<T> {
         Self::embedded_variable_pack(self, dest);
     }
 
+    fn is_empty_optional(&self) -> bool {
+        self.is_none()
+    }
+
     fn embedded_fixed_pack(&self, dest: &mut Vec<u8>) {
         if T::IS_OPTIONAL || !T::VARIABLE_SIZE {
             dest.extend_from_slice(&1u32.to_le_bytes())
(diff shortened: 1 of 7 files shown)


rust/test_fracpack/tests/test.rs
```diff
@@ -336,7 +337,6 @@ fn t1() -> Result<()> {
         let unpacked = OuterStruct::unpack(&packed[..], &mut 0)?;
         assert_eq!(*t, unpacked);
         test_fracpack::bridge::ffi::round_trip_outer_struct(i, &packed[..]);
-        // TODO: optionals after fixed-data portion ends
```
sparkplug0025, line 339: I believe this is not necessary, since all the structs that we are comparing with C++ examples has trailing optionals and we make sure they are equal. LMK if you think otherwise.

rust/psibase_macros/src/fracpack_macro.rs
```diff
@@ -726,3 +843,36 @@ fn process_enum(
         #unpack_impl
     })
 } // process_enum
+
+// Helper function to determine if a type is Option<T>
+fn is_option_type(ty: &syn::Type) -> bool {
```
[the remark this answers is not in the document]
sparkplug0025: Nice catch! Will add an unit test for it btw.

rust/psibase_macros/src/fracpack_macro.rs
```diff
@@ -294,13 +294,22 @@ fn process_struct(
...
+        quote! {<#ty as #fracpack_mod::Pack>::IS_OPTIONAL}
+    });
+
+    let check_optional_fields = fields.iter().map(|field| {
+        let name = &field.name;
+        let ty = &field.ty;
+        if opts.definition_will_not_change {
+            quote! {true}
+        } else {
+            quote! {!<#ty as #fracpack_mod::Pack>::IS_OPTIONAL || !self.#name.is_empty_optional()}
```
swatanabe, line 309: Isn't checking `IS_OPTIONAL` redundant?

rust/psibase_macros/src/fracpack_macro.rs
```diff
@@ -312,6 +321,24 @@ fn process_struct(
...
+
+    let fixed_size = fields
+        .iter()
+        .map(|field| {
+            let ty = &field.ty;
+            quote! {<#ty as #fracpack_mod::Pack>::FIXED_SIZE}
+        })
+        .fold(quote! {0}, |acc, new| quote! {#acc + #new});
+
+    let heap_size = fields
```
swatanabe, line 333: The name `heap_size` is somewhat confusing. We've been using the term "heap" to refer to the region after the fixed size fields.

rust/psibase_macros/src/fracpack_macro.rs
```diff
@@ -726,3 +831,39 @@ fn process_enum(
...
+                if let syn::PathArguments::AngleBracketed(_) = &segment.arguments {
+                    return true;
+                }
+            }
+        }
+    }
+    false
+}
+
+fn find_last_possible_trailing_optional_field_idx(
```
swatanabe, line 850: This looks like dead code

rust/fracpack/src/fracpack.rs
swatanabe, line 1: tuples need the same handling of optionals as structs.

swatanabe:
The validation looks like it's weaker than in C++. I suspect that a number of [these tests](https://github.com/gofractally/psibase/blob/main/libraries/psio/tests/test_fracpack.cpp#L253) would fail when applied to the rust implementation. I don't know this is critical to handle right now, though.

rust/psibase_macros/src/fracpack_macro.rs
sparkplug0025: I agree. This name was introduced by Todd and I just kept using it... not sure if in the v0 implementation it was in fact just the heap size and as it evolves it does not make sense anymore...

What name would you like to suggest here? Notice that it is only used for the packing fn and it does this assertion right below:

```
let heap = #heap_size;
assert!(heap as u16 as u32 == heap); // TODO: return error
```

rust/fracpack/src/fracpack.rs
sparkplug0025: Capturing as a follow up issue...

sparkplug0025:
@swatanabe 
> The validation looks like it's weaker than in C++. I suspect that a number of [these tests](https://github.com/gofractally/psibase/blob/main/libraries/psio/tests/test_fracpack.cpp#L253) would fail when applied to the rust implementation. I don't know this is critical to handle right now, though.

 I want to move forward with this PR just to wrap up the rust tester work. I created issues to track the remaining improvements:
- #731 
- #732

Does it sound good? If so, you can approve and merge!

rust/psibase_macros/src/fracpack_macro.rs
swatanabe: `heap` makes sense. `heap_size` doesn't. It's the size of the fixed data region, which is equal to the offset of the start of the heap.

swatanabe (approved):

PR #734, opened 2024-06-05
Add range operations to triedent

swatanabe:


libraries/triedent/test/test_range_compare.cpp
```diff
@@ -0,0 +1,376 @@
...
+{
+   std::vector<std::string_view> rows;
+   std::string_view              lower;
+   std::string_view              upper;
+};
+
+TEST_CASE("test splice")
+{
+   using namespace std::literals::string_view_literals;
+   static std::vector<take_data> data{
```
James-Mart, line 303: I think you meant `splice_data` here

James-Mart (approved):

PR #737, opened 2024-06-10
Add shadcn/ui to X-Admin dashboard

Velua:
- [x] Installs `shadcn-ui` adopting several components and deleting previously used ones. 
- [x] Uses a progress bar for installation and use of a table rather than a list for packages to be installed. 
- [x] Adds a dark mode / light mode 
- [x] Moves the shutdown / reboot page to a menu at the top right with the dark / light mode switcher.


swatanabe:
I clicked around on the new XAdmin service a bit:
- I can't see any visual indication of the links in the banner: "No chain running. Create a new chain or connect to an existing chain"
- On the boot page, at the final confirmation before install, the text "A list of packages to be installed." underneath the table seems redundant.
- In the configuration page, the logger fields that are not relevant to the current logger type used to be hidden. They are now visible but not editable.
- The "save changes" button on the configuration page used to be disabled when nothing had been modified. That is no longer the case.

James-Mart:
### Found a UI bug

Navigate through the following button sequence:
* Boot > Choose services to install > Next > Back > Install all services > Next

Notice that after clicking "back" the "Install all services" radio button is already selected. But clicking "next" takes you back to the "Choose services to install" screen instead of switching to the "Install all services" flow.

James-Mart:
Instead of 
image

I prefer what you did on the main home screen:
image

I think the two buttons should be the `moon` button to adjust the theme, and a `power` button for shutdown/restart

Velua:
> I clicked around on the new XAdmin service a bit:
> 
>     * I can't see any visual indication of the links in the banner: "No chain running. Create a new chain or connect to an existing chain"
> 
>     * On the boot page, at the final confirmation before install, the text "A list of packages to be installed." underneath the table seems redundant.
> 
>     * In the configuration page, the logger fields that are not relevant to the current logger type used to be hidden. They are now visible but not editable.
> 
>     * The "save changes" button on the configuration page used to be disabled when nothing had been modified. That is no longer the case.

@swatanabe 
1. Fixed
2. Removed
3. Fixed
4. This seems to be occurring on `main` as well, seems like the polling and subsequent effect makes the form `dirty` state flick between `true` and `false`. I'd like to address this in another PR, as we're considering a refactor large enough this gets fixed as a side effect. 

Ekrankopio de 2024-06-12 11-15-14

Velua:
> Instead of image
> 
> I prefer what you did on the main home screen: image
> 
> I think the two buttons should be the `moon` button to adjust the theme, and a `power` button for shutdown/restart

Implemented

Velua:
> ### Found a UI bug
> 
> Navigate through the following button sequence:
> 
>     * Boot > Choose services to install > Next > Back > Install all services > Next
> 
> 
> Notice that after clicking "back" the "Install all services" radio button is already selected. But clicking "next" takes you back to the "Choose services to install" screen instead of switching to the "Install all services" flow.

Found the issue and fixed.

James-Mart:
@brandonfancher I believe my own comments and steven's were addressed, just in case you were waiting on that before reviewing

brandonfancher (approved):

PR #741, opened 2024-06-14
Attestation UI

cool-ant:
This is the "initial" implementation of the Attestation UI for the (not yet extant) Attestation service


services/user/Attestation/ui/package.json
```diff
@@ -0,0 +1,34 @@
...
+    "preview": "vite preview"
+  },
+  "dependencies": {
+    "@rollup/plugin-alias": "^5.1.0",
+    "d3": "^7.9.0",
+    "react": "^18.2.0",
+    "react-dom": "^18.2.0",
+    "vite-plugin-top-level-await": "^1.4.1",
+    "vite-plugin-wasm": "^3.3.0",
+    "vite-tsconfig-paths": "^4.3.2"
```
brandonfancher, line 19: These can be `devDependencies`, if I'm not mistaken.

services/user/Attestation/ui/src/Sidebar.tsx
```diff
@@ -0,0 +1,138 @@
...
+              type="color"
+              id="attestation-backlink-color"
+              className="ui-attestation-backlinks"
+              value={uiOptions.colors[SELECTOR.BACK_LINK]}
+              onChange={handlers.handleLinkColorChange(SELECTOR.BACK_LINK)}
+            />
+          </div>
+        </>
+      ) : (
+        <div></div>
```
brandonfancher, line 68: Usually, we just put `null` here in the empty cases (vs. an empty div).

brandonfancher (approved):
Looks good to me! Builds and runs in dev and preview well. I'm adding just a couple comments to address. And there is a good amount of commented code... but sounds like you're still working on cleanup. Approved.

PR #750, opened 2024-06-17
Update changelog title

James-Mart:
Front-end Infra -> Plugin Infra


swatanabe:
Does this mean that we retroactively changed the labels on pre-plugin issues and PRs?

James-Mart:
> Does this mean that we retroactively changed the labels on pre-plugin issues and PRs?

Yes, the theory is that anything that fell into the "front end infra" was effectively plugin-related infra, we just didn't have the language yet.

swatanabe (approved):

Issue #742, opened 2024-06-17
Admin app rework


James-Mart:
# Description

The admin app needs to be revamped for increased usability and functionality.
Primary use-cases for the app:
1. Creating a new psibase network
2. Joining an existing psibase network
3. Configuring a node that has already joined a network
4. Monitoring a node that has already joined a network

This task is primarily focused on the first 3 use-cases, as monitoring is lower priority than network creation/joining at this stage. 

# Tasks

```[tasklist]
- [x] By default, show the "Join a Network" wizard if psinode has no existing connection to any network
- [x] Allow access to the "Create a Network" wizard if the user does not intend to join any existing network
- [x] Rework single configuration page from into a horizontally navigable experience
- [x] Rework logs page for easier filtering
- [x] Add a new home page to the Admin app for when the app is run while connected to a network. This home page shows some basic statistics about the network, or potentially about the node.
- [ ] Add missing configurations, mainly for key management and TLS certificate management
```
-----------------------

@Velua feel free to add any notes or items that I am not remembering from our conversation

Velua:
## Paths
- Join network or create network
- - Upgrade to block producing node
Create network should immediately resolve to block producing form, join network won't. 
**Connection to the network is done with a single node, multiple URLs not needed**

## Peers page
- Should be allowed to terminate connections
- Switch between transient and persistent connections. 
- Wishlist: Show ping for connections
- Wishlist: See connection account name

## Logs page
- Should be able to toggle between remembering the last 25 logs and all logs. 


## New dashboard page
Some kinda welcome page for something cute, Graphana may or may not be running so cannot assume it's going to be ready like we currently do.

Velua:
- [ ] Rename BlockProducer name to NodeAccountName on the Join network flow. 

# Create Page
Radio boxes as part of the Create network wizard, 
Ask between dev and prod (difference is the TokenUsers package)
Ask between None governance and fractal democracy (disabled for now as we don't have it)
These questions compose the DataTables tick boxes determinine what gets installed.
The table is hidden by default and openable via a "Custom settings" accordion which displays a data table.
Unticking A will cause a dialog to ask confirmation that B and C is also lost as a consequence. 

# Join Page
- [ ] Connect to URL - Send a API request to check it's online and responding before moving to next page. 
- [ ] NodeAccountName

James-Mart:
Misc feedback gathered while testing the latest update

### Boot wizard

[...]

PR #761, opened 2024-06-19
X-Admin app improvements

Velua:
- [x] Remove in house hash routing, adopt [React Router Hash Routing](https://reactrouter.com/en/main/router-components/hash-router).
- [x] Remove custom hooks for queries and adopt React Query. 
- [x] Zod type API responses / Centralised all chain API requests. 
- [x] Add new dashboard page alongside graphana
- [x] Remove old boot page / form
- [x] Add setup / Join wizard


James-Mart:
Locally, I merged with the latest main, built and tested this. But I'm getting this ZodError on validating the `http_timeout` field of the config:
image

If I change the zod schema to expect a string then it works and lets me boot.

James-Mart:
This looks very awesome

James-Mart (changes requested):
I left various feedback on [this issue](https://github.com/gofractally/psibase/issues/742) that I gathered while using the interface. Nothing I'd block this PR over.

Only thing that seems necessary to fix before merge (from a usability perspective) is the [zod parse error](https://github.com/gofractally/psibase/pull/761#issuecomment-2269282618) that prevents the chain from booting.

James-Mart:
Arrow button covered by expanded Advanced tab
image

James-Mart:
My config seems to get corrupted at some point... a "service" key is getting deleted somehow
image

James-Mart:
Also if I start out with `http_timeout = 4s`, I notice later in my config it is saying 
`http-timeout = 4000 s`

Is it converting to `ms` and using the wrong label?

James-Mart:
If I add a peer, and then I click "disconnect", I think there's two things that are strange:
1. After a while, the connection is automatically re-established. This is unsurprising because presumably the peer is still in the configured list of peers, but the UX feels strange.
2. If I disconnect, and then click the "..." button, the option still says "Disconnect" instead of "Connect"
image

swatanabe:
> Also if I start out with `http_timeout = 4s`, I notice later in my config it is saying `http-timeout = 4000 s`
> 
> Is it converting to `ms` and using the wrong label?

If that's happening, it's probably a bug on the native side. The `/native/admin/config` endpoint takes an integer in microseconds, with no units.

Velua:
> If I add a peer, and then I click "disconnect", I think there's two things that are strange:
> 
>     1. After a while, the connection is automatically re-established. This is unsurprising because presumably the peer is still in the configured list of peers, but the UX feels strange. I think we can probably improve this UX later.
> 
>     2. If I disconnect, and then click the "..." button, the option still says "Disconnect" instead of "Connect"
>        image

Yeah I think the solution here is just to prevent crossing lanes, 
- Transient connections get disconnect
- Persistent connections get remove

It's awkward if persistent connections get both only to reconnect again anyway. 

- [ X ] Render "disconnect" button only on active transient connections. 
- [ X ] Render "remove" only on online & offline persistent connections.

swatanabe:
> My config seems to get corrupted at some point... a "service" key is getting deleted somehow

It's a native bug that seems to be caused by having multiple entries for `localhost`

James-Mart:
> Transient connections get disconnect

Although why even bother disconnecting if it's a transient connection, since it'll just come back again, right? Unless we can blacklist somehow @swatanabe

CMakeLists.txt
```diff
@@ -703,15 +703,26 @@ psibase_package(
 )
 
 psibase_package(
-    OUTPUT ${SERVICE_DIR}/Default.psi
-    NAME Default
+    OUTPUT ${SERVICE_DIR}/DevDefault.psi
+    NAME DevDefault
```
swatanabe, line 707: `psibase boot` needs to be updated

James-Mart: Tester as well, yeah?
https://github.com/gofractally/psibase/blob/98f3198a66f4754b29468ce783bd7a3496fd6219/rust/psibase/src/main.rs#L646
and
https://github.com/gofractally/psibase/blob/98f3198a66f4754b29468ce783bd7a3496fd6219/rust/psibase/src/tester.rs#L62

James-Mart (approved):

PR #768, opened 2024-06-25
Prepare for cargo component update

James-Mart:


James-Mart:
Apparently, just using `cargo component build --release` wasn't enough for the build to actually use a sensible release configuration. So this PR adds some configuration for release builds that turns on optimization and strips out debug info, etc.. 
Effectively this gets the size of the generated components to be *much* closer to what I expected to begin with.

### Size comparison

Before this PR:
* accounts plugin: 1.78 mb
* auth_sig plugin: 1.91 mb
* invite plugin: 1.9 mb

After this PR:
* accounts plugin: 136 kb
* auth_sig plugin: 206 kb
* invite plugin: 227 kb

swatanabe (approved):

swatanabe (approved):

PR #775, opened 2024-07-05
Deploy token plugin by default

James-Mart:
I noticed the token app has a plugin, but it's not automatically deployed. This deploys it to the chain by default.


James-Mart:
Requesting review from @swatanabe because haven't heard from @Velua

swatanabe (approved):

PR #779, opened 2024-07-10
Refactor supervisor

James-Mart:
Plugins no longer loaded in individual iframes. They load directly in supervisor instead.
The entire API has become synchronous, no need for restarting the plugin execution from scratch (groundhog day).
These changes, among various other cleanups, contributed to more than an order of magnitude in speedup for plugin execution.

For reviewers, I recommend starting with the changes to the documentation, particularly the `supervisor.md` document, which includes details about exactly how the new supervisor implementation works.


services/user/CommonApi/common/packages/common-lib/src/messaging/FunctionCallRequest.ts
```diff
@@ -12,6 +12,15 @@ export interface QualifiedFunctionCallArgs extends FunctionCallArgs {
...
 
+export const isQualifiedFunctionCallArgs = (obj: any): obj is QualifiedFunctionCallArgs => {
+    return typeof obj === 'object' && obj !== null
+        && typeof obj.service === 'string'
+        && typeof obj.plugin === 'string'
+        && (typeof obj.intf === 'string' || obj.intf === undefined)
+        && typeof obj.method === 'string'
+        && Array.isArray(obj.params);
+}
+
```
Velua, line 23: One day we should look into zod in this section, so we could instead do something like this...

```
const QualifiedFunctionCallArgsSchema = z.object({
    service: z.string(),
    plugin: z.string().optional(),
    intf: z.string(),
    method: z.string(),
    params: z.array()
})

export type QualifiedFunctionCallArgs = z.infer<typeof QualifiedFunctionCallArgsSchema>

export const isQualifiedFunctionCallArgs = (obj: any): obj is QualifiedFunctionCallArgs => {
    return QualifiedFunctionCallArgsSchema.safeParse(obj).success
}
```

services/user/CommonApi/common/packages/common-lib/src/supervisor.ts
```diff
@@ -121,23 +122,31 @@ export class Supervisor {
         const expected = this.pendingRequest.call;
         const received = response.call;
         const unexpected =
-            expected.method != received.method ||
-            expected.service != received.service;
+            expected.method !== received.method ||
+            expected.service !== received.service;
+
+        const {result} = response;
```
Velua, line 128: Is your prettier running? Because this suggests it's not and the next dev who visits is gonna end up making formatting changes. 
```suggestion
        const { result } = response;
```

services/user/Supervisor/ui/src/callStack.ts
```diff
@@ -0,0 +1,61 @@
...
+            args: args,
+            startTime: Date.now(),
+        });
+    }
+
+    pop(): Call | undefined {
+        if (this.storage.length === 0) {
+            console.error(`Tried to pop empty callstack`);
+            return undefined;
+        }
```
Velua, line 32: You could use your `isEmpty` method for this.

services/user/Supervisor/ui/src/callStack.ts
```diff
@@ -0,0 +1,61 @@
...
+    caller: OriginationData;
+    args: QualifiedFunctionCallArgs;
+    startTime?: number;
+}
+
+// Callstack implementation that manages frames for every inter-plugin call
+export class CallStack {
+    private storage: Array<Call> = [];
+
+    push(sender: OriginationData, args: QualifiedFunctionCallArgs): void {
```
Velua, line 16: Is it worth distinguishing these methods between `private` and `public`?
```suggestion
    public push(sender: OriginationData, args: QualifiedFunctionCallArgs): void {
```

services/user/Supervisor/ui/src/callStack.ts
```diff
@@ -0,0 +1,61 @@
...
+    }
+
+    pop(): Call | undefined {
+        if (this.storage.length === 0) {
+            console.error(`Tried to pop empty callstack`);
+            return undefined;
+        }
+
+        let popped = this.peek(0)!;
+        let resolutionTime = Date.now() - popped.startTime!;
```
Velua, line 35: We use `const` instead of `let` by default, only reaching for `let` when we want to mutate variables.

Velua (approved):

services/user/CommonApi/common/packages/common-lib/src/supervisor.ts
James-Mart: Fixed. I ran prettier in the `Supervisor` project but forgot `common-lib`.
Also, I've spent some time this morning figuring out how to set up [multi-root workspaces](https://code.visualstudio.com/docs/editor/multi-root-workspaces) in vscode... it's pretty cool. And now prettier should always be running on save by default.

services/user/Supervisor/ui/src/callStack.ts
James-Mart: done

services/user/Supervisor/ui/src/callStack.ts
James-Mart: I was just marking when things were `private` since they are public by default. Although I see a few places where I applied this strategy inconsistently, so I've gone back and made sure all my classes use the strategy consistently.

services/user/Supervisor/ui/src/callStack.ts
James-Mart: Fixed

services/user/Supervisor/ui/src/component-loading/loader.ts
```diff
@@ -0,0 +1,224 @@
...
+export async function loadPlugin(
+    wasmBytes: Uint8Array,
+    pluginHost: HostInterface,
+    api: ComponentAPI,
+) {
+    const imports = mergeImports([
+        await getWasiImports(),
+        await getHostImports(),
+        await getProxiedImports(api.importedFuncs),
+    ]);
```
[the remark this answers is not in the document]
James-Mart: No races, good catch! Fixed

services/user/Supervisor/ui/src/component-loading/loader.ts
```diff
@@ -0,0 +1,224 @@
...
+    const { app } = pluginHost.getSelf();
+    assert(app !== undefined, "Plugin must correspond to a psibase service");
+
+    const pluginModule = await load(wasmBytes, imports, `${app!} plugin`);
```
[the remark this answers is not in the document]
James-Mart: Cool! Fixed.

services/user/Supervisor/ui/src/plugin/plugin.ts
```diff
@@ -0,0 +1,106 @@
...
+            throw new PluginInvalid(this.id);
+        }
+
+        const { exportedFuncs } = this.componentAPI;
+
+        if (intf !== undefined) {
+            const foundInterface = exportedFuncs.interfaces.find(
+                (i) => i.name === intf,
+            );
+            return !!foundInterface?.funcs.find((f) => f === method);
```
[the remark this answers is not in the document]
James-Mart: Fixed

services/user/Supervisor/ui/src/plugin/plugin.ts
```diff
@@ -0,0 +1,106 @@
...
+        }
+
+        return !!exportedFuncs.funcs.find((f) => f === method);
```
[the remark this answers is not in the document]
James-Mart: Fixed

services/user/Supervisor/ui/src/serviceContext.ts
```diff
@@ -0,0 +1,43 @@
...
+            return { plugin: p, new: false };
+        }
+
+        p = new Plugin({ service: this.service, plugin }, this.hostInterface);
+        this.plugins.push(p);
+        return { plugin: p, new: true };
+    }
+
+    hasLoaded(plugin: Plugin): boolean {
+        return !!this.plugins.find(
```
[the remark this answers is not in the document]
James-Mart: Fixed

services/user/Supervisor/ui/src/supervisor.ts
```diff
@@ -0,0 +1,270 @@
...
+                {
+                    service: args.service,
+                    plugin: args.plugin,
+                },
+            ]);
+
+            // Make a *synchronous* call into the plugin. It can be fully synchronous since everything was
+            //   preloaded.
+            // TODO: Consider if a plugin runs an infinite loop. We need a way to terminate the
+            //   current call and report the faulty plugin.
```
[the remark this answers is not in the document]
James-Mart: Good point, and I don't think that's worth it.  If a plugin runs an infinite loop, I suppose users will just complain and not use that app anymore.

services/user/Supervisor/ui/src/supervisor.ts
```diff
@@ -0,0 +1,270 @@
...
+    }
+
+    // This is an entrypoint for apps to call into plugins.
+    async entry(
+        callerOrigin: string,
+        args: QualifiedFunctionCallArgs,
+    ): Promise<any> {
+        try {
+            // Handle if a call is already in progress
+            // TODO - Buffer multiple calls from root app
```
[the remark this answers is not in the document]
James-Mart: I suppose it's not necessary. Removed

PR #783, opened 2024-07-17
add account exists query

James-Mart:
Allows the accounts plugin to more easily implement a check for whether a particular account exists, which then enables third-party plugins to check for the existence of accounts.

### DIFF
--- a/services/system/Accounts/src/RAccounts.cpp
+++ b/services/system/Accounts/src/RAccounts.cpp
@@ -19,12 +19,13 @@ namespace SystemService
 {
    struct AccountsQuery
    {
-      auto exists(AccountNumber account) const
-      {  //
-         return to<Accounts>().exists(account);
+      auto getAccount(AccountNumber account) const
+      {
+         Accounts::Tables tables{Accounts::service};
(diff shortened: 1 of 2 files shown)


services/system/Accounts/src/RAccounts.cpp
```diff
@@ -16,50 +17,23 @@ using Tables = psibase::ServiceTables<psibase::WebContentTable>;
...
-            auto accountsTables = Accounts::Tables(Accounts::service);
-            auto accountTable   = accountsTables.open<AccountTable>();
-            auto accountIndex   = accountTable.getIndex<0>();
-
-            std::vector<Account> rows;
-            for (auto it = accountIndex.begin(); it != accountIndex.end(); ++it)
-               rows.push_back(*it);
-            return to_json(rows);
-         }
+      auto exists(AccountNumber account) const
```
swatanabe, line 22: Instead of a boolean, I'd prefer a query that returns the Account or null if it doesn't exist.

James-Mart: Done

swatanabe (approved):

PR #784, opened 2024-07-18
Prevent uploading overshadowed artifacts

James-Mart:
If a service has a registered http server, then it should upload artifacts to its own server, rather than to the default `sites` server. Any artifacts uploaded to sites will never be served since requests will already be proxied to the registered server.

Therefore, this PR updates the sites HTTP server to assert if it detects that the user is trying to upload such overshadowed artifacts.

CC @cool-ant


swatanabe:
I though we were planning to do the opposite and always use `sites` for static content, so we can have a single unified file manager.

PR #785, opened 2024-07-18
Add management of unapplied transactions to transact service

swatanabe:
API Changes
---

For app developers the changes in this PR are minimal. The new endpoint `http://transact.<ROOT HOST>/push_transaction` has nearly the same interface as the existing `/native/push_transaction`. In particular, it returns a trace as soon as a transaction either fails or is included in a block. Transactions are forwarded between block producers. Non-voting nodes do not forward transactions and submitting a transaction to such a node will likely cause the connection to hang. `/native/push_transaction` remains for now but should be considered deprecated.

Internals
---

- `psinode` now calls `transact::nextTransaction` to find a transaction to apply while building a block. 
- New callbacks `acceptTransaction` and `rejectTransaction`, receive the transaction id and transaction trace as arguments.
- New intrinsics `socketSend` and `socketAutoClose` to interact with sockets.
- `http-server::serve` now takes an `i32` socket handle and should call `socketSend` with the response instead of returning it.
- Predefined socket that a producer can use to broadcast to all other producers.
- New entry point `http-server::recv` handles incoming socket messages.


James-Mart:
This doc should be updated:
https://github.com/gofractally/psibase/blob/70d4d8f7524fc44022cb08091be1c55f5cc40e9d/doc/psidk/src/specifications/data-formats/package.md?plain=1#L42

And in general, docs/diagrams would be helpful here to more quickly understand the relationships between native/httpserver/transact and how sockets are managed.

James-Mart (approved):

PR #788, opened 2024-07-26
Remove Reflect

swatanabe:
- Add actions to schema so that the schema can fill all the roles previously handled by `Reflect`


rust/psibase/src/services/invite.rs
```diff
@@ -18,11 +17,8 @@ pub struct InviteRecord {
     state: u8,
 }
 
-#[derive(
-    Debug, Copy, Clone, Pack, Unpack, Reflect, Serialize, Deserialize, SimpleObject, InputObject,
-)]
+#[derive(Debug, Copy, Clone, Pack, Unpack, Serialize, Deserialize, SimpleObject, InputObject)]
```
James-Mart, line 20: Was `ToSchema` intended here?

swatanabe: It wouldn't hurt to add `ToSchema`, but it isn't actually needed yet, since `NewAccountRecord` isn't used in any action or event signatures.

James-Mart (approved):

PR #790, opened 2024-07-29
Add rust/psibase unit tests to ctest

swatanabe:
- Combine rust unit tests into a single run of cargo
- Mark rs-tests as using all available cores, so ctest doesn't try to run other tests in parallel with it
- Clear assumptions so that optimizing non-requested packages doesn't just fail.
- Make depgraph tests compile again

### DIFF
--- a/rust/CMakeLists.txt
+++ b/rust/CMakeLists.txt
@@ -1,12 +1,6 @@
 enable_testing()
 
 add_test(
-    NAME CargoPsibase-test
-    COMMAND cargo test --manifest-path ${CMAKE_CURRENT_SOURCE_DIR}/cargo-psibase/Cargo.toml
-    WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/cargo-psibase/
-)
-
-add_test(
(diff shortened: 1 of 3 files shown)


rust/CMakeLists.txt
```diff
@@ -18,5 +12,11 @@ add_test(
 
 add_test(
     NAME rs-tests
-    COMMAND cargo test --manifest-path ${CMAKE_CURRENT_SOURCE_DIR}/Cargo.toml --target-dir ${CMAKE_CURRENT_BINARY_DIR} -p test_fracpack
+    COMMAND cargo test --manifest-path ${CMAKE_CURRENT_SOURCE_DIR}/Cargo.toml --target-dir ${CMAKE_CURRENT_BINARY_DIR} -p test_fracpack -p psibase -p cargo-psibase
```
sparkplug0025, line 15: WDYT about adding `test_contract` and `test_contract_2` here too (although it needs to be executed from `cargo psibase test`)? I think it would be nice since these two services are our main examples that we always keep up-to-date (or at least try to).

PR #793, opened 2024-07-30
Add webmail app skeleton

James-Mart:
Adds a skeleton for a webmail app
cc @brandonfancher 

Supported REST query:
    `GET /messages?sender=alice&receiver=bob`
Either sender or receiver (or both) must be specified.


CMakeLists.txt
```diff
@@ -615,14 +616,29 @@ psibase_package(
     DEPENDS ${InvitesPlugin_DEP}
 )
 
+psibase_package(
+    OUTPUT ${SERVICE_DIR}/Email.psi
+    NAME Email
```
swatanabe, line 621: I don't think we should call this "email" unless it's actually able to interface to SMTP somehow. Otherwise, anyone seeing the name is just going to be confused.

services/user/Email/src/Email.cpp
```diff
@@ -0,0 +1,131 @@
...
+                 const std::string&     body)
+{
+   check(to<SystemService::Accounts>().exists(recipient),
+         "target recipient account does not exist");
+   emit().history().sent(getSender(), recipient, subject, body);
+}
+
+// Supported REST queries:
+// GET /messages/sender?user=alice
+// GET /messages/recipient?user=alice
```
swatanabe, line 85: I would make both sender and receiver query parameters (and maybe allow both together)

services/user/Email/src/Email.cpp
```diff
@@ -0,0 +1,131 @@
...
+      }
+      if (property)
+      {
+         auto user = extractUser(request.target);
+         if (user == nullopt)
+            return nullopt;
+
+         return to<REvents>().serveSys(
+             makeQuery(request, "SELECT * FROM \"history.email.sent\" WHERE " + property.value() +
+                                    " = '" + user.value() + "' ORDER BY ROWID"));
```
swatanabe, line 107: SQL injection

services/user/Email/src/Email.cpp
```diff
@@ -0,0 +1,131 @@
...
+
+// Supported REST queries:
+// GET /messages/sender?user=alice
+// GET /messages/recipient?user=alice
+optional<HttpReply> serveRestApi(const HttpRequest& request)
+{
+   if (request.method == "GET")
+   {
+      optional<string> property;
+      if (request.target.find("/messages/sender") != string::npos)
```
swatanabe, line 91: Do you really mean `find` and not `starts_with`?

services/user/Email/src/Email.cpp
```diff
@@ -0,0 +1,131 @@
...
+              .rootHost    = in.rootHost,
+              .method      = "POST",
+              .target      = "/sql",
+              .contentType = "application/sql",
+              .body        = std::vector(s.begin(), s.end())};
+   }
+
+   auto extractUser = [](const std::string& target) -> optional<string>
+   {
+      auto pos = target.find("?user=");
```
swatanabe, line 32: This can mis-parse the URL in lots of ways. "The characters slash ("/") and question mark ("?") may represent data within the query component" ([RFC 3986](https://www.rfc-editor.org/rfc/rfc3986#page-24))

CMakeLists.txt
James-Mart: How about "Webmail"?

James-Mart: Renamed it webmail. If you want yet another name we can align later and I'll make a separate pr

services/user/Email/src/Email.cpp
James-Mart: Done

services/user/Email/src/Email.cpp
James-Mart: But it's a read-only table.
Nonetheless, the new version confirms that the values passed for sender and receiver are valid and existing AccountNumbers.

services/user/Email/src/Email.cpp
James-Mart: Forgot about starts_with, fixed

services/user/Email/src/Email.cpp
James-Mart: The new logic should address this

services/user/Webmail/src/Webmail.cpp
```diff
@@ -0,0 +1,178 @@
...
+      return to<SystemService::Accounts>().exists(acc);
+   };
+
+   map<string, string> parseQuery(const string& query)
+   {
+      map<string, string> params;
+      istringstream       queryStream(query);
+      string              pair;
+
+      while (getline(queryStream, pair, '&'))
```
swatanabe, line 47: Doesn't this require a trailing '&'?

swatanabe: nvm. iostreams seems like a bit of overkill though.

services/user/Webmail/src/Webmail.cpp
```diff
@@ -0,0 +1,178 @@
...
+}
+
+optional<HttpReply> Webmail::serveSys(HttpRequest request)
+{
+   if (auto result = serveContent(request, Tables{}))
+      return result;
+   if (auto result = serveSimpleUI<Webmail, true>(request))
+      return result;
+   if (auto result = serveRestApi(request))
+      return result;
```
cool-ant, line 167: I don't know when/where these might overlap, but wouldn't you want the serveSimpleUI to come last, given it's basically a fallback?

swatanabe (approved):

services/user/Webmail/src/Webmail.cpp
James-Mart: The UI part is a fallback, so it should at least come after `serveContent`

PR #798, opened 2024-08-02
Chainmail app UI

brandonfancher:
## Special attention for:
- Did I update `CMakeLists.txt` properly?

## In this PR
- Chain now boots with webmail UI
- Home page lists "Chain mail" as an app
- Slight improvements to home layout/responsiveness
- Drafts supported in `localstorage`

## Limitations
- 3 hard-coded accounts: `alice`, `bob`, and `webmail`
- `alice` alone can send; all three accounts can receive
- Emails cannot be deleted. Support for that in plugin/service forthcoming

## Future work
- Delete emails
- Mark read/unread?
- Support for threading replies?
- Dark mode
- Empty UI states
- Send from accounts other than `alice`
- More control bar options for markdown editor
- Splash page/dialog explaining to the user it's just for demonstration purposes (until it's not)
- Additional (potential) fields on message: `id`, `datetime`, `status`, `read`, `inReplyTo: id`, etc.
- `zod.parse` incoming messages with `strict` on (https://github.com/gofractally/psibase/pull/798#discussion_r1702299492)
- Audit vite config; lots of unused stuff in there


services/user/Webmail/ui/src/hooks/use-mail.ts
```diff
@@ -0,0 +1,131 @@
...
+                inReplyTo: null,
+                subject: msg.subject,
+                body: msg.body,
+            }) as Message,
+    );
+};
+
+const getIncomingMessages = async (account: string) => {
+    const res = await fetch(`/messages?receiver=${account}`);
+    const rawMessages = (await res.json()) as RawMessage[];
```
Velua, line 50: Given we already have it installed it would be kinda nice to `zod.parse` this with `strict` on as well.

CMakeLists.txt
```diff
@@ -626,7 +627,9 @@ psibase_package(
         WASM ${CMAKE_CURRENT_BINARY_DIR}/Webmail.wasm
         SERVER webmail
         DATA ${COMPONENT_BIN_DIR}/${WebmailPlugin_OUTPUT_FILE} /plugin.wasm
+        DATA GLOB ${CMAKE_CURRENT_SOURCE_DIR}/services/user/Webmail/ui/dist/* /
         INIT
+    DEPENDS ${Webmail_js_DEP} ${CMAKE_CURRENT_SOURCE_DIR}/services/user/Webmail/ui/dist/index.html
```
swatanabe, line 632: `.../index.html` is redundant, because `${Webmail_js_DEP}` already includes it.

brandonfancher: So should that line simply be `DEPENDS ${Webmail_js_DEP}`?

services/user/Webmail/ui/src/components/nav.tsx
```diff
@@ -0,0 +1,96 @@
...
+                        <a
+                            key={index}
+                            href={link.href}
+                            className={cn(
+                                buttonVariants({
+                                    variant: link.variant,
+                                    size: "sm",
+                                }),
+                                link.variant === "default" &&
+                                    "dark:bg-muted dark:text-white dark:hover:bg-muted dark:hover:text-white",
```
Velua, line 73: Would variables work here? Just wondering if we can stop all use of hard coding colours like this and go all in on [variables](https://ui.shadcn.com/docs/theming#list-of-variables)

Velua: Especially if we're gonna change our theming later on, would be awesome changing a bunch of config files across several apps rather than looking for class names we wrote.

services/user/Webmail/ui/src/hooks/use-mail.ts
brandonfancher: Good idea. I'll add that to the fast follow in the description above.

services/user/Webmail/ui/src/components/nav.tsx
brandonfancher: Some of these are variables, aren't they? Like `muted`? Or maybe I don't understand.
See: https://github.com/gofractally/psibase/pull/798/files#diff-d9598b721971bbf6eac63ac68007806de62deab4c4e6075b5e9e470dcbae3ccdR5-R36

Whatever the case, I'm fine with harmonizing color treatment across apps. This was just path of least resistance to something demoable.

CMakeLists.txt
swatanabe: yes

James-Mart:
I like the name chainmail
 ⛓️ 📫

brandonfancher:
> I like the name chainmail ⛓️ 📫

Me too. Credit for that name goes to @swatanabe.

services/user/Webmail/ui/src/main.tsx
```diff
@@ -0,0 +1,60 @@
...
+import { Drafts, Editor, Home, Sent, Viewer } from "@routes";
+import { TooltipProvider } from "@shadcn/tooltip";
+
+import DefaultLayout from "./layouts/default";
+
+import "./styles/globals.css";
+
+const queryClient = new QueryClient();
+
+const router = createHashRouter([
```
James-Mart, line 17: In the next version, we can use an SPA instead of a hash router, we just need to update the http server to return the same index for all requests, except for perhaps some whitelisted few API endpoints (like `messages` and the standard endpoints for `action_templates`, etc.)

services/user/Webmail/ui/src/components/nav.tsx
James-Mart: Can these be dynamic? Such that a `theme` app could be used to set the value of the theme variables across all system apps (or potentially even user apps that want to conform to style guidelines?)

services/user/Webmail/ui/vite.config.ts
James-Mart, line 1: Definitely need to audit this at some point, I think there's a fair bit of outdated stuff in here

services/user/Webmail/ui/src/components/editor/index.ts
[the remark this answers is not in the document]
brandonfancher: Yes. Mermaid has an issue right now where it breaks the ui bundle build in production mode. Known issue. LaTeX/KaTeX bundles tons of large font files and extra JS and CSS files. Plus, if there's any vector for script injection, it's that. Happened to [github recently](https://stevemats.medium.com/css-injection-on-github-profiles-from-unicode-exploits-to-new-bypass-techniques-f73f343f05d8). I'm sure there are mitigation possibilities, but mermaid and LaTeX makes more sense for technical social posts than it does emails inline.

services/user/Webmail/ui/vite.config.ts
brandonfancher: Agreed. I'll add that to the backlog in the description above. Those will become their own tasks at some point.

James-Mart (approved):

PR #805, opened 2024-08-08
Propagate graphql errors as errors instead of success response

cool-ant:
Supervisor was returning graphql errors as successes, which caused unrelated downstream issues because everything downstream was expecting good results but getting body = "errors": { ...error stuff... }
Now graphql errors are thrown in the JS context

### DIFF
--- a/services/user/Supervisor/ui/src/pluginHost.ts
+++ b/services/user/Supervisor/ui/src/pluginHost.ts
@@ -140,11 +140,11 @@ export class PluginHost implements HostInterface {
     ): Result<string, RecoverableErrorPayload> {
         const res = this.postGraphQL(`${this.self.origin}/graphql`, graphql);
         if ("body" in res) {
-            if (res.body.includes("errors")) { // works?
+            if (res.body.includes("errors")) {
                 const json = JSON.parse(res.body);
-                throw this.recoverableError(json.errors.message);
+                return this.recoverableError(json.errors.message);
             } else {
-                    return res.body;
+                return res.body;
             }
         }
         return res;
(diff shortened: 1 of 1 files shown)


services/user/Supervisor/ui/src/pluginHost.ts
```diff
@@ -140,7 +140,12 @@ export class PluginHost implements HostInterface {
     ): Result<string, RecoverableErrorPayload> {
         const res = this.postGraphQL(`${this.self.origin}/graphql`, graphql);
         if ("body" in res) {
-            return res.body;
+            if (res.body.includes("errors")) { // works?
+                const json = JSON.parse(res.body);
+                throw this.recoverableError(json.errors.message);
```
James-Mart, line 145: If you're going to throw instead of return, then you should `throw new Error`. 
`recoverableError` is `return`-ed when the error should be handleable (using `?` or equivalent) within the plugin code.

services/user/Supervisor/ui/src/pluginHost.ts
```diff
@@ -140,7 +140,12 @@ export class PluginHost implements HostInterface {
...
+            } else {
+                    return res.body;
```
James-Mart, line 147: Indentation

services/user/Supervisor/ui/src/pluginHost.ts
```diff
@@ -140,7 +140,12 @@ export class PluginHost implements HostInterface {
     ): Result<string, RecoverableErrorPayload> {
         const res = this.postGraphQL(`${this.self.origin}/graphql`, graphql);
         if ("body" in res) {
-            return res.body;
+            if (res.body.includes("errors")) { // works?
```
James-Mart, line 143: Remove comment

James-Mart (approved):

PR #809, opened 2024-08-13
Add getAccount() to Accounts plugin

cool-ant:
These are the changes made in the mm/identity-plugin branch as part of the identity service/plugin

### DIFF
--- a/services/system/Accounts/plugin/src/errors.rs
+++ b/services/system/Accounts/plugin/src/errors.rs
@@ -3,7 +3,7 @@ use crate::bindings::host::common::types::{Error, PluginId};
 #[derive(PartialEq, Eq, Hash)]
 pub enum ErrorType {
     NotYetImplemented,
-    InvalidAccountNumber,
+    AccountDNE,
     QueryError,
 }
 
@@ -22,15 +22,15 @@ impl ErrorType {
                 producer: my_plugin_id(),
                 message: format!("Not yet implemented: {}", msg),
             },
-            ErrorType::InvalidAccountNumber => Error {
+            ErrorType::AccountDNE => Error {
                 code: self as u32,
                 producer: my_plugin_id(),
-                message: format!("Invalid account number: {}", msg),
+                message: format!("Account number DNE: {}", msg),
             },
             ErrorType::QueryError => Error {
                 code: self as u32,
                 producer: my_plugin_id(),
-                message: format!("Failed to post graphql query: {}", msg),
+                message: format!("Graphql query error: {}", msg),
             },
         }
     }
--- a/services/system/Accounts/plugin/src/errors.rs
+++ b/services/system/Accounts/plugin/src/errors.rs
@@ -4,6 +4,7 @@ use crate::bindings::host::common::types::{Error, PluginId};
 pub enum ErrorType {
     NotYetImplemented,
     AccountDNE,
+    InvalidAccountName,
     QueryError,
 }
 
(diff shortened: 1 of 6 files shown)


services/system/Accounts/plugin/src/errors.rs
```diff
@@ -26,6 +27,11 @@ impl ErrorType {
                 producer: my_plugin_id(),
                 message: format!("Invalid account number: {}", msg),
             },
+            ErrorType::QueryError => Error {
+                code: self as u32,
+                producer: my_plugin_id(),
+                message: format!("Failed to post graphql query: {}", msg),
```
James-Mart, line 33: This QueryError says, "failed to post..." but it should probably just say something more general, like, "GraphQL query error: ..." since it's not only emitted when posting fails.

services/system/Accounts/plugin/src/lib.rs
```diff
@@ -29,6 +43,39 @@ impl Accounts for AccountsPlugin {
...
+            });
+
+        let account = account_result.and_then(|response_root: ResponseRoot| {
+            response_root
+                .data
+                .getAccount
+                .ok_or_else(|| InvalidAccountNumber.err(&name))
+        })?;
+
+        if account.accountNum.value != AccountNumber::from_exact(&name).unwrap().value {
```
James-Mart, line 65: Don't need the roundtrip check. Only names that roundtrip could ever be returned from this query.

services/system/Accounts/plugin/src/lib.rs
```diff
@@ -29,6 +43,39 @@ impl Accounts for AccountsPlugin {
...
+            .map_err(|e| QueryError.err(&e.message))
+            .and_then(|result| {
+                serde_json::from_str(&result).map_err(|e| QueryError.err(&e.to_string()))
+            });
+
+        let account = account_result.and_then(|response_root: ResponseRoot| {
+            response_root
+                .data
+                .getAccount
+                .ok_or_else(|| InvalidAccountNumber.err(&name))
```
James-Mart, line 62: Error should be `AccountDNE` rather than `InvalidAccountNumber`. Just because it doesn't exist doesn't mean it's an invalid account number.

James-Mart (changes requested):

services/system/Accounts/plugin/src/lib.rs
cool-ant: Not worth having the plugin do the work that will save node CPU / Account service time?

James-Mart: If that's the goal then it should be done before the query is made. At this point the query is complete so it's too late to check if it was a valid account name.
Also, if you keep it, you should change the error to say, 'invalid account *name*' rather than `number`.

services/system/Accounts/plugin/src/lib.rs
```diff
@@ -29,9 +43,40 @@ impl Accounts for AccountsPlugin {
...
+            .and_then(|result| {
+                serde_json::from_str(&result).map_err(|e| QueryError.err(&e.to_string()))
+            });
+
+        let account = account_result.and_then(|response_root: ResponseRoot| {
+            response_root
+                .data
+                .getAccount
+                .ok_or_else(|| AccountDNE.err(&name))
+        })?;
```
James-Mart, line 65: Sorry, just realized one more issue. This actually shouldn't throw an accountDNE error. This function returns a result<option<...>> so a valid return value is Ok(None)

Issue #753, opened 2024-06-17
Branding app


James-Mart:
# Introduction
 
psibase is the name of the protocol, but any actual deployment will want its own branding with a name, and a logo. We should create such a branding app sooner rather than later, so that our apps can stop hardcoding "psibase" and the psibase logo, and instead ask a plugin for the branding details and use those.

# Requirements

A page, accessible from the home page by any registered infrastructure provider, that allows them to update various details about the branding of the network. All details are stored on-chain.

The most important details for this page to include are:
* Network name
* Logo

PR #819, opened 2024-08-23
Rebranding App

cool-ant:
This app starts a chain with default values for chain name and image and allows the `branding` account to update both.
Fulfills #753


services/user/Branding/plugin/src/lib.rs
```diff
@@ -0,0 +1,56 @@
...
+            contentType: String::from("image/svg+xml"),
+            content: Hex(logo),
+        }
+        .packed();
+        CommonServer::add_action_to_transaction("storeSys", &packed_logo_args).unwrap();
+    }
+}
+
+#[derive(serde::Deserialize, Debug)]
+#[allow(non_snake_case)]
```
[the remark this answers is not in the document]
cool-ant: Wow. That's really slick. Thanks!

services/user/Branding/service/src/lib.rs
```diff
@@ -0,0 +1,76 @@
...
+                name: String::from("psibase"),
+            }
+        }
+    }
+
+    #[table(record = "WebContentRow", index = 1)]
+    struct WebContentTable;
+
+    #[action]
+    fn set_network_name(name: String) {
```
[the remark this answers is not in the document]
cool-ant: didn't know that. ok. thx

services/user/Branding/plugin/src/lib.rs
```diff
@@ -0,0 +1,56 @@
...
+impl Api for BrandingPlugin {
+    fn set_network_name(name: String) {
+        let packed_network_name_args = branding::action_structs::set_network_name { name }.packed();
+        CommonServer::add_action_to_transaction("set_network_name", &packed_network_name_args)
+            .unwrap();
+    }
+    fn set_logo(logo: Vec<u8>) {
+        let packed_logo_args = branding::action_structs::storeSys {
+            path: String::from("/network_logo.svg"),
+            contentType: String::from("image/svg+xml"),
```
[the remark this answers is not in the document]
cool-ant: - The ui should certainly check, and it's easy to do there. I'll add that.
- By the time we're calling the plugin, the image is already a Uint8Array, so the only "verification" we could do would be for the user to send a separate param with the file name OR go through the cost of decoding the ByteArray assuming it's an svg and look for a successful decoding, all of which I think is minimally helpful.
- The service would be where to do a final check, but again, at that point, we'd have to verify that the byte stream has meta data in it that says it's an SVG. Otherwise, we just store it, assuming it's right, and let the user bear the consequences if it's not.

As for determining SVG format directly from the bytestream, many SVGs start the stream with "<svg", but apparently, they can also be encoded in XML, which is far harder to interpret, leading us to decode the whole file to be sure. and a hacker could just set those initial bytes if the concern was security so...

Potential Concerns we could Address:
- Security: I don't think there's much we can do here other than don't do anything with the bytes but store and serve them.
- Spam: set a max upload size? i think tx/storage costs handle this.
- other?

Thoughts?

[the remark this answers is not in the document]
swatanabe: > so the only "verification" we could do would be for the user to send a separate param with the file name

The file name is unreliable

> OR go through the cost of decoding the ByteArray assuming it's an svg and look for a successful decoding, all of which I think is minimally helpful.

That's what I was thinking. Parsing XML is way too heavyweight to do in the service, so it needs to be done in either the plugin or the ui (I don't really have a strong opinion about which). My main concern is someone not realizing it has to be SVG, uploading a different image format, and getting very confused.

> many SVGs start the stream with "<svg", but apparently, they can also be encoded in XML

It's always XML. `<svg` *is* the beginning of an XML tag.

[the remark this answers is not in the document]
cool-ant: So, the UI *says* to upload an SVG. I could certainly check that what they selected was SVG and warn them, *but* important note: this is an admin app, so I haven't been aiming at putting this kind of polish on something that's an admin tool.
I'm mainly concerned with bigger issues like, e.g., security, which in this case, if we can only enforce effectively in the plugin, that leaves the service vulnerable, so I suggest we do nothing, given we can't (and I don't think need to) handle this thoroughly and completely.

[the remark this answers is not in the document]
James-Mart: Doing validation should ideally done in the plugin, unless it's for some reason significantly simpler to do on the front-end.
Maybe a crate like [this](https://docs.rs/nsvg/0.5.1/nsvg/struct.SvgImage.html#method.parse_str) could work?

services/user/Branding/service/src/lib.rs
```diff
@@ -0,0 +1,77 @@
...
+            }
+        }
+    }
+
+    #[table(record = "WebContentRow", index = 1)]
+    struct WebContentTable;
+
+    #[action]
+    #[allow(non_snake_case)]
+    fn networkNameChanged(name: String) {
```
James-Mart, line 37: Suggested tense convention:
* action: imperative (`setNetworkName`)
* event: present perfect (`networkNameChanged`)

services/user/Branding/service/src/lib.rs
```diff
@@ -0,0 +1,77 @@
...
+    }
+
+    #[event(history)]
+    pub fn networkNameChanged(name: String) {}
+
+    struct Query;
+
+    #[Object]
+    impl Query {
+        async fn network_name(&self) -> async_graphql::Result<String, async_graphql::Error> {
```
James-Mart, line 52: Seems like this can just return a string, no?

services/user/Branding/service/Cargo.toml
James-Mart, line 1: We should start using the new rust package capabilities ([docs](https://github.com/gofractally/psibase/blob/main/doc/psidk/src/development/services/rust-service/package.md)).

And then update `DevDefault` and `ProdDefault` packages in CMake with the new package.

services/user/Branding/plugin/src/errors.rs
```diff
@@ -0,0 +1,25 @@
+use crate::bindings::host::common::types::{Error, PluginId};
+
+#[derive(PartialEq, Eq, Hash)]
+pub enum ErrorType {
+    QueryResponseParseError,
+}
+
+fn my_plugin_id() -> PluginId {
+    return PluginId {
+        service: "identity".to_string(),
```
James-Mart, line 10: This should be changed to "branding"

James-Mart (dismissed):

services/user/Branding/plugin/src/lib.rs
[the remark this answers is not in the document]
cool-ant: Ok, I've added an `accept` attrib on the file picker as well as a check once selected.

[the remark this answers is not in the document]
cool-ant: Given this is an operator app, I'm not going to spend the time now adding a lib to detect SVG, especially given the first request we get from users is likely to support some format other than SVG. lol

services/user/Branding/service/src/lib.rs
James-Mart: Can you explain marking this as resolved? The tense still seems strange to me.

services/user/Branding/service/src/lib.rs
James-Mart: Marked as resolved but no action here either... when you mark as resolved if you're not going to do anything it would be helpful to leave a comment explaining your thinking

services/user/Branding/service/src/lib.rs
cool-ant: You're not looking at the most recent code? The diff shows the changes.

services/user/Branding/service/src/lib.rs
cool-ant: same here

services/user/Branding/ui/src/App.tsx
```diff
@@ -0,0 +1,214 @@
...
+                <div className="col-span-6 mt-6 grid grid-cols-6">
+                    <Label htmlFor="networkName" className="col-span-2">
+                        Network name
+                    </Label>
+                    <Input
+                        id="networkName"
+                        className="col-span-4"
+                        onChange={(e) => {
+                            setNetworkName(e.target.value);
+                            setChangesMade(true);
```
swatanabe, line 196: Once `changesMade` is set, it is never reset. Submitting successfully does not reset it. Changing the network name back to its original value doesn't reset it either.

services/user/Branding/service/src/lib.rs
```diff
@@ -0,0 +1,77 @@
...
+    struct Query;
+
+    #[Object]
+    impl Query {
+        async fn network_name(&self) -> String {
+            let curr_val = NetworkNameTable::new().get_index_pk().get(&SingletonKey {});
+            match curr_val {
+                Some(val) => val.name,
+                None => String::from("psibase"),
+            }
```
swatanabe, line 57: Since you've already impl'ed `Default`, you can use `unwrap_or_default` here.

services/user/Branding/service/src/lib.rs
```diff
@@ -0,0 +1,77 @@
...
+        }
+    }
+
+    #[action]
+    #[allow(non_snake_case)]
+    fn serveSys(request: HttpRequest) -> Option<HttpReply> {
+        None.or_else(|| serve_content(&request, &WebContentTable::new()))
+            .or_else(|| serve_graphql(&request, Query))
+            .or_else(|| serve_graphiql(&request))
+            .or_else(|| serve_simple_ui::<Wrapper>(&request))
```
swatanabe, line 67: Are these needed when we have a full UI?

services/user/Branding/service/src/lib.rs
cool-ant: Good point, but how would I write that concisely, given I need to grab the .name field after unwrapping (only of is it was a Some())?

services/user/Branding/service/src/lib.rs
cool-ant: graphiql, for debugging and whatnot (ensure that a failed call would pass if structured properly).
simple_ui? not really, but also could be useful for debugging.
Easy enough to add back in, I suppose.
It'd be useful to have an agreement among the team about when to include these or not.

services/user/Branding/ui/src/App.tsx
```diff
@@ -0,0 +1,214 @@
...
+    }, []);
+
+    const getNetworkName = async () => {
+        const queriedNetworkName = (await supervisor.functionCall({
+            service: "branding",
+            intf: "queries",
+            method: "getNetworkName",
+            params: [],
+        })) as string;
+        setNetworkName(queriedNetworkName);
```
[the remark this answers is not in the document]
cool-ant: Yes, but they'd be rushing to enter something in the < 1 sec before it gets populated. I could add fancier stuff like disabling the UI element until this call is finished, but again, this is not a traditionally user-facing app. Operators will use it *very* infrequently, so I don't see a need to get this as tight as other apps.

services/user/Branding/ui/src/App.tsx
```diff
@@ -0,0 +1,214 @@
...
+                        }
+                    >
+                        <img
+                            src={previewImgUrl}
+                            height="96"
+                            width="96"
+                            className="block h-24 w-24 object-cover"
+                        />
+                        <div className="text-extralight align-bottom text-sm">
+                            Preview
```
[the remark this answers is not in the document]
cool-ant: if you want to load a different image after a preview is generated, you just click the original logo again (same instructions as before you loaded the first logo).
Cmd-R / Ctrl-R will fix it for the operators who will use this.

services/user/Branding/ui/src/App.tsx
```diff
@@ -0,0 +1,214 @@
...
+            params: [],
+        })) as string;
+        setNetworkName(queriedNetworkName);
+    };
+    const updateAssets: React.MouseEventHandler<HTMLButtonElement> = async (
+        e,
+    ) => {
+        e.preventDefault();
+        try {
+            let res = await supervisor.functionCall({
```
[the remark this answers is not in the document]
cool-ant: good call. likely just relic code from debugging code that logged the res. not needed now.  🙏

services/user/Branding/service/src/lib.rs
swatanabe: `<Option<T>>::unwrap_or_default` returns `T`, so there is always a `name` field.

services/user/Branding/ui/src/App.tsx
```diff
@@ -0,0 +1,214 @@
...
+                res = await supervisor.functionCall({
+                    service: "branding",
+                    intf: "api",
+                    method: "setNetworkName",
+                    params: [networkName],
+                });
+            }
+
+            if (logoBytes.length) {
+                res = await supervisor.functionCall({
```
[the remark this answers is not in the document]
cool-ant: doesn't matter in this case. This allows for only one item to be updated without complication.
If this were a higher-profile, more-frequently used app, we'd have a single action that takes both netname and logoimage with some means of indicating no update, and it would put both in a single tx.

swatanabe (approved):

PR #831, opened 2024-09-04
Registry app - Phase 1

sparkplug0025:
Registry App

image


sparkplug0025:
@brandonfancher & @Velua can you please review the UI structure, the form hooks etc. To make sure it is aligned with the other frontend apps patterns?

@James-Mart package wip, need to sync with you!

services/user/Workshop/service/src/lib.rs
```diff
@@ -0,0 +1,138 @@
...
+            status,
+        };
+
+        app_metadata_table.put(&app_metadata).unwrap();
+
+        Wrapper::emit().history().setAppMetadata(app_metadata);
+    }
+
+    #[action]
+    fn storeSys(path: String, contentType: String, content: Hex<Vec<u8>>) {
```
James-Mart, line 76: `storeSys` (and therefore the `WebContentTable`) should be in the query service alongside `serveSys`.

services/user/Workshop/service/src/lib.rs
```diff
@@ -0,0 +1,138 @@
...
+        println!("{} {}", path, contentType);
+        store_content(path, contentType, content, &WebContentTable::new()).unwrap()
+    }
+
+    #[event(history)]
+    pub fn setAppMetadata(app_metadata: AppMetadata) {}
```
James-Mart, line 82: Suggested tense convention:

action: imperative (`setAppMetadata`)
event: present perfect (`appMetaChanged`)

services/user/Workshop/service/src/lib.rs
```diff
@@ -0,0 +1,138 @@
...
+#[allow(non_snake_case)]
+pub mod service {
+    use async_graphql::*;
+    use psibase::*;
+    use serde::{Deserialize, Serialize};
+
+    /// Holds metadata for a registered app
+    #[table(name = "AppMetadataTable", index = 0)]
+    #[derive(Debug, Clone, Fracpack, ToSchema, Serialize, Deserialize, SimpleObject)]
+    pub struct AppMetadata {
```
James-Mart, line 11: These weren't in the github issue (until right now because I just added them), but could you also add 
* `tags: List<String>` - Devs can add up to 3 tags that can further be used to categorize the application
* `created: TimePointSec` - This can be set with getting `currentBlock().time` from the `Transact` service. Tracks the unix time in seconds that the app was registered.

services/user/Workshop/query/src/lib.rs
```diff
@@ -0,0 +1,37 @@
...
+        async fn app_metadata(
+            &self,
+            account_id: AccountNumber,
+        ) -> Option<workshop::service::AppMetadata> {
+            workshop::service::AppMetadataTable::new()
+                .get_index_pk()
+                .get(&account_id)
+        }
+
+        async fn event(
```
James-Mart, line 22: You probably saw & copied this `event` query function somewhere else, but it's not very useful. I think we can probably delete this (and eventually remove it from everywhere else, too).

brandonfancher:
This looks good. I see that some of the UI and account selection controls were borrowed from webmail/chainmail. There is a newer in-progress (will be merged soon) branch of that that actually queries available accounts and changes the account with supervisor from the dropdown (vs. using the mocked out accounts list fixture.) That can be found in the `bf/chainmail-enhancements` branch. That also has a dark mode toggle. I recommend updating the account switcher from that if it's relevant to this app and bringing in the dark mode toggle.

UI code wise, everything looks great!

James-Mart:
Dark mode, pretty please!

services/user/Workshop/service/src/lib.rs
sparkplug0025: when I read query my first instinct was that it is a read only service (I thought the same when I saw the example named `r-<package>`). If we are adding the storesys it would go against a query or a read only service. So what is the definition of the query service on the package?

swatanabe: A query service's primary job is to handle HTTP requests. In this mode it is read-only. The data is stored in the query service because it is only used for the HTTP interface. It's essentially static content associated with the service.

James-Mart: I think I might prefer calling it an "http-server" or something. Query / RPC service don't seem as readily understandable.

sparkplug0025:
@swatanabe can you please help me troubleshoot why I'm receiving the error below when running `cargo package` from the [Workshop root folder](https://github.com/gofractally/psibase/pull/831/files#diff-f262a9b61b1a02c66cb1518fc0f20e9d86e9e74b6366a5d0564f986296dc26c6)?

`error: r-workshop is not a dependency of workshop`

swatanabe:
> @swatanabe can you please help me troubleshoot why I'm receiving the error below when running `cargo package` from the [Workshop root folder](https://github.com/gofractally/psibase/pull/831/files#diff-f262a9b61b1a02c66cb1518fc0f20e9d86e9e74b6366a5d0564f986296dc26c6)?
> 
> `error: r-workshop is not a dependency of workshop`

It's triggered by `server = "r-workshop"`

When trying to resolve "r-workshop" we look at the workspace members and the immediate dependencies of "workshop". "r-workshop" isn't in either.

leordev:
@James-Mart @Velua the set/get with the icon, tags autocomplete, owners and redirect uris is working!

sparkplug0025:
@Velua @Velua the set/get form is working with the icon, tags autocomplete, owners and the redirect URIs!

services/user/Registry/Cargo.toml
```diff
@@ -0,0 +1,36 @@
...
+rust-version = "1.64"
+edition = "2021"
+publish = false
+repository = "https://github.com/gofractally/psibase"
+homepage = "https://psibase.io"
+
+[package]
+edition = "2021"
+name = "registry_package"
+version = "0.1.0"
```
James-Mart, line 16: I think you need the edition and version keys, but I suggest you use `edition.workspace = true` and `edition.version = true` to avoid duplicating the values

services/user/Registry/Cargo.toml
```diff
@@ -0,0 +1,36 @@
...
+
+[package.metadata.psibase]
+package-name = "RegistryPackage"
```
James-Mart, line 19: The package shouldn't have "Package" in the name. That would be like adding "Crate" into the name of rust crates.
I think you should call the package "AppRegistry" (regardless that the service is only called, "registry").

services/user/Registry/plugin/src/lib.rs
James-Mart, line 1: Bit of organization in this file wouldn't hurt, although I won't hold up this PR for it. But best practice is probably to split types and type-cast implementations, graphql stuff, etc. out into separate modules.

James-Mart, line 1: Also, all the `println`'s (and those that are commented out) should be removed unless they have a specific purpose

services/user/Registry/plugin/src/lib.rs
```diff
@@ -0,0 +1,217 @@
...
+impl Developer for RegistryPlugin {
+    fn set_app_metadata(metadata: AppMetadata) -> Result<(), CommonTypes::Error> {
+        let owners_vec = metadata
+            .owners
+            .iter()
+            .map(|owner| owner.parse().unwrap())
+            .collect();
+        Transact::add_action_to_transaction(
+            "setMetadata",
+            &RegistryService::action_structs::setMetadata {
```
James-Mart, line 107: Another organization suggestion:
`impl From<AppMetadata> for RegistryService::action_structs::setMetadata`

services/user/Registry/query/src/lib.rs
```diff
@@ -0,0 +1,35 @@
...
+    use ::registry::service;
+    use ::registry::Wrapper as ServiceWrapper;
+
+    pub struct Query;
+
+    #[Object]
+    impl Query {
+        /// Get the app metadata for a specific app ID
+        async fn app_metadata(&self, account_id: AccountNumber) -> Option<service::AppMetadataWithTags> {
+            println!("getting app_metadata: {}", account_id);
```
James-Mart, line 15: Remove `println`

James-Mart (commented):
Check CMakeLists and copy what is done for `Chainmail.psi`.  Specifically, this registry should have a `cargo_psibase_package` entry in CMakeLists, and that package should also be added to `DevDefault` and `ProdDefault` packages.

Also, treat my review comments here about organization as TODOs. Not going to hold up the PR for them, but you could improve the organizational quality in a follow-up pr or something.

services/user/Registry/service/src/lib.rs
```diff
@@ -0,0 +1,378 @@
...
+
+    fn remove_obsolete_tags(existing_app_tags: &[AppTags], new_app_tags: &[AppTags]) {
+        let app_tags_table = AppTagsTable::new();
+        for existing_tag in existing_app_tags {
+            if !new_app_tags
+                .iter()
+                .any(|at| at.tag_id == existing_tag.tag_id)
+            {
+                app_tags_table.remove(existing_tag);
+                // TODO: Check if there are any apps using the tag from tags_table and remove if not
```
James-Mart, line 325: Also a good idea.

Can a range query hold the second tuple element constant and modulate the first? If so then I think it should be as easy as:

```
if app_tags_table
    .get_index_pk()
    .range((0, old_tag.tag_id)..(u64::MAX, old_tag.tag_id))
    .count() == 0 {
        tags_table.remove(old_tag.tag_id);
    }
```

Actually, you probably want a secondary index for app_tags_table that reverses the tuple order anyway, to make filtering by tag more efficient.

services/user/Registry/service/src/lib.rs
```diff
@@ -0,0 +1,378 @@
...
+        let from = partial_tag.to_lowercase();
+        let excluded_to = increment_last_char(from.clone());
+
+        let tags: Vec<String> = idx
+            .range((from, 0)..(excluded_to, 0))
+            .take(10)
+            .map(|tag_row| tag_row.tag)
+            .collect();
+
+        Some(RelatedTags { tags })
```
James-Mart, line 228: This always returns Some even if `tags.len() == 0`

Velua (approved):

services/user/Registry/Cargo.toml
```diff
@@ -0,0 +1,35 @@
...
+resolver = "2"
+members = ["service", "query", "plugin"]
+
+[workspace.package]
+version = "0.1.0"
+rust-version = "1.64"
+edition = "2021"
+publish = false
+repository = "https://github.com/gofractally/psibase"
+homepage = "https://psibase.io"
```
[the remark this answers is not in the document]
James-Mart: Hah, that's probably true that we shouldn't reference `psibase.io`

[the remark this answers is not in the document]
sparkplug0025: Here I followed the Branding and Chainmail package config pattern. Should I just remove the `homepage`  line? Is it fine to leave the repo url as is?

services/user/Registry/Cargo.toml
sparkplug0025: I did it and now I'm receiving these warnings (when I run cargo psibase package, test etc.):
```
warning: /home/sohdev/Workspace/fractally/psibase/services/user/Registry/Cargo.toml: unused manifest key: package.edition.version
```

This is how the file is looking like:
```
[workspace]
resolver = "2"
members = ["service", "query", "plugin"]

[workspace.package]
version = "0.1.0"
rust-version = "1.64"
edition = "2021"
publish = false
repository = "https://github.com/gofractally/psibase"
homepage = "https://psibase.io"

[package]
name = "registry_package"
edition.workspace = true
edition.version = true

[package.metadata.psibase]
package-name = "AppRegistry"
description = "Registry for psibase apps"
services = ["registry"]

[profile.release]
codegen-units = 1
opt-level = "s"
debug = false
strip = true
lto = true

[lib]
crate-type = ["rlib"]

[dependencies]
registry = { path = "service" }
r-registry = { path = "query" }
registry-plugin = { path = "plugin" }
```

Is that correct?

services/user/Registry/Cargo.toml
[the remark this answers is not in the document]
James-Mart: Yeah I'd either delete it or set the homepage also to the repository page. Chainmail and Branding should probably do the same

services/user/Registry/Cargo.toml
swatanabe: > `unused manifest key: package.edition.version`

I think you meant `version.workspace`

services/user/Registry/Cargo.toml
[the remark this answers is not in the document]
swatanabe: I don't think `psibase.io` is wrong here. I would expect the home page for an app to be owned by the app developer (in this case, us). The fact that our website is hosted on a specific chain is incidental.

services/user/Registry/README.md
```diff
@@ -0,0 +1,18 @@
+# Psibase Registry
+
+Steps to run the app in dev mode:
+
+```sh
+# after building psibase at the root of the repo
+cd build
+./psinode --http-timeout 10 db
```
swatanabe, line 8: Is the value of 4 in the default config template too short? If we need to clutter docs like this without explanation, it would be better just to increase the default.

services/user/Registry/plugin/src/lib.rs
```diff
@@ -0,0 +1,217 @@
...
+}
+
+#[allow(non_snake_case)]
+#[derive(Deserialize, Debug)]
+struct AppMetadataResponseData {
+    metadata: ServiceAppMetadata,
+    tags: Vec<TagRecord>,
+}
+
+#[allow(non_snake_case)]
```
swatanabe, line 41: Another way is to use the regular rust naming convention and change the serialization with `#[serde(rename_all = "camelCase")]`

services/user/Registry/service/src/utils.rs
```diff
@@ -0,0 +1,10 @@
+/// Increment the last character of a string
+///
+/// Example: "abc" -> "abd"
+pub fn increment_last_char(s: String) -> String {
+    let mut chars: Vec<char> = s.chars().collect();
+    if let Some(last) = chars.last_mut() {
+        *last = char::from_u32((*last as u32) + 1).unwrap_or(*last);
```
swatanabe, line 7: Unfortunately this doesn't always work, e.g. if the last character is U+D7FF.

services/user/Registry/service/src/lib.rs
```diff
@@ -0,0 +1,378 @@
...
+
+        /// The name of the tag
+        pub tag: String,
+    }
+
+    impl TagRecord {
+        #[secondary_key(1)]
+        fn by_tags(&self) -> (String, u32) {
+            (self.tag.clone(), self.id)
+        }
```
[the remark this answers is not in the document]
swatanabe: If you leave the id off, it will enforce uniqueness.

services/user/Registry/README.md
sparkplug0025: not sure, I cant test anymore because we removed the UI... 
I pretty much just copied and pasted how I run my cmd, let me remove this arg.

services/user/Registry/service/src/utils.rs
sparkplug0025: Since we are restricting the chars to a-z and 0-9 the function is safe. Let me add a explicit comment about it.

services/user/Registry/Cargo.toml
James-Mart: Oops, my brain shut of momentarily as I wrote the comment before @sparkplug0025, I meant `version.workspace`  but wrote `edition.version` haha. I updated the original comment to make it correct

services/user/Registry/service/src/lib.rs
```diff
@@ -0,0 +1,697 @@
...
+            DbId::HistoryEvent,
+            SERVICE,
+            MethodNumber::from("appMetadataChanged"),
+            0,
+        );
+        EventsSvc::call().addIndex(
+            DbId::HistoryEvent,
+            SERVICE,
+            MethodNumber::from("appStatusChanged"),
+            1,
```
James-Mart, line 285: I think what you want:
* IS a historical view of all status changes for a particular app
* is NOT a historical view of all apps that historically had a particular status

Right? If so, it implies that you want col 0 to be the index for this `appStatusChanged` event.

services/user/Registry/service/src/lib.rs
```diff
@@ -0,0 +1,697 @@
...
+            .range(from..excluded_to)
+            .take(10)
+            .map(|tag_row| tag_row.tag)
+            .collect();
+
+        RelatedTags { tags }
+    }
+
+    #[event(history)]
+    fn appMetadataChanged(app_metadata: AppMetadata) {}
```
James-Mart, line 445: This is a big event, actually, and not sure how useful it is. I think most similar web apps don't give you a historical view of past descriptions and such. Not to mention you're duplicating data in each event that is impossible to change (`created_at`).

Honestly I'd probably remove this event entirely and just add specific events for specific metadata fields that we expect to want to want to track historically. Which... the only one I can think of is status, and you already have a specific event for that. So maybe just delete this event unless you've got other reasons for it

services/user/Registry/README.md
James-Mart: Not sure why we raced to delete the UI, it wouldn't have been terrible to have two UIs for a while, especially while Registry was still in development

services/user/Registry/Cargo.toml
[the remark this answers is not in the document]
James-Mart: > I would expect the home page for an app to be owned by the app developer (in this case, us). The fact that our website is hosted on a specific chain is incidental.

That's typically true, and I don't feel strongly about this. But
1. `psibase.io` is neither a company homepage nor a homepage for the registry app
2. In this case it's probably still more prudent to remove it to maximize the perceived distance between the company and any particular network instantiation.

James-Mart (approved):

services/user/Registry/service/src/utils.rs
swatanabe: In that case, you can avoid copying the entire string by using `into_bytes`/`from_utf8`.

PR #836, opened 2024-09-05
Fix warnings

James-Mart:
Silences two of the warnings I noticed in build output

### DIFF
--- a/libraries/net/include/psibase/blocknet.hpp
+++ b/libraries/net/include/psibase/blocknet.hpp
@@ -202,9 +202,7 @@ namespace psibase::net
             }
          }
          assert(!"Unknown peer connection");
-
-         // Silence warning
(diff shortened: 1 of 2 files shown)


libraries/net/include/psibase/blocknet.hpp
```diff
@@ -202,6 +202,9 @@ namespace psibase::net
             }
          }
          assert(!"Unknown peer connection");
+
+         // Silence warning
+         return *std::make_unique<peer_connection>(0);
```
swatanabe, line 207: Adding nonsense code is a poor way to remove warnings. Try `__builtin_unreachable`

James-Mart: Ah, nice. Fixed

swatanabe (approved):

PR #837, opened 2024-09-05
add some concepts to blob compare

James-Mart:
Just for fun/practice.

Also increases readability, probably improves error messages...

### DIFF
--- a/libraries/psibase/common/include/psibase/blob.hpp
+++ b/libraries/psibase/common/include/psibase/blob.hpp
@@ -4,16 +4,29 @@
 #include <algorithm>
 #include <type_traits>
 
+template <typename T>
+concept IsByteType =
+    std::is_same_v<std::decay_t<T>, char> || std::is_same_v<std::decay_t<T>, unsigned char>;
+
+template <typename T>
+concept IsContainerLike = requires(T t) {
+   {
+      t.data()
+   } -> std::convertible_to<const void*>;
+   {
+      t.size()
+   } -> std::convertible_to<std::size_t>;
+};
+
 namespace psibase
 {
    // Compare vector, string, string_view, shared_string (unsigned)
    template <typename A, typename B>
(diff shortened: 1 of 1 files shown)


libraries/psibase/common/include/psibase/blob.hpp
```diff
@@ -4,16 +4,29 @@
 #include <algorithm>
 #include <type_traits>
 
+template <typename T>
+concept IsByteType =
+    std::is_same_v<std::decay_t<T>, char> || std::is_same_v<std::decay_t<T>, unsigned char>;
```
swatanabe, line 9: `std::decay_t` should be applied outside this concept. I don't think that `IsByteType<char&>` should be true. Also, `remove_cvref` is usually preferable to `decay`.

libraries/psibase/common/include/psibase/blob.hpp
```diff
@@ -4,16 +4,29 @@
...
+
+template <typename T>
+concept IsContainerLike = requires(T t) {
```
swatanabe, line 12: Should use `const&`.

swatanabe, line 12: This name is too generic for what it actually checks. Actually, if we're going to use concepts, it should be `std::ranges::contiguous_range<T> && std::ranges::sized_range<T>` (and use `std::ranges::data`/`std::ranges::size`).

libraries/psibase/common/include/psibase/blob.hpp
James-Mart: done/done

libraries/psibase/common/include/psibase/blob.hpp
James-Mart: no longer relevant

James-Mart: This should now be addressed, what do you think?

swatanabe (approved):

PR #839, opened 2024-09-05
Updating X-admin's yarn.lock so we don't see it updated every time we build

cool-ant:
Not sure this will stop the problem, but it seems to be the only one that gets updated eveyr build, so worth a shot.


brandonfancher (approved):

James-Mart:
Thank the lord

James-Mart:
> The version committed in this PR matches yarn 1.22.22. The previous version was what I was seeing with yarn 1.22.21.

Perhaps we should align on a specific yarn version. The builders currently just use `npm i -g yarn`

PR #840, opened 2024-09-06
Add postinstall and data to cargo psibase package

swatanabe:
I think that `yarn` can be run via `build.rs`, but I haven't tried it.


James-Mart:
😢 https://github.com/gofractally/psibase/commit/8fcb4947ea78440edb00390fbcfe872dbcc27380

swatanabe:
> 😢 [8fcb494](https://github.com/gofractally/psibase/commit/8fcb4947ea78440edb00390fbcfe872dbcc27380)

It's rather odd that we thought that it would be easier to write in CMake than in rust...

James-Mart (approved):

PR #841, opened 2024-09-09
Chainmail UI enhancements

brandonfancher:
- Empty UI states
- Fetch available accounts
- Send from accounts other than `alice`
- Dark mode
- Success and error messages on send
- UI fixes and improvements


services/user/Webmail/ui/src/layouts/default.tsx
```diff
@@ -6,38 +6,43 @@ import {
...
-                    collapsible={true}
-                    minSize={15}
-                    maxSize={20}
-                    onCollapse={() => setIsCollapsed(true)}
-                    onExpand={() => setIsCollapsed(false)}
-                    className={cn(
-                        isCollapsed &&
-                            "min-w-[50px] transition-all duration-300 ease-in-out",
-                    )}
+        <ThemeProvider defaultTheme="dark" storageKey="vite-ui-theme">
```
cool-ant, line 18: Not sure what's actually happening in this module, but does this defaultTheme need to be "dark" or "light"? Or should it also be "system"?

cool-ant (approved):

services/user/Webmail/ui/src/layouts/default.tsx
brandonfancher: We've been going with `dark` as the `defaultTheme` across psibase apps.

PR #853, opened 2024-09-19
Renaming Webmail to Chainmail

cool-ant:
This touched a lot of files, so splitting out the rename from the archiving functionality PR


brandonfancher (approved):

services/user/Branding/ui/src/App.tsx
```diff
@@ -20,7 +20,7 @@ export const App = () => {
     const init = async () => {
         await supervisor.onLoaded();
         supervisor.preLoadPlugins([{ service: "branding" }]);
-        await getNetworkName();
+        setTimeout(getNetworkName, 1500);
```
James-Mart, line 23: Does awaiting not work?

James-Mart: @cool-ant

cool-ant: Good catch; I did that for the demo to emulate what won't work now because... ?Supervisor?
So no, it wasn't working.
I'll create a PR to reverse this change.

cool-ant: Nothing to do with Supervisor. I just couldn't verify this yet without post-install steps for chainmail

Issue #821, opened 2024-08-27
Chainmail: archiving and persisting messages


brandonfancher:
Emails are emitted as events. They aren't stored in state.

We will not support _deleting_ email messages; user's can't delete messages that are kept in the event history. Instead, we will support _archiving_ messages. There will be an archive event emitted. Queries via the plugin can join that information to the message.

Because messages are not stored in state and some nodes may opt to prune the event log, messages may disappear at any time. This is not always predictable. Yes, there may be archival nodes that pop up. On those nodes, the user can expect messages to persist. But if the user is not on an archival node, they may lose visibility into older/pruned messages.

For this reason, we will implement an aggressive 10-day auto-archive feature. Users can manually archive a message, but any message older than 10 days will not display in the user's inbox.
 - UI should display last 10 days of messages to inbox (need timestamp for this?). This allows different UIs to use same plugin/service. [get time from inline call to Transact service -- blocktime]

The chainmail app will have an "Archive" view. There, a user can see their non-pruned archived messages. The plugin will expose a query to the UI to fetch those. That "Archive" should prominently display a banner informing the user that archived emails are subject to purging at the discretion of the node operator.

The user, however, may want to save emails--preventing them from being purged or automatically archived. This would be a separate action that commits the given message to the chain state. This could feature as a bookmark toggle on the message itself. Of course, unsave should also exist. #855 

Cases:
- save archived message => save
- save inbox message
- archiving saved message => unsave, archive
What out for:
- query when messages have been pruned

PR #854, opened 2024-09-19
Chainmail: implement archive functionality in service

cool-ant:
Addresses #821 (archive; save will be a separate item)
Notes:
- This PR is part 1 of 2, second being the save/unsave PR #856 , which will be an immediate follow-on and which explains some of the following items
- There are some comments in this PR related to #856 
- There are save/unsave placeholders related to #856 
- There are a few println statement that will also be cleaned up in #856


services/user/Chainmail/plugin/src/lib.rs
```diff
@@ -2,13 +2,47 @@
 mod bindings;
 
 use bindings::exports::chainmail::plugin::api::{Error, Guest as API};
+use bindings::exports::chainmail::plugin::queries::{Guest as QUERY, Message};
```
James-Mart, line 5: Convention for trait capitalization in rust would be `Api` and `Query`

brandonfancher (approved):

services/user/Chainmail/service/src/lib.rs
```diff
@@ -74,25 +68,61 @@ fn serve_rest_api(request: &HttpRequest) -> Option<HttpReply> {
...
-        return REventsSvc::call().serveSys(mq);
+
+        println!("query: {}", sql_query_str);
+
+        let query_response = REventsSvc::call().sqlQuery(sql_query_str);
+        println!("query_response: {}", query_response);
+
+        return Some(HttpReply {
+            status: 200,
+            contentType: request.contentType.clone(),
```
James-Mart, line 122: No `contentType` header is set on the HTTP request that originates from the `getJson` plugin request. You should probably just use "application/json" explicitly.

services/user/Chainmail/service/src/lib.rs
```diff
@@ -74,25 +68,61 @@ fn serve_rest_api(request: &HttpRequest) -> Option<HttpReply> {
...
+            },
+            where_clause_sender_receiver,
+            order_by_clause
         );
-        return REventsSvc::call().serveSys(mq);
+
+        println!("query: {}", sql_query_str);
+
+        let query_response = REventsSvc::call().sqlQuery(sql_query_str);
+        println!("query_response: {}", query_response);
```
James-Mart, line 118: Is there a reason to leave all these `println` statements in for now?

services/user/Chainmail/ui/src/components/mail-display.tsx
```diff
@@ -91,33 +92,138 @@ const ActionBar = ({
...
+        let id = parseInt(itemId);
+        const supervisor = await getSupervisor();
+        // TODO: Unsave if saved
+        // await supervisor.functionCall({
+        //     service: "chainmail",
+        //     intf: "api",
+        //     method: "unsave",
+        //     params: [id],
+        // });
+        // TODO: Improve error detection. This promise resolves with success before the transaction is pushed.
```
James-Mart, line 105: Is it? The tx is not _applied_ before promise resolution, but the tx is supposed to be pushed as part of the resolution of the plugin call (before resolving this promise).

services/user/Sites/src/Sites.cpp
James-Mart, line 1: This should be reverted. It is not needed since #881 was merged

services/user/Chainmail/plugin/src/lib.rs
```diff
@@ -1,15 +1,28 @@
...
 
-impl API for ChainmailPlugin {
+#[derive(Debug, Deserialize)]
+struct MessageSerde {
+    msg_id: String,
+    receiver: String,
+    sender: String,
+    subject: String,
+    body: String,
+}
```
James-Mart, line 23: You can add 
```
[package.metadata.component.bindings]
derives = ["serde::Deserialize"]
```
To the plugin's `Cargo.toml` in order for the `Message` struct generated bindings to derive from `Deserialize`.

services/user/Chainmail/service/src/lib.rs
cool-ant: As stated in the PR comments, they will all be removed in the imminent save/unsave PR.

services/user/Sites/src/Sites.cpp
cool-ant: reverted in #885

Issue #855, opened 2024-09-19
Chainmail: persisting messages


cool-ant:
The user may want to save emails--preventing them from being purged or automatically archived. This would be a separate action that commits the given message to the chain state. This could feature as a bookmark toggle on the message itself. Of course, unsave should also exist.

James' thoughts 2024.09.27
I think sent messages should be stored in/displayed from localstorage, rather than from a chain query.
Because once we add message encryption, the sent messages will appear garbled since the sender can't decrypt them. And if they are shown from localstorage, then they won't actually disappear like received messages unless we introduce a localstorage clean-up policy.
[10:43](https://gofractally.slack.com/archives/C02SQDUM7JM/p1727451819808639)
And if you "save" a sent message, then it would actually re-encrypt the sent message with your own keys, so you can continue to see and read the sent message

on get_msgs(), read saved messages and union them into the return set
 - add saved flag to saved messages
add get_saved_msgs() (not a separate call; just a flag?)

remove `unsave` from plugin

host::Server could parse and add PageInfo
 - Events service needs GQL support

Later: paginate later

PR #856, opened 2024-09-23
Chainmail: implement save/unsave message #855

cool-ant:
Implements/fulfills #855


doc/psidk/src/development/services/rust-service/reference/chicken-scratch.md
James-Mart, line 1: Some of this doesn't belong in the mdbook, but instead belongs in a github issue.

In general, the doc updates should be split out from this chainmail save/unsave message implementation, lest you hold up the chainmail update for a much longer time than necessary. In fact, I'd recommend you split the doc updates themselves into many/multiple individual PRs so they can be independently reviewed on where the new docs get inserted and their content

services/user/Chainmail/service/src/event_query_helpers.rs
```diff
@@ -0,0 +1,111 @@
...
+
+    let itr = query.split("&");
+    for p in itr {
+        let kv = p.split_once("=").unwrap();
+        params.insert(kv.0.to_string(), kv.1.trim_start_matches('=').to_string());
+    }
+    params
+}
+
+fn get_where_clause_from_sender_receiver_params(params: HashMap<String, String>) -> Option<String> {
```
James-Mart, line 31: You don't need "from_sender_receiver_params" in the function name, because it would already read `get_where_clause(params)` so it's clear it's from the params

services/user/Chainmail/service/src/event_query_helpers.rs
```diff
@@ -0,0 +1,111 @@
...
+        let query_start = query_start.unwrap();
+
+        let query = request.target.split_at(query_start + 1).1;
+        let params = parse_query(query);
+
+        let where_clause: String;
+        let mq: HttpRequest;
+        // handle id param requests (for specific message)
+        if params.contains_key("id") {
+            where_clause = format!("WHERE CONCAT(receiver, rowid) = {}", params.get("id")?);
```
James-Mart, line 87: Is this ID actually fixed? If the event database is pruned, then the email this refers to would probably shift. Probably better to just add an auto-incrementing ID field for each email.

services/user/Chainmail/plugin/src/lib.rs
```diff
@@ -22,6 +32,35 @@ impl API for ChainmailPlugin {
...
+        let res = server::get_json(&format!("/messages?id={}", event_id))?;
+
+        let msg = serde_json::from_str::<Message>(&res)
+            .map_err(|err| ErrorType::QueryResponseParseError.err(err.to_string().as_str()))?;
+
+        // save the message to state
+        Transact::add_action_to_transaction(
+            "save",
+            &chainmail::action_structs::save {
+                receiver: msg.receiver,
```
James-Mart, line 56: If you're saving a sent message, then it makes sense to save the receiver. If you're saving a received message, then it makes sense to save the sender. So you probably want to always save both sender + receiver.

services/user/Chainmail/plugin/wit/world.wit
```diff
@@ -5,6 +5,12 @@ interface api {
 
     // Send an email
     send: func(receiver: string, subject: string, body: string) -> result<_, error>;
+    
+    // Archive an email
+    archive: func(event-id: u64) -> result<_, error>;
+    
+    // Save an email
+    save: func(event-id: u64) -> result<_, error>;
```
James-Mart, line 13: Missing unsave

services/user/Chainmail/service/src/tables.rs
```diff
@@ -0,0 +1,20 @@
...
+
+impl TableRecord for SavedMessage {
+    type PrimaryKey = u64;
+
+    const SECONDARY_KEYS: u8 = 0;
+
+    fn get_primary_key(&self) -> Self::PrimaryKey {
+        self.event_id
+    }
+}
```
James-Mart, line 12: Is this because you can't use #[primary_key] when the struct is split into a separate module?

services/user/Chainmail/service/src/lib.rs
```diff
@@ -121,8 +29,44 @@ mod service {
...
+            sender,
+            subject,
+            body,
+        })
+    }
+
     #[event(history)]
     pub fn sent(sender: AccountNumber, receiver: AccountNumber, subject: String, body: String) {}
+    #[event(history)]
+    pub fn archive(event_id: String) {}
```
James-Mart, line 69: If you add an email id field, you can make this `archive(account: AccountNumber, email_id: u64)`

services/user/Chainmail/service/src/event_query_helpers.rs
swatanabe: `rowid` is the event index, which is stable.

services/user/Chainmail/service/src/tables.rs
cool-ant: Yes, I'm pretty sure #[primary_key] is macro-replaced by this impl. This is just what it looks like when the macro can't do it for you

brandonfancher:
@cool-ant There is an unhandled TODO in `mail-list.tsx`. There's commented out UI there for displaying a pin on a message in the message in the list if it's saved. For that, we need to know from the message's metadata whether it is saved or not. Can you add a `saved` property?

services/user/Chainmail/service/src/event_query_helpers.rs
```diff
@@ -0,0 +1,137 @@
...
+        "SELECT {} FROM {} WHERE {} {} {} {}",
+        select_clause,
+        from_clause,
+        where_clause_archived_or_not,
+        if s_opt.is_some() || r_opt.is_some() {
+            "AND"
+        } else {
+            ""
+        },
+        if params.contains_key("id") {
```
swatanabe, line 93: Is it possible to have both `id` and either `sender` or `receiver`? I think that combination will produce `AND AND`.

services/user/Chainmail/plugin/src/serde_structs.rs
```diff
@@ -0,0 +1,37 @@
...
+pub struct TempMessageForDeserialization {
+    #[serde(deserialize_with = "deserialize_number_from_string")]
+    pub msg_id: u64,
+    pub receiver: String,
+    pub sender: String,
+    pub subject: String,
+    pub body: String,
+    #[serde(deserialize_with = "deserialize_timepoint")]
+    pub datetime: u32,
+}
```
James-Mart, line 17: Hmm, is it possible to deserialize with either:
* The `SavedMessage` struct in the service, or
* The generated `Message` object from your bindings?

Just seems a bit strange to essentially have three definitions of a Message type... Maybe it's needed though?

services/user/Chainmail/plugin/src/lib.rs
```diff
@@ -36,66 +35,71 @@ impl Api for ChainmailPlugin {
...
-    let mut resp_val: Vec<MessageSerde>;
-    if resp.is_err() {
-        return Err(errors::ErrorType::QueryResponseParseError.err(&resp.unwrap_err().to_string()));
-    } else {
-        resp_val = resp.unwrap();
+        Transact::add_action_to_transaction(
+            "save",
+            &chainmail::action_structs::save {
+                subject: msg.subject.clone(),
+                body: msg.body.clone(),
```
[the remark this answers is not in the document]
cool-ant: ah, interesting scenario. likely caused by debugging statements I had after this. Good thing to be vigilant about going forward. Thanks

services/user/Chainmail/service/src/lib.rs
```diff
@@ -158,29 +47,100 @@ mod service {
...
+    fn save(
+        msg_id: u64,
+        receiver: AccountNumber,
+        sender: AccountNumber,
+        subject: String,
+        body: String,
+        datetime: u32,
+    ) {
+        check(
+            get_sender() == receiver,
```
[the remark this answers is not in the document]
cool-ant: @James-Mart If I look up the message being referenced (by msg_id) and confirm that message has the other details specified int he args to the fn, then this check work, right? because we always know get_sender() is valid, and if we know the msg being referenced is valid, then this check works, right? So I could add the check for msg-by-id. If I can't find the msg, we should fail because either 1) this is a fake message being "saved" or 2) the message was pruned and so can't be saved.

services/user/Chainmail/plugin/src/serde_structs.rs
cool-ant: I agree it seems there should be a better way, and i'm pushing that direction, but I haven't found a good one yet. At least isolating these definitions as I hvae in this project starts to build the required output of a potential tweak to the service macro that could auto-generate local types, but I haven't found a good way yet.

I think it's also noteworthy when you say "three definitions", `SavedMessage` is a table record and could therefore significantly diverge from the FE of a message, so I'd argue is more generally only 2 distinct definitions. I'm trying to get those 2 to be the same. Haven't gotten there yet.

services/user/Chainmail/service/src/lib.rs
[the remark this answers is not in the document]
James-Mart: Events are in a write-only table.

Anyway, nothing wrong with people saving fake messages, since they are paying. And (as mentioned) we can add the ability to verify saved messages later.

services/user/Chainmail/plugin/src/queries.rs
```diff
@@ -0,0 +1,49 @@
...
+pub fn query_messages_endpoint(
+    sender: Option<String>,
+    receiver: Option<String>,
+    archived_requested: bool,
+) -> Result<Vec<Message>, Error> {
+    let mut endpoint = String::from("/api/messages?");
+    if archived_requested {
+        endpoint += "archived=true&";
+    }
+    if sender.is_some() {
```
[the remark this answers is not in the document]
cool-ant: understood, but this isn't *quite* that straight-forward given how I test for and use the value.

services/user/Chainmail/service/src/event_query_helpers.rs
cool-ant: reordered to avoid this. great catch. 🙏

services/user/Chainmail/service/src/lib.rs
```diff
@@ -158,29 +47,100 @@ mod service {
...
+                body,
+                datetime: TimePointSec::from(datetime),
+            })
+            .unwrap();
+    }
+
+    /// Unsave a message
+    /// `unsave` releases the state storage for a previously saved message
+    #[action]
+    fn unsave(msg_id: u64, sender: AccountNumber, subject: String, body: String, datetime: u32) {
```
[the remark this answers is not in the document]
cool-ant: Any reason not to leave this exposed as an action but not as part of the plugin API? Plugin API is meant to be the API for the frontend, which handles the bulk of what we discussed, but removing it from the service would potentially restrict a custom frontend use case. I'm inclined to leave this for that reason unless you strongly object.

services/user/Chainmail/service/src/tables.rs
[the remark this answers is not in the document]
cool-ant: @James-Mart Any reason you included msgid in the futurely-opaque payload? Keeping that as a secondary index (or primary actually) would allow us to correlate SavedMessages with Sent/Archived messages (as events) to know, for instance, whether or not they've been pruned or archived. As long as we encrypt the msg at all stages (including the payload to events), I don't see msg-id leaking any important information.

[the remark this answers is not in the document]
James-Mart: Yep, good call! I think msgid would be outside the message blob, And I think it's fine that there's no way to validate that the msgid is correct at the time you're saving the message, because worst case scenario is that only the person who submits an invalid payload to `save` has the view of their inbox screwed up

services/user/Chainmail/service/src/lib.rs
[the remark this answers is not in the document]
James-Mart: Nope, works for me

services/user/Chainmail/service/src/tables.rs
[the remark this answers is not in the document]
James-Mart: As discussed: The plan is to leave this as is for now, and then later when we add message encryption we may have to rework some of the service logic

PR #859, opened 2024-09-24
Default deploy chainmail

James-Mart:
This PR ensures that the ChainMail app is deployed to the chain by default.

To accomplish this, the [new functionality](https://github.com/gofractally/psibase/pull/840) in `cargo-psibase` that allows for UI and postinstall scripts to be specified is leveraged in chainmail.

Also, this PR introduces a new CMake function `cargo_psibase_package` that allows psibase devs to tell CMake about packages that are built using `cargo-psibase package` rather than using the CMake `psibase_package` function. Therefore cargo-psibase-managed packages can be integrated into the regular build system just like the other packages.

### DIFF
--- a/doc/psidk/src/development/services/rust-service/package.md
+++ b/doc/psidk/src/development/services/rust-service/package.md
@@ -7,7 +7,6 @@ A [psibase package](../../../specifications/data-formats/package.md) may be comp
 The available fields are:
 
 - `package-name`: Must be present on the top-level crate for the package.
-- `description`: A description of the package.
 - `services`: A list of services to include in the psibase package.
 - `server`: May be present on any crate that builds a service. The value is a crate which will handle HTTP requests sent to this service. The other crate will be built and included in the current package.
 - `plugin`: May be present on any crate that builds a service. The value is a crate that should be built with `cargo component` and uploaded as `/plugin.wasm`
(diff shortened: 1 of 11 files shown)


doc/psidk/src/development/services/rust-service/package.md
```diff
@@ -7,6 +7,7 @@ A [psibase package](../../../specifications/data-formats/package.md) may be comp
 The available fields are:
 
 - `package-name`: Must be present on the top-level crate for the package.
+- `description`: A description of the package.
```
swatanabe, line 10: `description` and `version` are taken from `[package]`, not `[package.metadata.psibase]`

James-Mart: Oh. I'll remove this and also move the description field in the `test_package` to `[package]`.

James-Mart: Done

libraries/psibase/sdk/pack_service.cmake
```diff
@@ -306,3 +306,47 @@ function(psibase_package)
...
+        INSTALL_COMMAND ""
+        BUILD_ALWAYS 1
+        DEPENDS ${ARG_DEPENDS} cargo-psibase
+    )
+
+    add_custom_command(
+        OUTPUT ${PACKAGE_OUTPUT}
+        COMMAND ${CMAKE_COMMAND} -E true
+        DEPENDS ${TARGET_NAME}_ext
+    )
```
swatanabe, line 343: This is weird. The reason that it was failing without this is that the other `add_custom_command` only has a file level dependency on `${PACKAGE_OUTPUT}` but not a target level dependency on `${TARGET_NAME}_ext`.

James-Mart: Fixed

swatanabe (approved):

PR #861, opened 2024-09-24
Union cargo package dependencies

James-Mart:
This allows package dependencies to be specified in any crate in a cargo-package instead of only the service crate.
Dependencies put in multiple crates will be unioned for inclusion in the final package.

This PR also moves the dependencies for the ChainMail package from the service crate to the root crate. It does not demonstrate unioning capabilities which were manually tested.

-----
Merge to `main` once #859 is merged.


rust/cargo-psibase/src/package.rs
```diff
@@ -205,7 +205,7 @@ pub async fn build_package(
     metadata: &MetadataIndex<'_>,
     service: Option<&str>,
 ) -> Result<PackageInfo, anyhow::Error> {
-    let mut depends = vec![];
+    let mut depends_set = HashSet::new();
```
swatanabe, line 208: The elements are already being sorted. Adding a `dedup` at that point is probably more efficient.

swatanabe (approved):

PR #866, opened 2024-09-25
Chainmail: patching line of debugging code

cool-ant:
I mistakenly included a debugging line I meant to revert before the PR. It got missed and committed by mistake. This just reverses that line.

### DIFF
--- a/services/user/Branding/ui/src/App.tsx
+++ b/services/user/Branding/ui/src/App.tsx
@@ -19,7 +19,6 @@ export const App = () => {
 
     const init = async () => {
         await supervisor.onLoaded();
-        supervisor.preLoadPlugins([{ service: "branding" }]);
         await getNetworkName();
     };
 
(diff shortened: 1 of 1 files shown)


services/user/Branding/ui/src/App.tsx
```diff
@@ -20,7 +20,7 @@ export const App = () => {
     const init = async () => {
         await supervisor.onLoaded();
         supervisor.preLoadPlugins([{ service: "branding" }]);
-        setTimeout(getNetworkName, 1500);
+        await getNetworkName();
```
James-Mart, line 23: If I understood your other comment correctly, `await` here was not working at some point. 
But I assume from this change that it is now working, yeah?

cool-ant: yes

James-Mart (approved):

PR #868, opened 2024-09-26
Add plugin data persistence doc

### LINKED ISSUE #740: Write plugin data persistence specification
# Task

Write a specification and add it to the mdbook docs for psibase

# Specification

## Access to database operations 

To enable database actions within plugins, a plugin must import database read/write functionality. 

Read/write functionality is provided by a system plugin in the `clientdata` namespace. That plugin provides a high level key/value API for plugins to use that is specific to psibase plugins.

As an MVP implementation, the API will simply provide a get/set implementation such as:
```
    set      : func(key : string, value : string) -> result<_, error>;
    get      : func(key : string) -> result<string, error>;
```
### Design implications

There is actually no standardized wasi-keyvalue interface (The [proposal](https://github.com/WebAssembly/wasi-keyvalue) is still in stage 2 at the time of writing), so any interface we enable today will be subject to future change. This design specifies that plugins should import a custom psibase-plugin-specific API that itself leverages the `wasi-keyvalue` standard. 

## Host provided functionality

The `clientdata` plugin must itself import functionality that allows it to interact with some persistent storage layer. 

This host functionality is a browser shim implementation of the `wasi-keyvalue` API. Technically, any plugin can import this API and use it directly. But using the standard clientdata plugin simplifies the integration (uses standard error types, exposes a more convenient interface).

The `wasi-keyvalue` implementation currently uses LocalStorage as its data backing, because it is a synchronous interface which is easiest to integrate with wasm.

## Persistence

When private user data is written, it is immediately persisted. 

## Concurrency

All reads/writes are currently synchronous, with no concurrency.

## Synchronization across devices

Plugins all run within the domain of the supervisor, which ensures that all plugin data is written-to/read-from a storage backing tied to the supervisor domain. Cloning storage in the supervisor domain between TLDs is a way to synchronize plugin data across devices.

James-Mart:
Closes #740


doc/psidk/src/specifications/app-architecture/plugins.md
```diff
@@ -82,30 +82,64 @@ Supervisor->>psinode: Submit transaction
...
+### Access to database operations
+
+The psibase plugin host provides a `wasi:keyvalue` interface for plugins to use. However, the wasi-keyvalue interface [proposal](https://github.com/WebAssembly/wasi-keyvalue) is still in stage 2 of standardization, and therefore the implementation provided by the host is subject to change.
+
+For more stable interactions with the database, a plugin should import the `clientdata` plugin, which is a wrapper around the host's `wasi:keyvalue` interface.
+
+This `clientdata` plugin provides a high level key/value API for plugins.
+
+```
+    get: func(key: string) -> option<list<u8>>;
```
swatanabe, line 124: This description doesn't make it clear that each caller gets an independent key space.

doc/psidk/src/specifications/app-architecture/plugins.md
```diff
@@ -82,30 +82,64 @@ Supervisor->>psinode: Submit transaction
...
+    set: func(key: string, value: list<u8>) -> result<_, error>;
+    delete: func(key: string);
+```
+
+### Data backing
+
+The `wasi-keyvalue` implementation currently uses LocalStorage as its data backing, because it is a synchronous interface which is easiest to integrate with wasm.
+
+Note: This imposes a major restriction on the amount of memory available for psibase plugins: Currently there is a hard limit of 5MB for the total amount of data stored in LocalStorage across all psibase apps. This restriction will be lifted in a future implementation of plugin storage that uses a different data backing layer such as IndexedDB.
```
swatanabe, line 133: The limit is browser specific.

doc/psidk/src/specifications/app-architecture/plugins.md
James-Mart: Clarified

doc/psidk/src/specifications/app-architecture/plugins.md
James-Mart: Removed the number and just linked to a web page about the limit

swatanabe (approved):

PR #867, opened 2024-09-26
Use sites for storesys

James-Mart:
# Description

This PR makes the Sites app the default server for storing/serving UI artifacts, instead of requiring the same `storeSys` boilerplate to be used within every app.

Accordingly, both the package infra and psibase CLI tool were updated to call `storeSys` on `sites`.

Furthermore, I needed to add a Sites plugin to accommodate the Branding app, which was previously calling its own `storeSys` action manually.

I removed the `storageInterface`, `WebContentRow`, and `serveContent` as interfaces, since they are no longer needed.

# Docs changes

I also updated StorageInterface docs, CLI docs, and package support docs.

# Benefits

* Simpler to write services: 99% of cases it is unnecessary to add the store/serveContent boilerplate to a service.
* RPC services can now be used exclusively for servicing queries, which makes them read-only and easier to reason about
* Makes upgrades to UI storage/service easier as it is now centralized. For example it should now be fairly trivial to add gzip compression on all artifacts that use the Sites app.

PR #871, opened 2024-09-30
Site content http-server improvements

James-Mart:
# Background

Since #867, all web content is stored in and served from the same http server, implemented in the `Sites` service. Since all content now shares the Sites service http server, improvements made to clients that upload files can make more assumptions about the implementation of the server (since they all use the Sites implementation). This makes it possible to implement client-side file compression.

Furthermore, improvements made to the Sites server have greater impact, since the serving of files is centralized and improvements are now shared across all served content.

# Scope

This PR does three things:
1. Introduce a baseline for the the CSP header on all content (and allow csp customization)
2. Add the necessary headers for browsers to do better http caching
3. Adds client-side brotli compression in the psibase CLI tool and `package install` infra

## CSP header baseline

Content-Security-Policy is a good header to set for security purposes, as modern browsers restrict the capabilities of the document on the client-side according to the policy expressed in this header. The intention of this PR is not to minimize the surface area for attack in our CSP, but instead just to introduce a baseline that simply allows the capabilities that we are already using. Perhaps a future task will be to try to minimize the required capabilities of our infrastructure so we can tighten the default CSP.

Default policy:
```
          "default-src 'self';"
          "font-src 'self' https:;"
          "script-src 'self' 'unsafe-eval' 'unsafe-inline' blob: https:;"
          "img-src *;"
          "style-src 'self' 'unsafe-inline';"
          "frame-src *;"
          "connect-src * blob:;"
```
Explanation:
* script-src:
  * `unsafe-eval` is needed for WebAssembly.compile()
  * `unsafe-inline` is needed for inline <script> tags
  * `blob:` for loading dynamically generated modules
  * `https:` for dynamically loading libs from CDNs
* font-src:
  * `https:` for dynamically loading fonts from CDNs
* frame-src:
  * `*` is needed to allow embedding supervisor
* connect-src:
  * `blob:` for fetch-compiling blob urls
  * `*` allows fetching plugins, as well as websocket connections

### Dynamic CSP

The `setCsp(std::string path, std::string csp)` action can be used to customize the CSP headers for a particular piece of content, or `setCsp("*", ...)` can be used to set the CSP header for an entire site. The priority of the CSP header applied to content is given to a header set for a specific path, then the header set for an entire site, and finally will default to the default policy if no custom headers were provided.

## Http caching

On upload, the server now calculates and stores a seahash of the content at each path, adding 64 bits to the size of each record. The server now returns the hash in an [`etag`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/ETag) header, as well as setting the `cache-control` header to `no-cache`. 

This cache policy tells the browser to add, on any subsequent requests for the same content, an `If-None-Match` header with the previously sent ETag value. The server then has a chance to respond with `304 Unmodified`, allowing the browser to use its cached value for the asset, rather than initiating a fresh transfer. 

## Compression

The `psibase upload` CLI, and the package install infrastructure all use [Brotli](https://en.wikipedia.org/wiki/Brotli) compression with a compression quality of 4 on web artifacts of the following mime types:
* "text/plain"
* "text/html"
* "text/css"
* "application/javascript"
* "application/json"
* "application/xml"
* "application/rss+xml"
* "application/atom+xml"
* "image/svg+xml"
* "font/ttf"
* "font/otf"
* "application/wasm"

These file types have reasonably good compression results. 

`Brotli` is a newer algo than gzip, and ostensibly tends to have better compression ratios, and it's supported by all major browsers. 

<details>
<summary>Compression details on boot</summary>

| **File**                                                                             | **Type**                      | **Uncompressed size** | **Compressed size** | **Compression ratio** |
|--------------------------------------------------------------------------------------|-------------------------------|-----------------------|---------------------|-----------------------|
| data/sites/plugin.wasm                                                               | application/wasm              |             1,021,028 |             501,059 |                  2.04 |
| data/accounts/plugin.wasm                                                            | application/wasm              |               422,898 |             191,960 |                   2.2 |
| data/common-api/common/react-dom.development.js                                      | application/javascript        |             1,080,227 |             226,240 |                  4.77 |
| data/common-api/common/react.production.min.js                                       | application/javascript        |                10,751 |               4,395 |                  2.45 |
| data/common-api/common/useGraphQLQuery.mjs                                           | application/javascript        |                 3,104 |                 931 |                  3.33 |
| data/common-api/common/fonts/raleway.css                                             | text/css                      |                   345 |                 138 |                   2.5 |
| data/common-api/common/fonts/red-hat-mono.css                                        | text/css                      |                   170 |                 129 |                  1.32 |
| data/common-api/common/fonts/red-hat-mono-variable-normal.ttf                        | font/ttf                      |                66,064 |              35,220 |                  1.88 |
| data/common-api/common/fonts/raleway-variable-italic.ttf                             | font/ttf                      |               303,956 |             163,313 |                  1.86 |
| data/common-api/common/fonts/raleway-variable-normal.ttf                             | font/ttf                      |               309,720 |             162,221 |                  1.91 |
| data/common-api/common/useLocalStorageState.js                                       | application/javascript        |                 6,273 |               1,837 |                  3.41 |
| data/common-api/common/htm.module.js                                                 | application/javascript        |                 1,207 |                 645 |                  1.87 |
| data/common-api/common/useLocalStorage.mjs                                           | application/javascript        |                   902 |                 335 |                  2.69 |
| data/common-api/common/react-router-dom.min.js                                       | application/javascript        |                30,218 |              10,192 |                  2.96 |
| data/common-api/common/common-lib.js                                                 | application/javascript        |               249,894 |              61,061 |                  4.09 |
| data/common-api/common/react-dom.production.min.js                                   | application/javascript        |               131,835 |              43,519 |                  3.03 |
| data/common-api/common/semantic-ui-react.min.js                                      | application/javascript        |               269,974 |              74,166 |                  3.64 |
| data/common-api/common/react.development.js                                          | application/javascript        |               109,931 |              28,916 |                   3.8 |
| data/common-api/common/SimpleUI.mjs                                                  | application/javascript        |                 3,874 |               1,249 |                   3.1 |
| data/common-api/common/component_parser.wasm                                         | application/wasm              |             1,627,670 |             453,132 |                  3.59 |
| data/transact/plugin.wasm                                                            | application/wasm              |             1,150,034 |             356,957 |                  3.22 |
| data/supervisor/vite.svg                                                             | image/svg+xml                 |                 1,497 |                 791 |                  1.89 |
| data/supervisor/index.html                                                           | text/html                     |                   386 |                 219 |                  1.76 |
| data/supervisor/assets/js-component-bindgen-component.core2-7090fc5c.wasm            | application/wasm              |                16,100 |               7,162 |                  2.25 |
| data/supervisor/assets/js-component-bindgen-component.core-bbbfca46.wasm             | application/wasm              |             2,234,350 |             738,197 |                  3.03 |
| data/supervisor/assets/bindings_wasm_bg-ef168d8b.wasm                                | application/wasm              |               940,258 |             326,476 |                  2.88 |
| data/supervisor/assets/wasip2-shim-2e677a7c.js                                       | application/javascript        |                16,661 |               4,463 |                  3.73 |
| data/supervisor/assets/index-eb293a36.js                                             | application/javascript        |               737,905 |             180,866 |                  4.08 |
| data/supervisor/assets/wasi-keyvalue-8bd6d50b.js                                     | application/javascript        |                 5,982 |               1,767 |                  3.39 |
| data/chainmail/plugin.wasm                                                           | application/wasm              |                61,852 |              24,720 |                   2.5 |
| data/chainmail/index.js                                                              | application/javascript        |             1,297,069 |             407,440 |                  3.18 |
| data/chainmail/style.css                                                             | text/css                      |                48,962 |               8,928 |                  5.48 |
| data/chainmail/index.html                                                            | text/html                     |                   363 |                 185 |                  1.96 |
| data/homepage/index.js                                                               | application/javascript        |               483,141 |             156,578 |                  3.09 |
| data/homepage/style.css                                                              | text/css                      |                28,440 |               5,764 |                  4.93 |
| data/homepage/index.html                                                             | text/html                     |                   458 |                 241 |                   1.9 |
| data/tokens/vite.svg                                                                 | image/svg+xml                 |                 1,497 |                 791 |                  1.89 |
| data/tokens/index.html                                                               | text/html                     |                   454 |                 243 |                  1.87 |
| data/tokens/plugin.wasm                                                              | application/wasm              |               218,666 |              83,230 |                  2.63 |
| data/tokens/assets/index-Crx6mjGm.css                                                | text/css                      |                27,147 |               5,522 |                  4.92 |
| data/tokens/assets/index-B51xWaai.js                                                 | application/javascript        |               762,609 |             248,913 |                  3.06 |
| data/auth-sig/plugin.wasm                                                            | application/wasm              |               238,929 |              80,459 |                  2.97 |
| data/invite/plugin.wasm                                                              | application/wasm              |               165,763 |              70,272 |                  2.36 |
| data/explorer/_app/immutable/nodes/2.afab9e24.js                                     | application/javascript        |                 9,005 |               3,224 |                  2.79 |
| data/explorer/_app/immutable/nodes/4.ee76ecd9.js                                     | application/javascript        |                 5,771 |               2,507 |                   2.3 |
| data/explorer/_app/immutable/nodes/5.5c85d5e3.js                                     | application/javascript        |                 8,320 |               3,513 |                  2.37 |
| data/explorer/_app/immutable/nodes/3.dcec6e10.js                                     | application/javascript        |                 6,730 |               2,817 |                  2.39 |
| data/explorer/_app/immutable/nodes/1.2f1a6805.js                                     | application/javascript        |                   800 |                 523 |                  1.53 |
| data/explorer/_app/immutable/nodes/0.2203a769.js                                     | application/javascript        |                   863 |                 570 |                  1.51 |
| data/explorer/_app/immutable/entry/start.b624683f.js                                 | application/javascript        |                25,306 |              10,270 |                  2.46 |
| data/explorer/_app/immutable/entry/app.740f5aa9.js                                   | application/javascript        |                 7,067 |               2,550 |                  2.77 |
| data/explorer/_app/immutable/chunks/stores.31f28e4e.js                               | application/javascript        |                   238 |                 147 |                  1.62 |
| data/explorer/_app/immutable/chunks/Loader.svelte_svelte_type_style_lang.d83c45d4.js | application/javascript        |                 6,867 |               2,907 |                  2.36 |
| data/explorer/_app/immutable/chunks/Loader.8d0e4ad7.js                               | application/javascript        |                 1,944 |               1,068 |                  1.82 |
| data/explorer/_app/immutable/chunks/index.87fc741b.js                                | application/javascript        |                 8,429 |               3,442 |                  2.45 |
| data/explorer/_app/immutable/chunks/leftArrow.608e4bbe.js                            | application/javascript        |                 1,088 |                 657 |                  1.66 |
| data/explorer/_app/immutable/chunks/singletons.ed9234af.js                           | application/javascript        |                 2,904 |               1,487 |                  1.95 |
| data/explorer/_app/immutable/chunks/app-explorer-icon.9a1b7f40.js                    | application/javascript        |                 2,278 |               1,148 |                  1.98 |
| data/explorer/_app/immutable/chunks/Blocks.7b28f86c.js                               | application/javascript        |                 4,010 |               1,768 |                  2.27 |
| data/explorer/_app/immutable/assets/Loader.5ebdc467.css                              | text/css                      |                   990 |                 375 |                  2.64 |
| data/explorer/_app/immutable/assets/loader.dbd3a0f9.svg                              | image/svg+xml                 |                 1,401 |                 733 |                  1.91 |
| data/explorer/_app/immutable/assets/0.8fdf211d.css                                   | text/css                      |                16,940 |               4,055 |                  4.18 |
| data/explorer/_app/immutable/assets/search.e531d965.svg                              | image/svg+xml                 |                   439 |                 252 |                  1.74 |
| data/explorer/_app/immutable/assets/_layout.8fdf211d.css                             | text/css                      |                16,940 |               4,055 |                  4.18 |
| data/explorer/_app/immutable/assets/leftArrow.db2e4fcf.svg                           | image/svg+xml                 |                   304 |                 200 |                  1.52 |
| data/explorer/_app/immutable/assets/Loader.33e6d9c4.css                              | text/css                      |                   968 |                 361 |                  2.68 |
| data/explorer/_app/immutable/assets/incoming.12b7b707.svg                            | image/svg+xml                 |                   881 |                 426 |                  2.07 |
| data/explorer/_app/immutable/assets/dot.1cb740d7.svg                                 | image/svg+xml                 |                   165 |                 132 |                  1.25 |
| data/explorer/_app/immutable/assets/outgoing.fe310ac0.svg                            | image/svg+xml                 |                   885 |                 425 |                  2.08 |
| data/explorer/_app/immutable/assets/app-explorer-icon.5aa51ee9.svg                   | image/svg+xml                 |                 2,270 |                 947 |                   2.4 |
| data/explorer/_app/version.json                                                      | application/json              |                    27 |                  29 |                  0.93 |
| data/explorer/index.html                                                             | text/html                     |                 1,118 |                 462 |                  2.42 |
| data/clientdata/plugin.wasm                                                          | application/wasm              |                72,293 |              27,822 |                   2.6 |
| data/docs/mark.min.js                                                                | application/javascript        |                17,320 |               5,964 |                   2.9 |
| data/docs/searchindex.json                                                           | application/json              |             2,314,197 |             279,639 |                  8.28 |
| data/docs/book.js                                                                    | application/javascript        |                25,437 |               5,881 |                  4.33 |
| data/docs/default-apps/docs.html                                                     | text/html                     |                24,601 |               4,519 |                  5.44 |
| data/docs/default-apps/index.html                                                    | text/html                     |                25,236 |               4,559 |                  5.54 |
| data/docs/default-apps/setcode.html                                                  | text/html                     |                26,634 |               4,804 |                  5.54 |
| data/docs/default-apps/invite.html                                                   | text/html                     |                40,777 |               6,751 |                  6.04 |
| data/docs/default-apps/x-admin.html                                                  | text/html                     |                31,978 |               6,985 |                  4.58 |
| data/docs/default-apps/http-server.html                                              | text/html                     |                29,049 |               5,338 |                  5.44 |
| data/docs/default-apps/events.html                                                   | text/html                     |                29,410 |               5,499 |                  5.35 |
| data/docs/default-apps/transact.html                                                 | text/html                     |                41,196 |               7,454 |                  5.53 |
| data/docs/default-apps/auth-sig.html                                                 | text/html                     |                30,037 |               5,352 |                  5.61 |
| data/docs/default-apps/accounts.html                                                 | text/html                     |                31,834 |               5,429 |                  5.86 |
| data/docs/default-apps/sites.html                                                    | text/html                     |                29,782 |               5,374 |                  5.54 |
| data/docs/default-apps/common-api.html                                               | text/html                     |                25,591 |               4,672 |                  5.48 |
| data/docs/index.html                                                                 | text/html                     |                24,810 |               4,832 |                  5.13 |
| data/docs/elasticlunr.min.js                                                         | application/javascript        |                18,051 |               5,783 |                  3.12 |
| data/docs/ayu-highlight.css                                                          | text/css                      |                   929 |                 397 |                  2.34 |
| data/docs/highlight.css                                                              | text/css                      |                 1,197 |                 538 |                  2.22 |
| data/docs/doc-root.cpp                                                               | text/plain                    |                   455 |                 164 |                  2.77 |
| data/docs/highlight.js                                                               | application/javascript        |               137,537 |              46,478 |                  2.96 |
| data/docs/run-infrastructure/index.html                                              | text/html                     |                25,653 |               4,816 |                  5.33 |
| data/docs/run-infrastructure/cli/index.html                                          | text/html                     |                28,887 |               5,890 |                   4.9 |
| data/docs/run-infrastructure/cli/psibase.html                                        | text/html                     |                37,093 |               6,858 |                  5.41 |
| data/docs/run-infrastructure/cli/psinode.html                                        | text/html                     |                31,867 |               6,681 |                  4.77 |
| data/docs/run-infrastructure/configuration/logging.html                              | text/html                     |                40,306 |               8,668 |                  4.65 |
| data/docs/run-infrastructure/configuration/https.html                                | text/html                     |                27,098 |               5,270 |                  5.14 |
| data/docs/run-infrastructure/administration.html                                     | text/html                     |                54,453 |              11,598 |                   4.7 |
| data/docs/tomorrow-night.css                                                         | text/css                      |                 1,667 |                 613 |                  2.72 |
| data/docs/fonts/open-sans-v17-all-charsets-300italic.woff2                           | font/woff2                    |                40,656 |              40,660 |                     1 |
| data/docs/fonts/open-sans-v17-all-charsets-700italic.woff2                           | font/woff2                    |                40,800 |              40,804 |                     1 |
| data/docs/fonts/open-sans-v17-all-charsets-600italic.woff2                           | font/woff2                    |                42,120 |              42,124 |                     1 |
| data/docs/fonts/open-sans-v17-all-charsets-300.woff2                                 | font/woff2                    |                44,352 |              44,356 |                     1 |
| data/docs/fonts/open-sans-v17-all-charsets-regular.woff2                             | font/woff2                    |                43,236 |              43,240 |                     1 |
| data/docs/fonts/SOURCE-CODE-PRO-LICENSE.txt                                          | text/plain                    |                 4,528 |               1,905 |                  2.38 |
| data/docs/fonts/open-sans-v17-all-charsets-800.woff2                                 | font/woff2                    |                44,536 |              44,540 |                     1 |
| data/docs/fonts/open-sans-v17-all-charsets-italic.woff2                              | font/woff2                    |                41,076 |              41,080 |                     1 |
| data/docs/fonts/OPEN-SANS-LICENSE.txt                                                | text/plain                    |                11,358 |               3,891 |                  2.92 |
| data/docs/fonts/fonts.css                                                            | text/css                      |                 3,620 |                 549 |                  6.59 |
| data/docs/fonts/open-sans-v17-all-charsets-600.woff2                                 | font/woff2                    |                44,936 |              44,940 |                     1 |
| data/docs/fonts/open-sans-v17-all-charsets-800italic.woff2                           | font/woff2                    |                40,812 |              40,816 |                     1 |
| data/docs/fonts/open-sans-v17-all-charsets-700.woff2                                 | font/woff2                    |                44,988 |              44,992 |                     1 |
| data/docs/fonts/source-code-pro-v11-all-charsets-500.woff2                           | font/woff2                    |                59,140 |              59,144 |                     1 |
| data/docs/clipboard.min.js                                                           | application/javascript        |                10,754 |               3,416 |                  3.15 |
| data/docs/contribute/index.html                                                      | text/html                     |                23,320 |               4,166 |                   5.6 |
| data/docs/FontAwesome/fonts/fontawesome-webfont.woff2                                | font/woff2                    |                77,160 |              77,165 |                     1 |
| data/docs/FontAwesome/fonts/FontAwesome.ttf                                          | font/ttf                      |               165,548 |              98,507 |                  1.68 |
| data/docs/FontAwesome/fonts/fontawesome-webfont.eot                                  | application/vnd.ms-fontobject |               165,742 |              98,488 |                  1.68 |
| data/docs/FontAwesome/fonts/fontawesome-webfont.ttf                                  | font/ttf                      |               165,548 |              98,507 |                  1.68 |
| data/docs/FontAwesome/fonts/fontawesome-webfont.svg                                  | image/svg+xml                 |               444,379 |             132,894 |                  3.34 |
| data/docs/FontAwesome/css/font-awesome.css                                           | text/css                      |                31,000 |               6,988 |                  4.44 |
| data/docs/css/chrome.css                                                             | text/css                      |                10,436 |               2,799 |                  3.73 |
| data/docs/css/general.css                                                            | text/css                      |                 4,303 |               1,570 |                  2.74 |
| data/docs/css/variables.css                                                          | text/css                      |                 5,989 |               1,284 |                  4.66 |
| data/docs/searcher.js                                                                | application/javascript        |                18,547 |               5,277 |                  3.51 |
| data/docs/theme/js/pagetoc.js                                                        | application/javascript        |                 1,673 |                 494 |                  3.39 |
| data/docs/theme/js/mermaid-load.js                                                   | application/javascript        |                   286 |                 166 |                  1.72 |
| data/docs/theme/js/fix-github-edit-link.js                                           | application/javascript        |                   377 |                 179 |                  2.11 |
| data/docs/theme/js/app-init.js                                                       | application/javascript        |                   235 |                 169 |                  1.39 |
| data/docs/theme/css/pagetoc-tock.css                                                 | text/css                      |                   191 |                 105 |                  1.82 |
| data/docs/theme/css/pagetoc.css                                                      | text/css                      |                 1,403 |                 431 |                  3.26 |
| data/docs/development/index.html                                                     | text/html                     |                24,286 |               4,391 |                  5.53 |
| data/docs/development/front-ends/index.html                                          | text/html                     |                26,104 |               4,897 |                  5.33 |
| data/docs/development/front-ends/reference/http-requests.html                        | text/html                     |                36,738 |               7,738 |                  4.75 |
| data/docs/development/front-ends/reference/external.html                             | text/html                     |                26,885 |               5,172 |                   5.2 |
| data/docs/development/front-ends/reference/js-libraries.html                         | text/html                     |                36,635 |               7,180 |                   5.1 |
| data/docs/development/services/index.html                                            | text/html                     |                25,310 |               4,661 |                  5.43 |
| data/docs/development/services/webassembly.html                                      | text/html                     |                27,147 |               5,221 |                   5.2 |
| data/docs/development/services/cpp-service/minimal-ui/index.html                     | text/html                     |                29,307 |               5,673 |                  5.17 |
| data/docs/development/services/cpp-service/minimal-ui/CMakeLists.txt                 | text/plain                    |                   533 |                 329 |                  1.62 |
| data/docs/development/services/cpp-service/minimal-ui/example.cpp                    | text/plain                    |                   657 |                 323 |                  2.03 |
| data/docs/development/services/cpp-service/index.html                                | text/html                     |                24,559 |               4,226 |                  5.81 |
| data/docs/development/services/cpp-service/reference/table.html                      | text/html                     |                58,479 |               9,391 |                  6.23 |
| data/docs/development/services/cpp-service/reference/services-events.html            | text/html                     |                60,615 |               9,529 |                  6.36 |
| data/docs/development/services/cpp-service/reference/magic-numbers.html              | text/html                     |                35,371 |               5,715 |                  6.19 |
| data/docs/development/services/cpp-service/reference/cmake.html                      | text/html                     |                30,101 |               5,830 |                  5.16 |
| data/docs/development/services/cpp-service/reference/native-functions.html           | text/html                     |                81,658 |              11,109 |                  7.35 |
| data/docs/development/services/cpp-service/reference/system.html                     | text/html                     |                26,259 |               4,733 |                  5.55 |
| data/docs/development/services/cpp-service/reference/web-services.html               | text/html                     |                74,121 |              13,013 |                   5.7 |
| data/docs/development/services/cpp-service/CMakeLists.txt                            | text/plain                    |                   430 |                 235 |                  1.83 |
| data/docs/development/services/cpp-service/basic/index.html                          | text/html                     |                28,223 |               5,574 |                  5.06 |
| data/docs/development/services/cpp-service/basic/CMakeLists.txt                      | text/plain                    |                   511 |                 313 |                  1.63 |
| data/docs/development/services/cpp-service/basic/example.cpp                         | text/plain                    |                   595 |                 292 |                  2.04 |
| data/docs/development/services/cpp-service/calling/arithmetic.hpp                    | text/plain                    |                   649 |                 342 |                   1.9 |
| data/docs/development/services/cpp-service/calling/index.html                        | text/html                     |                31,292 |               6,149 |                  5.09 |
| data/docs/development/services/cpp-service/calling/caller.hpp                        | text/plain                    |                   360 |                 223 |                  1.61 |
| data/docs/development/services/cpp-service/calling/CMakeLists.txt                    | text/plain                    |                   682 |                 343 |                  1.99 |
| data/docs/development/services/cpp-service/calling/arithmetic.cpp                    | text/plain                    |                   448 |                 226 |                  1.98 |
| data/docs/development/services/cpp-service/calling/caller.cpp                        | text/plain                    |                   682 |                 332 |                  2.05 |
| data/docs/development/services/cpp-service/setup.html                                | text/html                     |                26,647 |               4,981 |                  5.35 |
| data/docs/development/services/rust-service/index.html                               | text/html                     |                24,805 |               4,240 |                  5.85 |
| data/docs/development/services/rust-service/testing.html                             | text/html                     |                27,625 |               5,439 |                  5.08 |
| data/docs/development/services/rust-service/graphql.html                             | text/html                     |                37,451 |               7,807 |                   4.8 |
| data/docs/development/services/rust-service/reference/web-services.html              | text/html                     |                29,451 |               5,639 |                  5.22 |
| data/docs/development/services/rust-service/tables.html                              | text/html                     |                42,728 |               8,936 |                  4.78 |
| data/docs/development/services/rust-service/minimal-ui.html                          | text/html                     |                28,670 |               5,594 |                  5.13 |
| data/docs/development/services/rust-service/package.html                             | text/html                     |                26,921 |               5,090 |                  5.29 |
| data/docs/development/services/rust-service/basic/index.html                         | text/html                     |                28,473 |               5,555 |                  5.13 |
| data/docs/development/services/rust-service/setup.html                               | text/html                     |                25,656 |               4,636 |                  5.53 |
| data/docs/development/services/rust-service/calling.html                             | text/html                     |                31,394 |               6,665 |                  4.71 |
| data/docs/development/services/standards.html                                        | text/html                     |                25,058 |               4,593 |                  5.46 |
| data/docs/development/services/action-scripts.html                                   | text/html                     |                32,133 |               6,946 |                  4.63 |
| data/docs/searchindex.js                                                             | application/javascript        |             2,314,228 |             280,841 |                  8.24 |
| data/docs/specifications/index.html                                                  | text/html                     |                41,698 |               6,902 |                  6.04 |
| data/docs/specifications/app-architecture/index.html                                 | text/html                     |                40,954 |               7,174 |                  5.71 |
| data/docs/specifications/app-architecture/supervisor.html                            | text/html                     |                36,319 |               7,224 |                  5.03 |
| data/docs/specifications/app-architecture/plugins.html                               | text/html                     |                33,623 |               7,381 |                  4.56 |
| data/docs/specifications/app-architecture/events.html                                | text/html                     |                30,883 |               6,575 |                   4.7 |
| data/docs/specifications/app-architecture/services.html                              | text/html                     |                27,736 |               5,528 |                  5.02 |
| data/docs/specifications/concepts/user-onboarding.html                               | text/html                     |                31,321 |               6,513 |                  4.81 |
| data/docs/specifications/data-formats/json.html                                      | text/html                     |                31,585 |               6,721 |                   4.7 |
| data/docs/specifications/data-formats/schema.html                                    | text/html                     |                38,457 |               7,561 |                  5.09 |
| data/docs/specifications/data-formats/package.html                                   | text/html                     |                27,933 |               5,412 |                  5.16 |
| data/docs/specifications/data-formats/account-numbers.html                           | text/html                     |                26,335 |               5,062 |                   5.2 |
| data/docs/specifications/data-formats/fracpack.html                                  | text/html                     |                38,443 |               8,319 |                  4.62 |
| data/docs/specifications/blockchain/index.html                                       | text/html                     |                29,056 |               6,049 |                   4.8 |
| data/docs/specifications/blockchain/validation.html                                  | text/html                     |                24,998 |               4,565 |                  5.48 |
| data/docs/specifications/blockchain/database.html                                    | text/html                     |                25,049 |               4,594 |                  5.45 |
| data/docs/specifications/blockchain/tapos.html                                       | text/html                     |                28,712 |               5,854 |                   4.9 |
| data/docs/specifications/blockchain/smart-authorization.html                         | text/html                     |                30,692 |               6,304 |                  4.87 |
| data/docs/specifications/blockchain/entry-points.html                                | text/html                     |                28,602 |               5,442 |                  5.26 |
| data/docs/specifications/blockchain/wasm-runtime.html                                | text/html                     |                25,994 |               4,810 |                   5.4 |
| data/docs/specifications/blockchain/peer-consensus/index.html                        | text/html                     |                26,492 |               4,973 |                  5.33 |
| data/docs/specifications/blockchain/peer-consensus/bft.html                          | text/html                     |                29,786 |               5,967 |                  4.99 |
| data/docs/specifications/blockchain/peer-consensus/joint-consensus.html              | text/html                     |                26,239 |               4,689 |                   5.6 |
| data/docs/specifications/blockchain/peer-consensus/cft.html                          | text/html                     |                26,528 |               4,971 |                  5.34 |
| data/docs/_img/logo-psibase-green.svg                                                | image/svg+xml                 |                   844 |                 471 |                  1.79 |
| data/docs/favicon.svg                                                                | image/svg+xml                 |                 1,835 |               1,012 |                  1.81 |
| data/docs/404.html                                                                   | text/html                     |                22,130 |               3,984 |                  5.55 |
|                                                                                      |                               |                       |                     |                       |
| **Total**                                                                            |                               |        **23,204,057** |       **6,866,345** |       **3.379389908** |

</details>

# Example statistics

At the time of writing, the Tokens app would ordinarily have 6.37MB of content stored on the server that the browser would need to load. Because of compression, that shrinks to 1.97MB, a 75% space and bandwidth savings.
image

Because of caching, on subsequent reloads of the page, only 437 bytes are transferred if none of the content changes on the server:
image


rust/psibase/src/main.rs
```diff
@@ -87,6 +87,11 @@ enum Command {
         package_source: Vec<String>,
 
         services: Vec<OsString>,
+
+        /// Configure compression level for boot package file uploads
+        /// (1=fastest, 11=most compression)
+        #[clap(short = 'z', long, value_name = "LEVEL", default_value = "4")]
```
swatanabe, line 93: I think the syntax should be a number of z's equal to the compression level. i.e. `-zzz`, or `-zzzzzzzzzzz` for the most compression.

programs/psinode/tests/test_psibase.py
```diff
@@ -96,8 +96,8 @@ def test_upgrade(self, cluster):
         a = cluster.complete(*testutil.generate_names(1))[0]
         a.boot(packages=['Minimal', 'Explorer', 'Sites'])
 
-        foo10 = TestPackage('foo', '1.0.0').depends('Sites').service('foo', data={'file1.txt': 'original', 'file2.txt': 'deleted'})
-        foo11 = TestPackage('foo', '1.1.0').depends('Sites').service('foo', data={'file1.txt': 'updated', 'file3.txt': 'added'})
+        foo10 = TestPackage('foo', '1.0.0').depends('Sites').service('foo', data={'file1.png': 'original', 'file2.png': 'deleted'})
+        foo11 = TestPackage('foo', '1.1.0').depends('Sites').service('foo', data={'file1.png': 'updated', 'file3.png': 'added'})
```
swatanabe, line 100: It would probably be better to ensure that it's decoded. I don't think these are valid PNGs:

"The gzip and deflate transfer-encodings are automatically decoded for you.

The br transfer-encoding is automatically decoded for you if a Brotli library like [brotli](https://pypi.org/project/brotli) or [brotlicffi](https://pypi.org/project/brotlicffi) is installed." (from the [requests docs](https://docs.python-requests.org/en/latest/user/quickstart/#binary-response-content))

rust/psibase/src/main.rs
```diff
@@ -268,11 +283,29 @@ fn to_hex(bytes: &[u8]) -> String {
     String::from_utf8(result).unwrap()
 }
 
-fn store_sys(sender: AccountNumber, path: &str, content_type: &str, content: &[u8]) -> Action {
+fn validate_compression(compression_level: u32) -> Result<(), anyhow::Error> {
```
swatanabe, line 286: clap can do this kind of validation: https://docs.rs/clap/latest/clap/struct.Arg.html#method.value_parser

programs/psinode/tests/test_psibase.py
swatanabe: We're actually cheating a bit here. The server is supposed to check `Accept-Encoding` before sending a compressed response.

James-Mart: Haha yeah this was definitely just a hack to trick the tooling into not compressing the file. 

Re: `Accept-Encoding`, looks like what we are doing is technically not cheating:
```
A request without an Accept-Encoding header field implies that the
   user agent has no preferences regarding content-codings.  Although
   this allows the server to use any content-coding in a response, it
   does not imply that the user agent will be able to correctly process
   all encodings.
```
[RFC 7231](https://datatracker.ietf.org/doc/html/rfc7231#section-5.3.4)

Although I think the implementation is incomplete, `Sites` should verify that the requests either has no `Accept-Encoding`, or that it explicitly specifies that `br` is accepted. In case the request has `Accept-Encoding` with a value that is not `br`, I think the server should return `406 Not Acceptable`. Agree?

swatanabe: Python *does* send `Accept-Encoding': gzip, deflate`.

James-Mart: > > If a non-empty Accept-Encoding header field is present in a request and none of the available representations for the response have a content coding that is listed as acceptable, the origin server SHOULD send a response without any content coding unless the identity coding is indicated as unacceptable.

For Python I will install the brotli lib so it adds it to the Accept-Encoding. But in general I guess we do have to cheat on this part. For cases where the client explicitly doesn't support Brotli all we can do is 406 it... Unless we want compression to happen server side, which I don't think we do.

swatanabe: > Unless we want compression to happen server side at the time of the request, which... I don't think we do?

Fallback *de*compression on the server side would be more viable.

rust/psibase/src/main.rs
James-Mart: Why? That seems odd/unconventional and less user-friendly

libraries/psibase_http/handle_request.cpp
```diff
@@ -647,6 +647,12 @@ namespace psibase::http
 
             auto        startTime = steady_clock::now();
             HttpRequest data;
+            for (auto iter = req.begin(); iter != req.end(); ++iter)
+            {
+               data.headers.emplace_back(
+                   HttpHeader{std::string(iter->name_string()), std::string(iter->value())});
```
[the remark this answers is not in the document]
James-Mart: Should I just whitelist the request headers that are checked?

> may reveal secrets to services

It should only reveal secrets to the default or registered http server services, though, right? Do you mean it may reveal secrets to node operators?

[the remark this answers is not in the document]
swatanabe: I'm mainly thinking of secrets that are shared between the client and the node operator. At the moment there's only the native login token, which is only for `x-admin` and therefore explicitly excluded from wasm.  Any tokens involved in billing for queries could become a problem in the future though.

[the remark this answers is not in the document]
swatanabe: Also, proxies might not always remove authentication headers.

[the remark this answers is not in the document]
James-Mart: I am thinking that we want to make this as similar to using your own webserver as possible, right? Whitelisting therefore feels too restrictive to me.

I think maybe it makes more sense to blacklist the particular headers that we don't want to expose to wasm.

If so, I am thinking the exclude list is just the authorization headers:
* `Authorization`
* `Proxy-Authorization`

Do you agree with this?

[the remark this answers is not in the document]
swatanabe: I agree that whitelisting feels too restrictive. OTOH, it's all too easy for blacklisting to miss something accidentally. I don't think there's a perfect solution, here. I just want to make sure that the issue has been considered.

libraries/psibase/common/include/psibase/Rpc.hpp
```diff
@@ -14,25 +14,26 @@ namespace psibase
...
-      std::string       contentType;  ///< "application/json", "text/html", ...
-      std::vector<char> body;         ///< Request body, e.g. POST data
-      PSIO_REFLECT(HttpRequest, host, rootHost, method, target, contentType, body)
+      std::string             host;         ///< Fully-qualified domain name
+      std::string             rootHost;     ///< host, but without service subdomain
+      std::string             method;       ///< "GET" or "POST"
+      std::string             target;       ///< Absolute path, e.g. "/index.js"
+      std::string             contentType;  ///< "application/json", "text/html", ...
+      std::vector<char>       body;         ///< Request body, e.g. POST data
+      std::vector<HttpHeader> headers;      ///< HTTP Headers
```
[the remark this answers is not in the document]
James-Mart: Moved before body, didn't touch `HttpReply`

services/user/Sites/plugin/src/lib.rs
```diff
@@ -32,12 +33,43 @@ fn normalize_path(path: &String) -> String {
...
+            | "application/rss+xml"
+            | "application/atom+xml"
+            | "image/svg+xml"
+            | "font/ttf"
+            | "font/otf"
+            | "application/wasm"
+    )
+}
+
+fn compress_content(content: &[u8], content_type: &str) -> (Vec<u8>, Option<String>) {
```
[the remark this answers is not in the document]
James-Mart: Fixed

libraries/psibase_http/handle_request.cpp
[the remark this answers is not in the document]
James-Mart: I decided to use blacklisting because I value being able to expect a header to pass through by default

programs/psinode/tests/test_psibase.py
James-Mart: Fallback decompression implemented for Brotli

rust/psibase/src/main.rs
James-Mart: Nice, done

James-Mart:
I snuck two more features into this PR:

HEAD requests are now supported to get headers for requests to Sites, and I also added an action for checking that a path is valid for a particular site. Not used anywhere currently, but was talking about needing something like that for the Registry app with @sparkplug0025 earlier.

services/user/Sites/src/Sites.cpp
```diff
@@ -6,13 +6,41 @@
...
 using namespace psibase;
 
 namespace SystemService
 {
    namespace
    {
+      std::string to_lower(const std::string& str)
+      {
+         std::string lower;
+         std::transform(str.begin(), str.end(), std::back_inserter(lower), ::tolower);
```
swatanabe, line 22: `std::ranges` is better

services/user/Sites/src/Sites.cpp
```diff
@@ -6,13 +6,41 @@
...
+         return lower;
+      }
+
+      template <typename T>
+      bool contains(const std::vector<T>& vec, const T& value)
+      {
+         return std::find(vec.begin(), vec.end(), value) != vec.end();
+      }
+
+      std::vector<std::string> split(const std::string& str, char delimiter)
```
swatanabe, line 32: What about `std::views::split`?

services/user/Brotli/plugin/src/errors.rs
```diff
@@ -0,0 +1,26 @@
+use crate::bindings::host::common::types::{Error, PluginId};
+use psibase::services::psi_brotli;
+
+#[derive(PartialEq, Eq, Hash)]
+pub enum ErrorType {
+    InvalidQuality,
```
swatanabe, line 6: As of #903, this entire file can be reduced to
```
psibase::plugin_error! {
    pub ErrorType
    InvalidQuality(quality: u8) => "Invalid quality: {quality}"`,
}
```

services/user/Sites/plugin/wit/world.wit
```diff
@@ -16,13 +16,19 @@ interface sites {
     /// 
     /// This file will be accessible at the subdomain of the caller, 
     /// at the specified path.
+    /// 
+    /// Maximum quality Brotli file compression is automatically applied to 
+    /// the file if compression on its content type is supported.
```
swatanabe, line 21: Why is the default different from the CLI?

services/user/Sites/src/Sites.cpp
```diff
@@ -105,27 +254,144 @@ namespace SystemService
...
    {
       Tables tables{};
       auto   table = tables.template open<SitesContentTable>();
 
       check(path.starts_with('/'), "Path doesn't begin with /");
+      auto hash = psio::detail::seahash(std::string_view(content.data(), content.size()));
+
+      if (contentEncoding)
+      {
+         check(std::find(validEncodings.begin(), validEncodings.end(), *contentEncoding) !=
```
swatanabe, line 383: Since validEncodings is static, I would make sure it's sorted and then use `std::ranges::binary_search`.

services/user/Sites/src/Sites.cpp
```diff
@@ -53,6 +83,125 @@ namespace SystemService
...
+          "font-src 'self' https:;"                                        //
+          "script-src 'self' 'unsafe-eval' 'unsafe-inline' blob: https:;"  //
+          "img-src *;"                                                     //
+          "style-src 'self' 'unsafe-inline';"                              //
+          "frame-src *;"                                                   //
+          "connect-src * blob:;"                                           //
+          ;
+
+      // Accepted content encodings
+      const std::array<std::string, 2> validEncodings = {"br", "gzip"};
```
swatanabe, line 112: I prefer built-in arrays for this, just because the size can be deduced. Also, using `string_view` instead of string allows static initialization.

services/user/Sites/src/Sites.cpp
```diff
@@ -53,6 +83,125 @@ namespace SystemService
...
+         //
+         // Justification on ignoring qvalues from RFC 7231:
+         //   "Note: Most HTTP/1.0 applications do not recognize or obey qvalues
+         //   associated with content-codings.  This means that qvalues might
+         //   not work..."
+         std::vector<std::string> rejected;
+         std::vector<std::string> encodings;
+         for (const auto& encoding : accepted_encodings)
+         {
+            auto reject_index = encoding.find(";q=0");
```
swatanabe, line 137: - spaces are allowed around the `;`
- This will match `;q=0.5`

services/user/Sites/src/Sites.cpp
```diff
@@ -6,13 +6,41 @@
...
    {
+      std::string to_lower(const std::string& str)
+      {
+         std::string lower;
+         std::transform(str.begin(), str.end(), std::back_inserter(lower), ::tolower);
+         return lower;
+      }
+
+      template <typename T>
+      bool contains(const std::vector<T>& vec, const T& value)
```
swatanabe, line 27: I've been considering bumping to c++23, which has `std::ranges::contains`. What do you think?

services/user/Sites/src/Sites.cpp
```diff
@@ -105,27 +254,144 @@ namespace SystemService
...
+                     {
+                        auto decompressorAccount = AccountNumber{*decompressor};
+
+                        check(to<Accounts>().exists(decompressorAccount),
+                              "[Fallback decompression error] Decompressor account not found");
+
+                        Actor<DecompressorInterface> decoder(Sites::service, decompressorAccount);
+                        content->content = decoder.decompress(content->content);
+                     }
+                     headers.push_back({"Content-Encoding", "identity"});
```
swatanabe, line 337: > Note that the coding named "identity" is reserved for its special role in [Accept-Encoding](https://www.rfc-editor.org/rfc/rfc9110#field.accept-encoding) and thus SHOULD NOT be included

services/user/Sites/src/Sites.cpp
```diff
@@ -105,27 +254,144 @@ namespace SystemService
...
+            //
+            // ...If a non-empty Accept-Encoding header field is present in a request and
+            // none of the available representations for the response have a content
+            // coding that is listed as acceptable, the origin server SHOULD send a
+            // response without any content coding unless the identity coding is
+            // indicated as unacceptable.""
+            auto acceptEncoding = get_header_value(request, "Accept-Encoding");
+            if (acceptEncoding)
+            {
+               auto trimmed           = std::regex_replace(*acceptEncoding, std::regex("\\s+"), "");
```
swatanabe, line 288: Removing all spaces allows nonsensical values like `b       r`

services/user/Sites/plugin/wit/world.wit
James-Mart: I think I was thinking CLI is more likely to be used in scripts for uploading large amounts of content, in which case level 11 can be quite slow.

Maybe they should just be the same. Also, not sure why I didn't parameterize it here. I'll address both of those

James-Mart: Maybe a front-end GUI should default to max compression, because it's more likely to be used by people who will just accept the default, and they might as well get max compression.

services/user/Sites/src/Sites.cpp
James-Mart: +1 would be nice

services/user/Sites/src/Sites.cpp
```diff
@@ -53,6 +80,188 @@ namespace SystemService
...
+          "script-src 'self' 'unsafe-eval' 'unsafe-inline' blob: https:;"  //
+          "img-src *;"                                                     //
+          "style-src 'self' 'unsafe-inline';"                              //
+          "frame-src *;"                                                   //
+          "connect-src * blob:;"                                           //
+          ;
+
+      // Accepted content encodings
+      constexpr std::string_view validEncodings[] = {
+          // KEEP SORTED
```
swatanabe, line 110: better than a comment:
```
static_assert(std::ranges::is_sorted(validEncodings));
```

services/user/Sites/src/Sites.cpp
```diff
@@ -53,6 +80,188 @@ namespace SystemService
...
+      // Content-coding-identifier, e.g. "br", "gzip", etc.
+      using cci = std::string;
+      // The account that holds the service code implementing the DecompressorInterface
+      using decompressor_account                                              = std::string;
+      const std::array<std::pair<cci, decompressor_account>, 1> decompressors = {
+          {{"br", "psi-brotli"}}  //
+      };
+
+      // Remove leading and trailing whitespace from a string
+      void trimSpaceAround(std::string& str)
```
swatanabe, line 124: We do have boost available: [boost::algorithm::trim](https://www.boost.org/doc/libs/1_86_0/doc/html/doxygen/trim_8hpp_1a43176b48fed44bc161b5ed635bea9cbe.html).

James-Mart: Ooh, thank you

swatanabe (approved):

PR #884, opened 2024-10-10
Supervisor: Handle multiple requests simultaneously

brandonfancher:
This implements a promise queue. Now supervisor can handle multiple requests simultaneously instead of just one.


cool-ant (approved):
:shipit:

James-Mart (approved):

PR #890, opened 2024-10-17
Remove CRTP from psibase::Service

swatanabe:
- Update to C++23
- Ubuntu 20.04 is no longer supported


swatanabe:
Requires GCC 14. Ubuntu 24.04 has GCC 13.

libraries/psibase/common/include/psibase/Service.hpp
James-Mart, line 1: Without the crtp, I'm not sure there's much benefit to the inheritance. Why not just move `emit` and `events` to a separate events library?

swatanabe: I didn't change the syntax of calling `emit()`. A free function would need the service to be passed explicitly.

swatanabe: Deducing `this` can replace most uses of CRTP.

James-Mart (approved):

PR #899, opened 2024-10-29
Fixing anyhow references in service macro

cool-ant:


rust/psibase_macros/psibase-macros-lib/src/tests/basics.rs
```diff
@@ -48,4 +48,25 @@ mod tests {
 
         let _after = service_macro_impl(attr, before);
     }
+
+    #[test]
+    // `service` macro should expand without requiring a `use` to provide its deps
```
swatanabe, line 53: It always expanded successfully. The generated code just didn't compile.

cool-ant: ok, comment updated for accuracy, and this test is part of the suite that runs as part of CI/CD, so we know that it expands *and compiles*

swatanabe: The original comment accurately described the test. It's just not a very useful test.

swatanabe (approved):

PR #900, opened 2024-10-30
json/fracpack conversions using a Schema

swatanabe:
This uses `serde_json::Value` and not `Serializer`/`Deserializer` because these traits are not object-safe, which we need for the customization layer.


rust/fracpack/src/schema.rs
```diff
@@ -514,3 +517,1703 @@ tuple_impl!(T0 T1 T2 T3 T4 T5 T6 T7 T8 T9 T10 T11 T12);
...
+            has_unknown: false,
+            known_end: true,
+        }
+    }
+    fn read_fixed(&mut self, len: u32) -> Result<Self, Error> {
+        assert!(self.known_end);
+        let Some(pos) = self.pos.checked_add(len) else {
+            return Err(Error::ReadPastEnd);
+        };
+        if self.pos as usize > self.data.len() {
```
James-Mart, line 543: I think you meant to check `pos` here, rather than `self.pos`, here.

rust/fracpack/src/schema.rs
```diff
@@ -514,3 +517,1703 @@ tuple_impl!(T0 T1 T2 T3 T4 T5 T6 T7 T8 T9 T10 T11 T12);
...
+        src: &[u8],
+        pos: &mut u32,
+    ) -> Result<serde_json::Value, Error> {
+        let len = if ty.is_variable_size() {
+            u32::unpack(src, pos)?
+        } else {
+            ty.fixed_size()
+        };
+        let start = *pos;
+        let end = *pos + len;
```
James-Mart, line 743: Is there any risk this could exceed `src.len`?

swatanabe: No. The data is validated before we get here.

James-Mart (approved):

PR #904, opened 2024-11-05
Accounts plugin support connection tokens

James-Mart:
## Changes

* Refactors `accounts` app into several modules
* Splits the `accounts` plugin api into different interfaces
* Adds connection token support to accounts
* Cleans up a bunch of test UI widgets from DemoApp1
* Adds a second plugin in the accounts namespace: "account-tokens" - houses the definitions of two token types: ConnectionTokens and InviteTokens, and provides serialization/deserialization for both.

## Connection tokens

The following flow shows an account creating a new connection token in response to a user clicking, "login" or "login with new account" or something

```mermaid
sequenceDiagram
    App->>Accounts: create_connection_token
    Accounts-->>App: return <token>
    note over App: redirect to accounts.psibase/login?token=<token>
    Accounts->>Accounts: decode_token
    Accounts-->>Accounts: login_direct(app, user)
    note over Accounts: redirect to app.psibase
```

After this, an account is connected to (and automatically logged-in to) the app. An app can switch the logged-in account between any connected accounts at will without redirection the accounts app.

## Invite tokens

Currently, the flow for an invite is similar. 
Consider a user wants to invite another person to an app.

Then the flow is something like:
```mermaid
sequenceDiagram

actor User
actor Friend
participant App
participant Accounts
participant Invite

User->>App: Click "generate invite link"
App->>Invite: create_invite_token
Invite-->>App: return <token>
App-->>User: <accounts.psibase/login?token=token>
note over User: Send to friend offline

note over Friend: Navigate to invite link
Accounts->>Accounts: decode_token
alt existing account
    Accounts-->>Accounts: login_direct(app, user)
else new account
    Accounts->>Invite: accept-create (creates new account)
    Accounts-->>Accounts: login_direct(app, user)
end
note over Accounts: redirect to app.psibase
```


James-Mart:
> I'm getting this error upon build

@Velua Odd, I can't reproduce this. At the top of that bindings.rs do you see `// Generated by 'wit-bindgen' 0.25.0` 
Also what's your `rustc --version`?

swatanabe:
> > I'm getting this error upon build
> 
> @Velua Odd, I can't reproduce this. At the top of that bindings.rs do you see `// Generated by 'wit-bindgen' 0.25.0` Also what's your `rustc --version`?

I've seen the same error in a different file. It looks like a bug in rustfmt. `rustfmt 1.7.1-stable (f6e511e 2024-10-15)`

swatanabe:
Note: I manually edited the file that I saw it in to remove the trailing space, and then `cargo fmt` worked.

services/system/Accounts/plugin/src/db/tokens_table.rs
```diff
@@ -0,0 +1,95 @@
...
+            .map(|t| <ConnectionTokens>::unpacked(&t).expect("Failed to unpack connection tokens"))
+            .unwrap_or_default();
+
+        tokens.remove_expired();
+
+        let token = ConnectionToken::new(sender);
+        let packed_token = &token.packed();
+
+        tokens.add(token);
+        Keyvalue::set(&DbKeys::CONNECT_TOKENS, &tokens.packed())
```
swatanabe, line 82: Normally I would expect tokens to be either encrypted (for privacy) or signed with a MAC (if the tokens are consumed by the same entity that created them) or asymmetric (if the token creator is separate from the consumer) key. This allows the token consumer to be stateless, and avoids having to worry about limits on the number of tokens.

swatanabe: In particular, the current setup makes it possible for an attacker to guess a valid token. I don't know whether that's exploitable, but it makes me wonder why a token is needed at all.

James-Mart: I'd like the entire client-side DB to eventually be encrypted.

Hmm, I was going to say the tokens can't be signed because there's nothing to sign with, but what if the accounts app generates it's own key in AuthSig::Keyvault client side, and signs the token.... that would probably actually work, wouldn't it

James-Mart: > In particular, the current setup makes it possible for an attacker to guess a valid token. I don't know whether that's exploitable

Malicious app could open a connection page for a third-party app, prompting the user to link their account to the third-party app. I think this would mostly just confuse the user? I'm not really sure what the motive would be for such an attacker. 

I think the this token exists out of an abundance of caution. Makes it impractical to confuse the user in this way, and prevents honest actors from accidentally prompting users for a connection to a different app. Do you prefer a different approach @swatanabe?

###  Impracticality of guessing a token

While it's theoretically possible to guess a token, I don't think it's practical. The user would need to generate but fail to consume a connection token (uncommon). The mal app would need to somehow know the app for which the user has an available token (impossible), and would need to correctly guess it before expiry. Guessing is probably at best a 1/300 chance since the expiry time in seconds is baked into the token. And there's no way to check if a guessed token is valid (except with user interaction).

services/system/Accounts/plugin/src/db/apps_table.rs
```diff
@@ -0,0 +1,134 @@
...
+pub struct AppsTable {
+    app: AppDetails,
+}
+impl AppsTable {
+    pub fn new(app: &AppDetails) -> Self {
+        Self { app: app.clone() }
+    }
+
+    fn prefix(&self) -> String {
+        // Do not namespace app data by protocol
```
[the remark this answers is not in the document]
James-Mart: As discussed, comment has been changed

services/system/Accounts/plugin/src/db/apps_table.rs
```diff
@@ -0,0 +1,134 @@
...
+
+        // Add to connected accounts if not already present
+        let connected_accounts = Keyvalue::get(&self.key(DbKeys::CONNECTED_ACCOUNTS));
+        let mut connected_accounts = connected_accounts
+            .map(|c| <ConnectedAccounts>::unpacked(&c).unwrap())
+            .unwrap_or_default();
+        connected_accounts.add(&user);
+
+        // Add to connected apps if not already present
+        let connected_apps = Keyvalue::get(&DbKeys::CONNECTED_APPS);
```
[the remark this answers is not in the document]
James-Mart: Good catch, fixed

services/system/Accounts/plugin/src/db/apps_table.rs
```diff
@@ -0,0 +1,134 @@
...
+use url::Url;
+
+use crate::db::keys::DbKeys;
+
+#[derive(Pack, Unpack)]
+struct ConnectedAccounts {
+    accounts: Vec<String>,
+}
+
+impl Default for ConnectedAccounts {
```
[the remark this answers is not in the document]
swatanabe: There are several similar `impl Default`s, too.

services/system/Accounts/plugin/src/db/apps_table.rs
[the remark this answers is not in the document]
swatanabe: If we care about limiting the key length, the scheme is probably the least significant part. I don't know that it's worth stripping it off byt itself. If the scheme is redundant, the port and most of the host are also redundant.

services/system/Accounts/plugin/src/db/tokens_table.rs
James-Mart: None of this is really used yet. Was prepping it for John to integrate
 
```mermaid
    App->>Accounts: active-app.create_connection_token
    Accounts-->>App: return <token>
    note over App: redirect to accounts.psibase/login?token=<token>
    Accounts->>Accounts: admin.decode_token
    Accounts-->>Accounts: login_direct(app, user)
    note over Accounts: redirect to app.psibase
```

services/user/Supervisor/ui/src/supervisor.ts
```diff
@@ -60,15 +60,17 @@ export class Supervisor implements AppInterface {
             "Redundant setting parent origination",
         );
 
-        if (callerOrigin === supervisorDomain) {
-            console.info(
-                "TODO: handle calls directly from the rootdomain itself",
-            );
+        if (callerOrigin === siblingUrl(null, null, null, true)) {
```
Velua, line 63: @James-Mart

services/system/Accounts/plugin/src/db/apps_table.rs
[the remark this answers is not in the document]
James-Mart: All were addressed

services/user/Supervisor/ui/src/supervisor.ts
James-Mart: I removed my origin rewrite and left the app explicitly set to homepage

James-Mart:
I believe all review topics have been addressed

James-Mart (commented):
John pointed out offline that we are missing the ability for the `accounts` app to get the list of all accounts.

I am putting this PR back into draft to add that support

services/system/Accounts/plugin/account_tokens/src/lib.rs
```diff
@@ -0,0 +1,32 @@
...
+        match token {
+            Token::InviteToken(invite_token) => URL_SAFE.encode(&invite_token.packed()),
+            Token::ConnectionToken(connection_token) => URL_SAFE.encode(&connection_token.packed()),
+        }
+    }
+
+    fn deserialize_token(token: String) -> Option<Token> {
+        let decoded = URL_SAFE.decode(&token).ok()?;
+
+        if let Ok(invite_token) = <InviteToken>::unpacked(&decoded) {
```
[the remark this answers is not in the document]
James-Mart: I see, I think that should address it? 
The users of `deserialize_token` were already expecting to get back the variant itself

swatanabe (approved):

PR #911, opened 2024-11-12
Moved `api` arg

cool-ant:
Moved `--api` / `-a` arg to first position to mirror function of same `psibase` arg.


swatanabe:
Why? I don't consider the argument order in psibase to be particularly great.

cool-ant:
This approach achieves 2 things
1) make the position of this arg the same between psibase and cargo-psibase, eliminating one more exception-to-the-rule we need to remember when we use our tools
2) it eliminates a small amount of code replication that will only get worse as we add other commands to cargo-psibase

swatanabe:
> > Why? I don't consider the argument order in psibase to be particularly great.
> 
> Symmetry is nice though

If symmetry is the goal, I'd rather change `psibase`

swatanabe:
`clap` has a `global` flag to allow options to appear in any position.

swatanabe:
Also, the `api` flag is only meaningful for commands that interact with the chain. Another way to reduce the code duplication is to bundle them in a struct and use `flatten`.

cool-ant:
Closing this PR in the interest of a clean PR to implement what we decided (applying these changes to psibase rather than to cargo-psibase

PR #915, opened 2024-11-15
Convert block time to microseconds

### LINKED ISSUE #242: Switch block time to microseconds

swatanabe:
Blocks are still always created at a 1-second intervals.

- Use 64-bit time
- Use `std::chrono` types in C++
- Better arithmetic for `TimePoint[U]Sec` in rust
- Make Schema JSON match direct JSON serialization

Fixes #242


libraries/psio/src/chrono.cpp
```diff
@@ -0,0 +1,133 @@
...
+         return false;
+      };
+      uint32_t y, m, d, h, min, sec;
+      if (!parse_uint(y, 4) || !parse_char('-') || !parse_uint(m, 2) || !parse_char('-') ||
+          !parse_uint(d, 2))
+         return false;
+      if (!parse_char('T') || !parse_uint(h, 2) || !parse_char(':') || !parse_uint(min, 2) ||
+          !parse_char(':') || !parse_uint(sec, 2))
+         return false;
+      if (h > 24 || min > 59 || sec > 60)
```
James-Mart, line 88: This may allow some invalid times, namely at hr 24. You could check for 
`if (hr == 24 && (min != 0 || sec != 0))`

libraries/psio/src/chrono.cpp
```diff
@@ -0,0 +1,133 @@
...
+      }
+      else if (parse_char('-') || parse_str("\u2212"))
+      {
+         if (!parse_uint(h, 2))
+            return false;
+         if (parse_char(':') & !parse_uint(min, 2))
+            return false;
+         if (h > 14 || min > 59)
+            return false;
+         if (h == 0 && min == 0)
```
James-Mart, line 127: I can't find anything specifically saying this is allowed or disallowed, but shouldn't this technically represent UTC?

libraries/psio/src/chrono.cpp
```diff
@@ -0,0 +1,133 @@
...
+      if (parse_char('Z'))
+      {
+      }
+      else if (parse_char('+'))
+      {
+         if (!parse_uint(h, 2))
+            return false;
+         if (parse_char(':') & !parse_uint(min, 2))
+            return false;
+         if (h > 14 || min > 59)
```
James-Mart, line 115: Is it possible for `min` to be uninitialized here, namely if timezone minute offsets aren't used?

libraries/psio/src/chrono.cpp
```diff
@@ -0,0 +1,133 @@
...
+      result = std::chrono::sys_days(std::chrono::year{static_cast<int>(y)} / m / d)
+                       .time_since_epoch()
+                       .count() *
+                   86400 +
+               h * 3600 + min * 60 + sec;
+      nsec = 0;
+      if (parse_char('.') || parse_char(','))
+      {
+         uint32_t scale = 100000000;
+         while (s != end && *s >= '0' && *s <= '9')
```
James-Mart, line 99: Not sure if there's a bug here but I'm just thinking if s contains more than 9 digits. I guess the value of nsec would remain the same... 
could consider early termination if scale == 0 though

rust/psibase/src/tester.rs
```diff
@@ -122,16 +122,18 @@ impl Chain {
     /// then starts a new block 1 second after the most recent.
     ///
     /// TODO: Support sub-second block times
```
James-Mart, line 124: Outdated comment

libraries/psio/src/chrono.cpp
```diff
@@ -0,0 +1,133 @@
...
+         }
+      }
+      if (parse_char('Z'))
+      {
+      }
+      else if (parse_char('+'))
+      {
+         if (!parse_uint(h, 2))
+            return false;
+         if (parse_char(':') & !parse_uint(min, 2))
```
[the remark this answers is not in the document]
swatanabe: It's intentionally not short-circuiting. +0000 and +00:00 are both allowed

libraries/psio/src/chrono.cpp
swatanabe: ISO 8601 requires UTC to be `+`.

libraries/psio/src/chrono.cpp
swatanabe: `parse_int` always initializes `min`.

libraries/psio/src/chrono.cpp
swatanabe: It still needs to consume the digits from the input string.

James-Mart (approved):

PR #917, opened 2024-11-15
Remove deprecated event-related code

James-Mart:


services/user/Symbol/include/services/user/Symbol.hpp
```diff
@@ -79,11 +69,7 @@ namespace UserService
    );
    PSIBASE_REFLECT_EVENTS(Symbol);
    PSIBASE_REFLECT_HISTORY_EVENTS(Symbol,
-      method(symCreated, prevEvent, symbol, owner, cost),
```
swatanabe, line 82: Why remove this?

James-Mart: I removed events for which I didn't see an obvious benefit, perhaps too aggressively. I will add the `symCreated` event back.

swatanabe (approved):

PR #926, opened 2024-11-20
Update JS dependencies

swatanabe:


services/user/Chainmail/ui/package.json
```diff
@@ -41,6 +46,7 @@
         "class-variance-authority": "^0.7.0",
         "clsx": "^2.1.1",
         "date-fns": "^3.6.0",
+        "element-internals-polyfill": "^0.1.49",
```
brandonfancher, line 49: I see new dependencies introduced here: `element-internals-polyfill` and the `@codemirror/*` dependencies. Were these required by another updated package?

services/user/Tokens/ui/package.json
```diff
@@ -19,33 +21,40 @@
...
+    "react-native": "^0.76.2",
     "react-spring": "^9.7.3",
+    "react-zdog": "^1.2.2",
     "sonner": "^1.4.41",
     "tailwind-merge": "^2.3.0",
     "tailwindcss-animate": "^1.0.7",
+    "three": "^0.170.0",
     "usehooks-ts": "^3.1.0",
     "vaul": "^0.9.0",
+    "zdog": "^1.1.3",
```
brandonfancher, line 45: I'm seeing new packages introduced here too. What's that about?

services/user/Chainmail/ui/package.json
swatanabe: I added everything that yarn complained about as "missing peer dependency"

swatanabe: Is there a better way to handle that?

services/user/Tokens/ui/package.json
brandonfancher: `react-spring` is an umbrella library for multiple platforms. It would be better here to remove these and change the `react-spring` library to `@react-spring/web`. That will make the warnings disappear. The import in `AnimateNumber.tsx` will also need to be updated.

Would you like me to push a commit to this branch with those changes?

services/user/Chainmail/ui/package.json
brandonfancher: This is fine here. Doesn't affect build/bundle size much at all. And gets rid of the warnings.

brandonfancher (approved):
I pushed the fix to tokens peer dependency warnings. LGTM.

PR #927, opened 2024-11-21
Generalize argument merging for GraphQL

### LINKED ISSUE #214: Better GraphQL support for subindices
## Serving a TableIndex

Currently, to serve a TableIndex from GraphQL is very easily supported, for example:
```
struct TokenQuery
{
   auto balances() const
   {
      return tokenSys.index<BalanceTable, 0>();
   }
}:
PSIO_REFLECT(TokenQuery, method(balances))

optional<HttpReply> RTokenSys::serveSys(HttpRequest request)
{
   if (auto result = serveGraphQL(request, TokenQuery{}))
      return result;

   return nullopt;
}

```

## Serving a subindex

I want the syntax for a subindex to be similar, something like
```
auto balance(AccountNumber user) const
{ 
  return tokenSys.index<BalanceTable, 1>().subindex(user);
}
```

But the issue is that the system expects GraphQL queries for table indices have no user-defined parameters, and all params are the autogenerated ones, `gt, ge, lt, le, first, last, before, after`.

Currently the only way to do it is to call makeConnection automatically, for example (from the Auth RPC service):
```
struct AuthEcQuery
   {
      auto accWithKey(psibase::PublicKey key) const
      {
         auto idx = AuthEcSys::Tables{AuthEcSys::service}
                        .open<AuthEcSys::AuthTable>()
                        .getIndex<1>()
                        .subindex(key);
         return makeConnection<Connection<AuthRecord, "AuthConnection", "AuthEdge">>(
             idx, {}, {}, {}, {}, {}, {}, {}, {});
      }
   };
   PSIO_REFLECT(AuthEcQuery, method(accWithKey, key))
```

Is there any way we can get simpler syntax for subindices? @tbfleming @swatanabe

swatanabe:
- Add TransformedConnection which applies a function to the elements of an index.
- Use new features in RTokens
- Add tests for Tokens queries

Fixes #214

### DIFF
--- a/services/user/Tokens/src/RTokens.cpp
+++ b/services/user/Tokens/src/RTokens.cpp
@@ -108,6 +108,18 @@ struct key_suffix_unqual<SplitKey<Rest, T>, S>
    using type = Rest;
 };
 
+template <typename Rest, typename T, typename... S>
+struct compatible_tuple_prefix<SplitKey<Rest, T>, std::tuple<S...>>
+{
+   static constexpr bool value = true;
+};
+
+template <typename Rest, typename T, typename... S>
+struct key_suffix_unqual<SplitKey<Rest, T>, std::tuple<S...>>
+{
+   using type = Rest;
+};
+
 using namespace TokenQueryTypes;
 struct TokenQuery
 {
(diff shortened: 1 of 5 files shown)


James-Mart (approved):

services/user/Tokens/src/RTokens.cpp
```diff
@@ -118,81 +130,52 @@ struct TokenQuery
...
-                    optional<uint32_t>    last,
-                    optional<std::string> before,
-                    optional<std::string> after) const
+   auto userCredits(AccountNumber user) const
    {
-      vector<Credit> credits;
-
       auto tokenTypeIdx = tokenService.open<TokenTable>().getIndex<0>();
       auto creditIdx    = tokenService.open<SharedBalanceTable>().getIndex<0>().subindex(
-          SplitKey<std::tuple<TID, AccountNumber>, AccountNumber>{user});
```
James-Mart, line 158: Why did you change this to BalanceKey? It is the same types, but BalanceKey is related to the BalanceTable, not the SharedBalanceTable

swatanabe: The key needs to be a struct with named members, not a tuple, so it can be passed as an argument (ge/gt/lt/le).

Issue #914, opened 2024-11-15
Add staged-tx documentation


James-Mart:
# Background

There are many possible auth services that share the feature that they allow 1-N accounts that are _not_ the sender of the tx to authorize the transaction by some configurable logic.

Currently implemented auth services:

- `auth-delegate` - Authority delegated to owner
- `producers` - Authority delegates to a threshold of currently active producers

Other examples:

- Authority delegated to a configurable threshold of signatures from a static set of potential signers (simple msig)
- Authority delegated to a configurable threshold of signatures from a dynamic set of potential signers (dynamic msig)
- Authority delegated to a threshold of total reputation, stake-weight, etc., from a static/dynamic set of potential signers

The goal of this proposal is to understand the common needs of auth services like this, and to specify an interface(s) that can be used to support such services in a generalized way. Furthermore, the goal is to create a common front-end infrastructure that can be used to allow accounts that are authorized with these auth services to be used in a user-friendly way.

# Proposal

The common needs of these auth services are a staging area for transactions to be proposed, accepted, and rejected.
A new interface shall be created to support "staged transactions". Staged transactions are those that undergo stages of proposal/acceptance/rejection before execution pursuant to the rules of the auth service of the sender of the staged transaction.

Another use-case for staged transactions is to allow for atomic swaps, or in general for atomic execution of dependent actions from different parties that do not trust each other.

# Server-side changes

## Service: staged-tx

This service manages transactions for auth services that allow for staged execution.

Stages:

1. Proposal
2. Acceptance (or rejection)
3. Execution (if accepted)

Note: This service interface could be folded into `transact` if we wanted to minimize execution contexts.

### Action interface

```rust
/// Proposes a new staged transaction containing the specified actions.
/// Returns the ID of the database record containing the staged transaction.
///
/// All actions must have the same sender.
///
/// * `actions` - The actions to be staged
fn propose(actions: Vec<Action>) -> u32

/// Indicates that the caller accepts the specified staged transaction
///
/// Depending on the staging rules enforced by the auth service of the sender
/// of the staged transaction, this could result in the execution of the transaction.
///
/// * `id`: The ID of the database record containing the staged transaction
/// * `txid`: The unique txid of the staged transaction
fn accept(id: u32, txid: [u8; 32])

/// Indicates that the caller rejects the staged transaction
///
/// * `id`: The ID of the database record containing the staged transaction
/// * `txid`: The unique txid of the staged transaction
fn reject(id: u32, txid: [u8; 32])

[...]

PR #934, opened 2024-11-29
Staged transaction service

James-Mart:
Adds staged-transaction service.
Currently only supported by the `auth-any` auth service.

Further reading, see [this issue](https://github.com/gofractally/psibase/issues/914).


services/system/AuthAny/src/AuthAny.cpp
```diff
@@ -1,29 +1,46 @@
...
-   void AuthAny::canAuthUserSys(psibase::AccountNumber user)
+   void AuthAny::canAuthUserSys(AccountNumber user)
    {
       if (enable_print)
          std::printf("user_check\n");
    }
+
+   void AuthAny::stagedAccept(uint32_t stagedTxId, AccountNumber actor)
+   {
+      check(getSender() == "staged-tx"_a, "can only be called by staged-tx");
```
swatanabe, line 34: `StagedTx::service`

services/system/StagedTx/service/src/lib.rs
```diff
@@ -0,0 +1,440 @@
...
+            "Only the first sender of the staged tx can execute it",
+        );
+
+        staged_tx.action_list.actions.iter().for_each(|action| {
+            let _ = Transact::call().runAs(action.clone(), Vec::new());
+        });
+
+        staged_tx.emit(StagedTxEvent::EXECUTED);
+
+        remove(id, txid);
```
swatanabe, line 364: `remove` should happen before running the actions to prevent multiple execution.

services/system/StagedTx/service/src/lib.rs
```diff
@@ -0,0 +1,440 @@
...
+    fn accept(id: u32, txid: [u8; 32]) {
+        let staged_tx = StagedTx::get(id, txid);
+
+        Response::upsert(id, true);
+
+        staged_tx
+            .staged_tx_policy()
+            .accept(staged_tx.id, get_sender());
+
+        staged_tx.emit(StagedTxEvent::ACCEPTED);
```
swatanabe, line 291: Is it okay for the `ACCEPTED` event to come after the `EXECUTED` and `DELETED` events?

services/system/StagedTx/service/src/lib.rs
```diff
@@ -0,0 +1,440 @@
...
+            (auth_service, self.id)
+        }
+
+        fn new(actions: Vec<Action>) -> Self {
+            check(
+                actions.len() > 0,
+                "Staged transaction must contain at least one action",
+            );
+
+            let sender = actions[0].sender;
```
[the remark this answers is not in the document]
James-Mart: Correct, there wasn't, because the `runAs` for the other sender will fail unless that sender's auth service allows runAsOtherReq (which, frankly, I don't know why any auth service would ever allow that).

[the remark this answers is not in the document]
James-Mart: And now there isn't for a second reason, which is that I explicitly restricted proposed staged transactions so that they must all be from the same sender to address the loophole you found.

services/system/StagedTx/service/src/lib.rs
James-Mart: No, fixed.

services/system/StagedTx/service/src/lib.rs
James-Mart: done

services/system/AuthAny/src/AuthAny.cpp
James-Mart: done

swatanabe (approved):

PR #935, opened 2024-11-30
Minor misc updates

James-Mart:


services/user/Tokens/src/RTokens.cpp
```diff
@@ -140,7 +140,6 @@ struct TokenQuery
                                    [tokenTypeIdx = std::move(tokenTypeIdx)](auto&& balance)
                                    {
                                       auto tokenOpt = tokenTypeIdx.get(balance.key.tokenId);
-                                      check(tokenOpt.has_value(), "Invalid token type");
```
swatanabe, line 143: Does this actually make a measurable difference? Catching broken invariants is probably more helpful than the performance benefit of skipping the check.

swatanabe (approved):

services/user/Tokens/src/RTokens.cpp
James-Mart: Reverted this

PR #936, opened 2024-11-30
Delete old rust tests

James-Mart:


swatanabe (commented):
I wondered why ctest wasn't failing `rs-sites-test` and `rs-elections-test`, but it looks like they're building the wrong crate. Please delete them from `rust/CMakeLists.txt` anyway.

swatanabe (approved):

PR #938, opened 2024-12-02
Improve boot related cli output

James-Mart:
# `psinode` start message

Old message on initial execution of `psinode`: `Need genesis block; use 'psibase boot' to boot chain`
New message:
`Node is not connected to any psibase network.`

If the x-admin app is hosted at a subdomain, the message is further improved to:
`Node is not connected to any psibase network. Visit '$protocol//$subdomain.psibase.127.0.0.1.sslip.io:$port' for node setup`
* Where `$protocol` is `https:` if there is an https listen directive, `http:` otherwise.
* Where `$subdomain` is the subdomain at which the x-admin app is hosted, (usually `x-admin.`)
* Where `$port` is the listen port, https preferred, http fallback

For example, message might be:
`Node is not connected to any psibase network. Visit 'https://x-admin.psibase.127.0.0.1.sslip.io:8070' for node setup`

# `psibase` boot message

Old message: `Ok.`
New message: `Successfully booted $api`
* Where `$api` is whatever URL was passed to the `-a` flag during boot

For example:
`Successfully booted http://psibase.127.0.0.1.sslip.io:8069/`


programs/psinode/main.cpp
```diff
@@ -2374,8 +2374,48 @@ void run(const std::string&              db_path,
...
+                  }
+               }
+
+               if (!xAdminSubdomain.empty())
+               {
+                  std::string protocol;
+                  std::string port;
+                  for (const auto& listen : http_config->listen)
+                  {
+                     if (std::holds_alternative<psibase::http::tcp_listen_spec<true>>(listen))
```
swatanabe, line 2394: `std::get_if` can test and extract at once.

rust/psibase/src/main.rs
```diff
@@ -729,7 +729,7 @@ async fn boot(
         progress.inc(1)
     }
     if !args.suppress_ok {
-        println!("Ok");
+        println!("Ok. Successfully booted {}", args.api);
```
swatanabe, line 732: "Ok" is redundant

programs/psinode/main.cpp
```diff
@@ -2374,8 +2374,48 @@ void run(const std::string&              db_path,
...
+                     {
+                        const auto& spec = std::get<psibase::http::tcp_listen_spec<false>>(listen);
+                        protocol         = "http://";
+                        port             = std::to_string(spec.endpoint.port());
+                     }
+                  }
+                  xAdminSubdomain = protocol + xAdminSubdomain + ":" + port;
+               }
+
+               std::string message = "Node is not yet booted. To boot, use the 'psibase boot' CLI";
```
swatanabe, line 2411: I think this should try to use the same terminology as the web UI.

programs/psinode/main.cpp
```diff
@@ -2374,8 +2374,48 @@ void run(const std::string&              db_path,
...
+               if (!xAdminSubdomain.empty())
+               {
+                  message += " or visit '" + xAdminSubdomain +
+                             "' app in browser for a graphical boot interface";
```
swatanabe, line 2415: I'd prefer to have less jargon here. Maybe say that the URL is the place to go to set up the node.

rust/psibase/src/main.rs
James-Mart: I actually explicitly left it so that it would be clear in the context of the boot subcommand what the `--suppress-ok` flag would do

programs/psinode/main.cpp
James-Mart: Are you referring to "Need genesis block;" != "Node is not yet booted."?

rust/psibase/src/main.rs
James-Mart: What about "Ok, booted http://psibase.127.0.0.1.sslip.io:8069/"

programs/psinode/main.cpp
James-Mart: I propose to change all instances of:
"Need genesis block; use 'psibase boot' to boot chain"
to
"Node is not yet booted. To boot, use the 'psibase boot' CLI"

And then also leave the extra special output exclusively in the psinode CLI. I want the "homepage" of an unbooted node to eventually redirect to x-admin.

programs/psinode/main.cpp
James-Mart: Oh, nice. Done

programs/psinode/main.cpp
James-Mart: Done

rust/psibase/src/main.rs
swatanabe: I'd actually rather get rid of `--suppress-ok` and link the message to the progress bar (`progress.finish_with_message`).

programs/psinode/main.cpp
swatanabe: I was referring to the whole thing. The setup page has "Join Network" / "Create Network".

swatanabe: I was *not* referring to what the API endpoints return for an unbooted chain.

rust/psibase/src/main.rs
James-Mart: Updating the arg and using the ProgressBar can be done separately. For now, I've updated it to "Successfully booted $api"

programs/psinode/main.cpp
James-Mart: Updated the message as discussed

swatanabe (approved):

PR #940, opened 2024-12-03
Add accounts UI + Login on dashboard

Velua:
Adds a `Login` button to the settings cog on the Dashboard, from there users can import an account and create invites. 

Invites can be accepted with an existing account or brand new account, then the user is returned to the dashboard which will recognise the new authenticated account. Users can add more accounts by returning to accounts by pressing "More" in the dashboard dropdown.


James-Mart:
You committed an old eos-vm submodule. I fix it [here](https://github.com/gofractally/psibase/pull/940/commits/53315d0f14b09e19742a1005c4528cc584d3e6c8)

brandonfancher (commented):
I reviewed from a UX perspective. Looking good! A few recommendations:

1. When you accept an invite and try to use the link again, it says that it was accepted by the account that accepted it. But when you reject an invite and try to use the link again, it says that it was rejected by `invite-sys`. I'd consider dropping the "by invite-sys." on the rejected path.
2. After you click "Accept Invite" in the create account dialog/modal and are returned to the list of accounts, the buttons "Accept Invite" and "Reject Invite" are still visible below the list of accounts. Those should go away once the invite is accepted.
3. Consider keeping the first character in the account names lowercase in the account list. "Myproducer" > "myproducer"
4. When selecting an account to log in with, consider removing the "Login" button at the bottom of the account list and simply redirecting to the homepage with the logged-in account the user clicked on. Fewer clicks and less UI.
5. The dark mode and gear buttons have no spacing between them and appear to be different style buttons.

Coming together nicely, @Velua!

services/system/Accounts/ui/src/components/AccountSelection.tsx
```diff
@@ -0,0 +1,722 @@
...
+                  className="flex p-4 shrink-0 items-center justify-center rounded-md border border-neutral-600 text-muted-foreground hover:text-primary hover:underline border-dashed"
+                >
+                  Create a new account
+                </button>
+              )}
+            </div>
+            {!isNoAccounts && (
+              <div className="my-3">
+                <Button
+                  disabled={!selectedAccount || isTxInProgress}
```
James-Mart, line 676: When creating an account, the accept invite button could be disabled if the new account name is invalid or taken. 
Otherwise (or possibly also), you need to add exception handling to onSubmit to handle the promise rejection when acceptWithNewAccount is called with an invalid name

services/system/Accounts/ui/src/components/AccountSelection.tsx
```diff
@@ -0,0 +1,722 @@
...
+
+  const {
+    data: accounts,
+    isLoading: isFetching,
+    error,
+    refetch: fetchAccounts,
+  } = useQuery({
+    queryKey: ["availableAccounts"],
+    queryFn: async (): Promise<AccountType[]> => {
+      supervisor.preLoadPlugins([{ service: "accounts" }]);
```
James-Mart, line 191: Bit excessive to preload plugins every time you want to run this query. Preload only needs to happen once on page load

services/system/Accounts/ui/src/components/AccountSelection.tsx
```diff
@@ -0,0 +1,722 @@
...
+                </button>
+              )}
+            </div>
+            {!isNoAccounts && (
+              <div className="my-3">
+                <Button
+                  disabled={!selectedAccount || isTxInProgress}
+                  onClick={() => {
+                    if (isInvite) {
+                      acceptInvite(z.string().parse(token));
```
James-Mart, line 679: The invite is actually accepted as soon as this modal is dealt with:
image

Which means that going back to this screen afterwards is incorrect:
image

There is actually no further opportunity to change your selection. If you created a new account, then you accepted the invite with that account

James-Mart: For example. Create an account (from an invite), and then close the tab before hitting the subsequent Accept/Reject button, and then reload invite link. 

It will (correctly) tell you that the invite was already accepted with the new account.

services/system/Accounts/ui/src/components/AccountSelection.tsx
```diff
@@ -0,0 +1,722 @@
...
+          <div className="max-w-lg mx-auto">
+            <div className="text-center text-muted-foreground py-2">
+              <span>
+                {isInitialLoading
+                  ? "Loading..."
+                  : isNoAccounts
+                  ? `No accounts available.`
+                  : isInvite
+                  ? `Select an account to accept invite to `
+                  : `Select an account to login to `}
```
James-Mart, line 619: I don't think "No accounts available" is an alternative to 
"Select an account to accept invite to" and "Select an account to login to"

For example, in this case:
image

Probably should still say, "Select an account to (accept invite / login) to <appname>", no?

services/system/Accounts/ui/src/components/AccountSelection.tsx
```diff
@@ -0,0 +1,722 @@
...
+        .string()
+        .array()
+        .parse(res)
+        .map((x) => ({ account: x, id: x }));
+    },
+  });
+
+  const isNoAccounts = accounts ? accounts.length == 0 : false;
+
+  console.log({ error }, "was the error");
```
James-Mart, line 210: At least wrap this in `(if error !== undefined)` or something

services/system/Accounts/ui/src/components/AccountSelection.tsx
```diff
@@ -0,0 +1,722 @@
...
+    refetch: refetchToken,
+  } = useDecodeInviteToken(token, decodedToken?.tag == "invite-token");
+
+  const { data: connectionToken, isLoading: isLoadingConnectionToken } =
+    useDecodeConnectionToken(token, decodedToken?.tag == "connection-token");
+
+  const isInitialLoading =
+    isLoadingInvite || isLoadingConnectionToken || isLoadingToken;
+
+  console.log({ decodedToken, inviteToken: invite, connectionToken });
```
James-Mart, line 269: This also results in a number of console.logs when all three inner properties are undefined

services/system/Accounts/ui/src/components/AccountSelection.tsx
```diff
@@ -0,0 +1,722 @@
...
+            },
+            account,
+          ],
+          service: "accounts",
+          intf: "admin",
+        }));
+      },
+    });
+
+  console.log(selectedAccount);
```
James-Mart, line 309: Results in printing undefined in cases where there's no selected account

James-Mart (commented):
This extra review is basically just about cleaning up console output

services/system/Accounts/ui/src/components/AccountSelection.tsx
Velua: Will immediately go to next page now.

services/system/Accounts/ui/src/components/AccountSelection.tsx
Velua: Button now disables, expects an account to be available when accepting an invite and expects it to be taken when doing an import.

services/system/Accounts/ui/src/components/AccountSelection.tsx
```diff
@@ -0,0 +1,722 @@
...
+            <div className="w-full justify-center flex">
+              <Button
+                onClick={() => {
+                  setIsCreatingAccount(false);
+                  setIsModalOpen(true);
+                }}
+                variant={isNoAccounts ? "default" : "link"}
+                className={cn({ "text-muted-foreground": !isNoAccounts })}
+              >
+                Import an account
```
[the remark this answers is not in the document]
Velua: Now only renders when working with connection tokens, not invite ones.

services/system/Accounts/ui/src/components/AccountSelection.tsx
```diff
@@ -0,0 +1,722 @@
...
+
+  const [isCreatingAccount, setIsCreatingAccount] = useState(true);
+
+  const { isDirty, isSubmitting } = form.formState;
+
+  const onSubmit = async (values: z.infer<typeof formSchema>) => {
+    if (isCreatingAccount) {
+      void (await createAccount(values.username));
+    } else {
+      void (await importAccount(values.username));
```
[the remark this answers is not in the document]
Velua: Form now performs the check before submission is allowed.

Velua:
> Would be nice if there was a header here for "Switch Accounts", maybe image

Done @James-Mart

Velua:
> I reviewed from a UX perspective. Looking good! A few recommendations:
> 
>     1. When you accept an invite and try to use the link again, it says that it was accepted by the account that accepted it. But when you reject an invite and try to use the link again, it says that it was rejected by `invite-sys`. I'd consider dropping the "by invite-sys." on the rejected path.
> 
>     2. After you click "Accept Invite" in the create account dialog/modal and are returned to the list of accounts, the buttons "Accept Invite" and "Reject Invite" are still visible below the list of accounts. Those should go away once the invite is accepted.
> 
>     3. Consider keeping the first character in the account names lowercase in the account list. "Myproducer" > "myproducer"
> 
>     4. When selecting an account to log in with, consider removing the "Login" button at the bottom of the account list and simply redirecting to the homepage with the logged-in account the user clicked on. Fewer clicks and less UI.
> 
>     5. The dark mode and gear buttons have no spacing between them and appear to be different style buttons.
> 
> 
> Coming together nicely, @Velua!

@brandonfancher 
1. Done
2. Done, just moves onto the next page.
3. Done
4. Done
5. Present on `main` will tackle in another PR.

CMakeLists.txt
```diff
@@ -385,11 +386,13 @@ psibase_package(
     SERVICE accounts
         WASM ${CMAKE_CURRENT_BINARY_DIR}/Accounts.wasm
         SERVER r-accounts
+        DATA GLOB ${CMAKE_CURRENT_SOURCE_DIR}/services/system/Accounts/ui/dist/* /
         DATA ${Plugins_OUTPUT_FILE_accounts} /plugin.wasm
         DATA ${Plugins_OUTPUT_FILE_account_tokens} /account-tokens.wasm
         INIT
     SERVICE r-accounts
         WASM ${CMAKE_CURRENT_BINARY_DIR}/RAccounts.wasm
+    DEPENDS ${Accounts_js_DEP} ${CMAKE_CURRENT_SOURCE_DIR}/services/system/Accounts/ui/dist/index.html
```
James-Mart, line 395: `${CMAKE_CURRENT_SOURCE_DIR}/services/system/Accounts/ui/dist/index.html` is redundant here, `index.html` is already included in ${Accounts_js_DEP}

Won't hold up the PR for it though

James-Mart (approved):

James-Mart:
Another issue is the loading time when I load up an invite.
image

While this is spinning, I see four calls to decode the invite, when only one should be sufficient.
image

Is this retrying in case the transaction that created the invite isn't final yet? If that's the case, we can remove the retries after https://github.com/gofractally/psibase/pull/948 gets merged.

James-Mart (changes requested):
Oops, just realized there's a bug. It seems I can't accept any invite now.
If I create an invite, and then try to create a new account with it, I get
`Uncaught (in promise) Error: Must have invite`

James-Mart (approved):

Velua:
> I think the UX can be improved in parallel so I don't want to hold up the PR any more, but now when I click on `myproducer` here it logs in instantly instead of waiting for me to confirm with "Login"
> 
> image

@James-Mart this is by design after discussion with Brandon

PR #942, opened 2024-12-03
Make C++ accept the rust schema for TimePointSec.

swatanabe:


libraries/psio/src/schema.cpp
```diff
@@ -152,12 +154,84 @@ namespace psio::schema_types
...
+         if (auto* itype = std::get_if<Int>(&type->original_type->value))
+         {
+            // read up to 8 bytes
+            std::uint64_t raw = 0;
+            for (int i = 0; i < (itype->bits + 7) / 8; ++i)
+            {
+               std::uint8_t digit;
+               if (!in.unpack<true, true>(&digit))
+                  return false;
+               raw += digit << (i * 8);
```
James-Mart, line 194: Shouldn't `digit` be casted to `uint64_t` before shifting?

swatanabe: Yes, it promotes to int, but that's not enough.

PR #948, opened 2024-12-05
make http reply only after tx irreversible

James-Mart:


James-Mart:
```
[2024-12-05T17:37:32.656Z] [warning]: onTransaction failed: service 'r-transact' aborted with message: service exceeded its memory limit EAE5D323506193D6E63D73E32E1A5EB7B3B63962C996F7510226A786B71B285E
```

swatanabe:
> Getting this on boot (although then boot is successful):
> 
> ```
> [2024-12-05T17:37:32.656Z] [warning]: onTransaction failed: service 'r-transact' aborted with message: service exceeded its memory limit EAE5D323506193D6E63D73E32E1A5EB7B3B63962C996F7510226A786B71B285E
> ```

It's making too many copies of the trace. Maybe try passing an rvalue and `move` into the put.

James-Mart:
> Maybe try passing an rvalue and move into the put.

Tried rvalue ref and also `psio::view`, neither worked

swatanabe:
> > Maybe try passing an rvalue and move into the put.
> 
> Tried rvalue ref and also `psio::view`, neither worked

It's `onBlock` failing, now.

services/system/Transact/src/RTransact.cpp
```diff
@@ -95,15 +107,34 @@ void RTransact::onBlock()
...
    }
+
+   // Send replies to all transactions in the block that are now irreversible
+   PSIBASE_SUBJECTIVE_TX
+   {
+      auto blockTracesIdx = Subjective{}.open<BlockTraceTable>().getIndex<0>();
+      for (auto i : irreversible)
+      {
+         auto index = blockTracesIdx.subindex(i);
+         for (auto trace : index)
```
swatanabe, line 131: need to erase the rows

services/system/Transact/src/RTransact.cpp
```diff
@@ -95,15 +107,34 @@ void RTransact::onBlock()
...
       irreversibleTime = r.time;
       if (r.blockNum < commitNum)
+      {
          reversible.remove(r);
+         irreversible.push_back(r.blockNum);
+      }
    }
+
+   // Send replies to all transactions in the block that are now irreversible
+   PSIBASE_SUBJECTIVE_TX
```
swatanabe, line 125: It's better to avoid sending to a socket inside a subjective transaction, as it can result in duplicate sending.

services/system/Transact/src/RTransact.cpp
```diff
@@ -95,15 +107,34 @@ void RTransact::onBlock()
...
+   {
+      auto blockTracesIdx = Subjective{}.open<BlockTraceTable>().getIndex<0>();
+      for (auto i : irreversible)
```
swatanabe, line 128: You'll probably need to switch `RTransact` to use the `psibase-service-full-malloc`, instead of `psibase-service-simple-malloc`

James-Mart: Why?

James-Mart: I was getting `libc++abi: bad_alloc was thrown in -fno-exceptions mode` with full malloc

swatanabe: Since multiple traces are being processed in the same execution, the memory usage can get really large if no memory is re-used.

services/system/Transact/src/RTransact.cpp
```diff
@@ -95,15 +107,41 @@ void RTransact::onBlock()
...
       irreversibleTime = r.time;
       if (r.blockNum < commitNum)
+      {
          reversible.remove(r);
+         irreversible.push_back(r.blockNum);
+      }
+   }
+
+   // Send replies to all transactions in the block that are now irreversible
+   std::vector<std::tuple<Checksum256, TransactionTrace>> traces;
```
swatanabe, line 125: This stores all the traces in memory at the same time.

James-Mart:
> If a transaction fails, the trace should not be reported until the transaction expires

Doesn't this mean the client could be waiting a very long time?

swatanabe:
> > If a transaction fails, the trace should not be reported until the transaction expires
> 
> Doesn't this mean the client could be waiting a very long time?

Yes, but it's the only consistent behavior when waiting for irreversible, and in a multi-node network, this wait will already happen, because the transaction isn't executed locally. With speculative execution, it would be okay to report a failure early as long as the transaction is not broadcasted. I've also considered having onTrx for a failed transaction insert a new transaction in the queue that reports the failure (using a subjective auth service that checks a list; effectively auth by current block producer).

services/system/Transact/src/RTransact.cpp
```diff
@@ -184,122 +184,244 @@ void RTransact::onTrx(const Checksum256& id, psio::view<const TransactionTrace>
...
+
+   if (waitFor == WaitFor::final_flag)
+   {
+      auto currentBlock = to<Transact>().currentBlock();
+      auto blockTxs     = Subjective{}.open<BlockTxsTable>();
+      PSIBASE_SUBJECTIVE_TX
+      {
+         blockTxs.put(BlockTxRecord{
+             .id       = id,
+             .blockNum = currentBlock.blockNum,
```
swatanabe, line 205: blockNum is only meaningful for successful transactions.

services/system/Transact/include/services/system/RTransact.hpp
```diff
@@ -84,9 +84,10 @@ namespace SystemService
    struct TraceClientRow
    {
       psibase::Checksum256         id;
+      uint8_t                      waitFor;
```
swatanabe, line 87: If the same transaction is submitted by multiple clients, they don't necessarily have the same parameters.

services/system/Transact/src/RTransact.cpp
```diff
@@ -184,122 +184,244 @@ void RTransact::onTrx(const Checksum256& id, psio::view<const TransactionTrace>
...
+      }
+   }
+
+   if (waitFor == WaitFor::final_flag)
+   {
+      auto currentBlock = to<Transact>().currentBlock();
+      auto blockTxs     = Subjective{}.open<BlockTxsTable>();
+      PSIBASE_SUBJECTIVE_TX
+      {
+         blockTxs.put(BlockTxRecord{
```
swatanabe, line 203: This can send the wrong trace in some fork switching cases. For example,
1) apply to fork A
2) fail to apply to fork B
3) switch back to fork A and make it irreversible

services/system/Transact/include/services/system/RTransact.hpp
James-Mart: For my own information, is there a reasonably common scenario in which the same transaction is submitted by multiple clients? I couldn't really think of why this was even tracked/supported.

services/system/Transact/src/RTransact.cpp
James-Mart: What is the right way to handle this? We don't have a subjective db that follows forks, right?

services/system/Transact/src/RTransact.cpp
James-Mart: Currently, it's set but unused for failed transactions. Do you want me to make it an optional and only set it on successful? Or two separate tables?

services/system/Transact/include/services/system/RTransact.hpp
swatanabe: It's probably not very useful, but if it isn't handled correctly, it can allow exploits.

services/system/Transact/src/RTransact.cpp
James-Mart: Separated

services/system/Transact/src/RTransact.cpp
```diff
@@ -184,122 +185,267 @@ void RTransact::onTrx(const Checksum256& id, psio::view<const TransactionTrace>
...
+
+         if (auto client = clientTable.get(pendingTx.id))
+         {
+            txids.push_back(pendingTx.id);
+         }
+         else
+         {  // Stop tracking an expired tx if no client is waiting for a reply.
+            pendingTxTable.erase(pendingTx.id);
+            dataTable.erase(pendingTx.id);
+            successfulTxTable.erase(pendingTx.id);
```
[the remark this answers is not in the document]
James-Mart: Actually, I can't think of a scenario when the `erase` call here will actually erase anything for either successful or failed tx tables, because they should only be added to in the case that there's a client waiting for a response, and then they are cleared separately when replies are sent. This is just a safety backup.

Regardless, it should be consistent, of course. I've now also added the call to `failedTxTable.erase`

services/system/Transact/src/RTransact.cpp
```diff
@@ -177,129 +179,327 @@ namespace
...
+         {
+            trace = psio::to_frac(tx->trace);
+            successfulTxs.erase(id);
+         }
+
+         PSIBASE_SUBJECTIVE_TX
+         {
+            if (auto tx = failedTxs.get(id))
+            {
+               if (!trace)
```
swatanabe, line 336: If you switch these conditions and make erase unconditional, it will avoid loading the failed trace into memory, when there's already a success trace.

services/system/Transact/src/RTransact.cpp
```diff
@@ -177,129 +179,327 @@ namespace
...
-   bool                          json;
-   bool                          bin;
-   PSIBASE_SUBJECTIVE_TX
+         PSIBASE_SUBJECTIVE_TX
+         {
+            if (auto row_opt = clients.get(id))
+            {
+               auto& row = *row_opt;
+               if (std::ranges::none_of(row.clients, clientFilter))
+                  abortMessage("No clients matching filter to claim a reply for this tx");
```
swatanabe, line 212: Can the caller actually guarantee this? `sendReply` checks whether there were any clients *after* calling `claimClientReply`.

services/system/Transact/src/RTransact.cpp
```diff
@@ -177,129 +179,327 @@ namespace
...
-            else
-               bin = true;
-            to<HttpServer>().claimReply(client.socket);
+            if (r.blockNum > commitNum)
+               break;
+            irreversibleTime = r.time;
+            if (r.blockNum < commitNum)
+            {
+               reversible.remove(r);
+               irreversible.push_back(r.blockNum);
```
swatanabe, line 254: This will delay the responses by one block. A block is irreversible if `r.blockNum <= commitNum`. The reason that the last irreversible block is not removed, is so there will always be at least one irreversible block to provide `irreversibleTime`.

services/system/Transact/src/RTransact.cpp
```diff
@@ -177,129 +179,327 @@ namespace
...
+         {
+            psio::get<0>(action->rawData().value()) = socket;
+            call(action.data(), action.size());
+         }
+      }
    }
+
+   void sendReplies(const std::vector<psibase::Checksum256>& txids)
+   {
+      stopTracking(txids);
```
swatanabe, line 318: If the node is killed after this, it can leave orphaned traces behind.

services/system/Transact/src/RTransact.cpp
swatanabe: Also, if you already have an object from `get`, `remove` is better than `erase`.

services/system/Transact/src/RTransact.cpp
James-Mart: Ah, I see. Thanks.
Fixed.

services/system/Transact/src/RTransact.cpp
James-Mart: Not sure how to handle this. I've moved the deletion to the end of `sendReplies`, but if I understand correctly, this is still a risk.

services/system/Transact/src/RTransact.cpp
swatanabe: What about this:
- send replies
- switch forks in a way the rolls back commitNum (Note: it isn't possible to fork out the block that is committed, but it *is* possible to switch to a fork that has a lower `commitNum` in the HEAD block)
- A new client pushes the same transaction with wait_for=applied

Now, we have both a successful trace and a client, but the conditions that caused the trace to be saved were based on the previous client, so the invariant that we're expecting doesn't hold.

services/system/Transact/src/RTransact.cpp
swatanabe: The order is now:
1. Remove trace
2. Remove client
3. Remove from queue

I think this is safe, as the queue entries drive cleanup. After a node restart, we don't care about the client and trace, except for the making sure that they get cleaned up.

services/system/Transact/src/RTransact.cpp
```diff
@@ -248,10 +248,15 @@ namespace
             if (r.blockNum > commitNum)
                break;
             irreversibleTime = r.time;
+
+            if (r.blockNum <= commitNum)
```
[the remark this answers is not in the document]
James-Mart: removed this condition

services/system/Transact/src/RTransact.cpp
```diff
@@ -177,129 +179,337 @@ namespace
...
       {
-         clients.remove(*row);
-         for (auto client : row->clients)
+         auto commitNum  = current.commitNum;
+         auto reversible = RTransact::WriteOnly{}.open<ReversibleBlocksTable>();
+         reversible.put({.blockNum = current.blockNum, .time = current.time});
+
+         BlockTime irreversibleTime = {};
+
+         std::vector<psibase::BlockNum> irreversible;
```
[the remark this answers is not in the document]
James-Mart: Now tracking first + count

services/system/Transact/src/RTransact.cpp
James-Mart: > I'm realizing that there are problems with pushing a transaction that has already been applied.

If you're not going to look into it right away, mind adding a github issue to track this?

swatanabe: In my scenario, the transaction is *not* executed again.

swatanabe: We're only forking back to a point before it was declared committed, not before it was executed.

swatanabe (approved):

PR #955, opened 2024-12-10
Updating to plugin_error macro

cool-ant:


services/user/Registry/plugin/src/errors.rs
```diff
@@ -1,37 +1,8 @@
...
-                producer: my_plugin_id(),
-                message: format!("Failed to post graphql query: {}", msg),
-            },
-            ErrorType::QueryDeserializeError => Error {
-                code: self as u32,
-                producer: my_plugin_id(),
-                message: format!("Failed to deserialize graphql query response: {}", msg),
-            },
-            ErrorType::NotFound => Error {
-                code: 404,
```
swatanabe, line 31: So that's where the 404 code was. `plugin_error!()` supports a discriminant specifically for this (`<variant> = <discriminant> => <fmt>`), but then I couldn't find it when I was doing the actual translation. I think Registry must still have been in PR at the time.

services/user/Registry/plugin/src/queries.rs
```diff
@@ -51,18 +52,18 @@ struct AppMetadataResponse {
 impl TryParseGqlResponse for AppMetadataResponseData {
     fn from_gql(response: String) -> Result<Self, CommonTypes::Error> {
         let response_root: ResponseRoot<AppMetadataResponse> = serde_json::from_str(&response)
-            .map_err(|e| QueryDeserializeError.err(&e.to_string()))?;
+            .map_err(|e| ErrorType::QueryDeserializeError(e.to_string()))?;
```
swatanabe, line 55: Isn't `QueryDeserializerError` in scope? It's not qualified anywhere else.

swatanabe (approved):

services/user/Registry/plugin/src/errors.rs
cool-ant: great. I'll look at the macro.

PR #957, opened 2024-12-11
Rework staged tx

James-Mart:
* make staged-tx a sudo service
* simplify auth service interface
* update staged tx service:
  * simplify interface
  * allow staging a single tx containing actions from multiple senders
  * use new auth service interface

### DIFF
--- a/services/system/AuthAny/src/AuthAny.cpp
+++ b/services/system/AuthAny/src/AuthAny.cpp
@@ -20,26 +20,29 @@ namespace SystemService
                               std::vector<Claim>         claims)
    {
       if (enable_print)
-         std::printf("auth_check\n");
+         std::printf("checkAuthSys\n");
    }
 
    void AuthAny::canAuthUserSys(AccountNumber user)
    {
       if (enable_print)
-         std::printf("user_check\n");
+         std::printf("canAuthUserSys\n");
    }
 
-   void AuthAny::stagedAccept(uint32_t stagedTxId, AccountNumber actor)
+   bool AuthAny::isAuthSys(AccountNumber sender, std::vector<AccountNumber> authorizers)
    {
-      check(getSender() == StagedTxService::service, "can only be called by staged-tx");
+      if (enable_print)
+         std::printf("isAuthSys\n");
 
-      auto [execute, allowedActions] = to<StagedTxService>().get_exec_info(stagedTxId);
-      recurse().to<Transact>().runAs(std::move(execute), allowedActions);
+      return true;
    }
 
-   void AuthAny::stagedReject(uint32_t stagedTxId, AccountNumber actor)
+   bool AuthAny::isRejectSys(AccountNumber sender, std::vector<AccountNumber> rejecters)
    {
-      check(false, "not supported");
+      if (enable_print)
+         std::printf("isRejectSys\n");
+
+      return true;
    }
 }  // namespace SystemService
 
--- a/services/system/AuthAny/src/AuthAny.cpp
+++ b/services/system/AuthAny/src/AuthAny.cpp
@@ -42,7 +42,7 @@ namespace SystemService
       if (enable_print)
          std::printf("isRejectSys\n");
 
-      return true;
+      return false;
    }
 }  // namespace SystemService
 
--- a/services/system/AuthSig/src/AuthSig.cpp
+++ b/services/system/AuthSig/src/AuthSig.cpp
@@ -87,13 +87,13 @@ namespace SystemService
       bool AuthSig::isAuthSys(psibase::AccountNumber              sender,
                               std::vector<psibase::AccountNumber> authorizers)
       {
-         return std::find(authorizers.begin(), authorizers.end(), sender) != authorizers.end();
+         return std::ranges::contains(authorizers, sender);
       }
 
       bool AuthSig::isRejectSys(psibase::AccountNumber              sender,
                                 std::vector<psibase::AccountNumber> rejecters)
       {
-         return std::find(rejecters.begin(), rejecters.end(), sender) != rejecters.end();
+         return std::ranges::contains(rejecters, sender);
       }
    }  // namespace AuthSig
 }  // namespace SystemService
--- a/services/system/AuthSig/src/AuthSig.cpp
+++ b/services/system/AuthSig/src/AuthSig.cpp
@@ -1,6 +1,7 @@
 #include <services/system/AuthSig.hpp>
 
 #include <psibase/dispatch.hpp>
+#include <services/system/Accounts.hpp>
 #include <services/system/Spki.hpp>
 #include <services/system/StagedTx.hpp>
 #include <services/system/VerifySig.hpp>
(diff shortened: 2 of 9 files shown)


services/system/AuthAny/src/AuthAny.cpp
```diff
@@ -20,26 +20,29 @@ namespace SystemService
...
    }
 
-   void AuthAny::stagedReject(uint32_t stagedTxId, AccountNumber actor)
+   bool AuthAny::isRejectSys(AccountNumber sender, std::vector<AccountNumber> rejecters)
    {
-      check(false, "not supported");
+      if (enable_print)
+         std::printf("isRejectSys\n");
+
+      return true;
```
swatanabe, line 45: Is this the right behavior? My understanding of the expected sematics of reject is "return true iff no set of accepts that does not conflict with the rejects will be sufficient to authorize"

services/system/AuthSig/src/AuthSig.cpp
```diff
@@ -83,32 +84,24 @@ namespace SystemService
...
       }
 
-      void AuthSig::stagedAccept(uint32_t staged_tx_id, psibase::AccountNumber actor)
+      bool AuthSig::isAuthSys(psibase::AccountNumber              sender,
+                              std::vector<psibase::AccountNumber> authorizers)
       {
-         check(getSender() == StagedTxService::service, "can only be called by staged-tx");
+         auto accDetails = to<Accounts>().getAccount(sender);
+         check(accDetails.has_value() && accDetails->authService == AuthSig::service,
+               "sender " + sender.str() + " does not use this auth service");
```
swatanabe, line 92: Is this check needed? `checkAuthSys` doesn't have it. If it doesn't hold, it's the caller's fault and the only harmful consequence is that the result is useless to the caller. I don't see it as being likely to happen by accident either, as the caller probably got this service in the first place by looking it up in accounts.

services/system/AuthSig/src/AuthSig.cpp
```diff
@@ -83,32 +84,24 @@ namespace SystemService
...
-         auto staged_tx = to<StagedTxService>().get_staged_tx(staged_tx_id);
-         auto actions   = staged_tx.action_list.actions;
-
-         check(actions.size() > 0, "staged tx has no actions");
-         if (actor == actions[0].sender)
-         {
-            auto [execute, allowedActions] = to<StagedTxService>().get_exec_info(staged_tx_id);
-            recurse().to<Transact>().runAs(std::move(execute), allowedActions);
-         }
+         return std::find(authorizers.begin(), authorizers.end(), sender) != authorizers.end();
```
swatanabe, line 94: std::ranges::contains

services/system/StagedTx/service/src/lib.rs
```diff
@@ -313,81 +314,57 @@ pub mod service {
...
+        staged_tx.reject();
 
-        staged_tx.emit(StagedTxEvent::REJECTED);
+        emit_update(staged_tx.txid, StagedTxEvent::REJECTED);
 
-        staged_tx
-            .staged_tx_policy()
-            .reject(staged_tx.id, get_sender());
+        let rejected =
+            staged_tx.action_list.actions.iter().all(|action| {
```
swatanabe, line 322: Should be `any` instead of `all`.

services/system/AuthAny/src/AuthAny.cpp
James-Mart: Hmm, I don't think accepts ever need to be considered during reject. We know the accept criteria wasn't met, otherwise it would already be accepted and this wouldn't be called. So we just need to check if a reject threshold was met. 

In `auth-any` specifically, this would only be called via indirection (when an auth-any account is on an msig for example), so the intention of returning true here is to imply that `authAny` always agrees with what everyone else wants.

swatanabe: Here's an example:

A transaction contains two actions:
action1 is authorized by msig 2/3 alice, bob, carol
action2 is authorized by auth-any

carol rejects the transaction

swatanabe: My point about accept was what I expect the relationship between the accept and reject thresholds to be.

reject(S, A) = ¬accept(S, A')

services/system/AuthSig/src/AuthSig.cpp
James-Mart: >  the only harmful consequence is that the result is useless to the caller

Without this check, the call will silently seem to succeed, and the auth service might even reply `true`, causing the caller's bug to not only result in a failure, but in an actual erroneous execution of the staged tx. This could happen if the account previously used this auth service so it actually has a configuration.

So... it's not strictly necessary but I was thinking the cost of this check is worthwhile even though I agree it's an unlikely bug.

services/system/StagedTx/service/src/lib.rs
James-Mart: Oh, good catch!

services/system/AuthSig/src/AuthSig.cpp
swatanabe: The auth service is always looked up right before calling it, so I don't see how a previous auth service matters.

James-Mart: > The auth service is always looked up right before calling it, 

The bug in the hypothetical disagrees with this by definition, otherwise there wouldn't be a bug. It's the caller's fault, I agree, but I was just considering how to minimize damage in this edge case. Perhaps it's just confusing, though. I'll remove it.

swatanabe: The actual usage always does a lookup and immediate call, and that's by far the easiest and most natural way to use it. It's just a lot of extra code for such an unlikely error.

James-Mart: removed

services/system/AuthSig/src/AuthSig.cpp
James-Mart: done

services/system/StagedTx/service/src/lib.rs
James-Mart: done

services/system/AuthAny/src/AuthAny.cpp
James-Mart: >  what I expect the relationship between the accept and reject thresholds to be

The thresholds can be whatever... and can be user-defined, for example an msig may require super-majority to accept, simple-majority to reject

swatanabe: I don't see reject as an independent operation. It is the opposite of accept, that is, it indicates that the rejector will not accept the transaction. Deleting the transaction is just a convenience that should happen when the authorization becomes unsatisfiable under that assumption.

swatanabe: Being able to satisfy both the accept and reject thresholds simultaneously is just confusing. It's ambiguous and the actual result depends on the order in which they are submitted, which creates a race.

swatanabe: > Didn't address this yet because I still don't understand.

Let's say that in addition to carol rejecting, alice and bob accept. There are two things that I find weird:
- Whether it is accepted or rejected depends on whether alice and bob are quick enough to beat carol.
- Adding an action unrelated to carol caused carol's reject to reject the entire transaction.

James-Mart: > There are two things that I find weird:

Okay, I agree with both of these things. I find them unimportant because this is a problem unique to people who are proposing transactions that use an auth-any account, which shouldn't really be a thing.

Should I just make auth-any always return false from reject?

swatanabe: > Should I just make auth-any always return false from reject?

I think so.

James-Mart: Done

swatanabe (approved):

PR #962, opened 2024-12-14
Update Chain mail to use new accounts system

brandonfancher:
Chain mail now redirects user to accounts app to sign in and manage accounts.


James-Mart:
Not blocking the PR over these, but I get these scrollbars on chrome

Sent/Received message view:
image

Compose view:
image

Velua (approved):
Works well!

brandonfancher:
@James-Mart Thanks for pointing that out. Must be a Windows things. I'll look into it.

PR #972, opened 2024-12-20
Transfer snapshots over p2p when the block log is truncated

swatanabe:
- [x] Ensure that snapshots during joint consensus include all headers from the start of joint consensus, so that chained snapshots work correctly.
- [x] Error handling during the the snapshot sending flow
- [x] Correct management of connection lifetime while sending a snapshot
- [x] After receiving a snapshot, update the connection state of all peers to account for the discontinous jump.
- [x] Clean up logging/debug prints


libraries/psibase/common/include/psibase/nativeTables.hpp
```diff
@@ -20,6 +20,7 @@ namespace psibase
    static constexpr NativeTableNum consensusChangeTable       = 10;  // subjective
    static constexpr NativeTableNum snapshotTable              = 11;  // subjective
    static constexpr NativeTableNum scheduledSnapshotTable     = 12;  // both
+   static constexpr NativeTableNum logTruncateTable           = 13;
```
James-Mart, line 23: Table is missing from `native_tables.rs`

James-Mart (approved):

PR #973, opened 2025-01-03
Add key imports to accounts app

Velua:
Adds `auth-sig` UI which enables exporting a private key for a certain account, `accounts` can handle importing it with a `?key=` query param.


James-Mart:
CI/CD failing:
```
make[2]: *** No rule to make target '../services/system/AuthSig/ui/dist/index.html', needed by 'share/psibase/packages/AuthSig.psi'.  Stop.
make[1]: *** [CMakeFiles/Makefile2:1731: CMakeFiles/AuthSig.dir/all] Error 2
make[1]: *** Waiting for unfinished jobs....
```

CMakeLists.txt
```diff
@@ -445,11 +445,13 @@ psibase_package(
         WASM ${CMAKE_CURRENT_BINARY_DIR}/AuthSig.wasm
         DATA ${Plugins_OUTPUT_FILE_auth_sig} /plugin.wasm
         SERVER r-auth-sig
+        DATA GLOB ${CMAKE_CURRENT_SOURCE_DIR}/services/system/AuthSig/ui/dist/* /
     SERVICE r-auth-sig
         WASM ${CMAKE_CURRENT_BINARY_DIR}/RAuthSig.wasm
     SERVICE verify-sig
         WASM ${CMAKE_CURRENT_BINARY_DIR}/VerifySig.wasm
         FLAGS isAuthService
+    DEPENDS ${AuthSig_js_DEP} ${CMAKE_CURRENT_SOURCE_DIR}/services/system/AuthSig/ui/dist/index.html
```
James-Mart, line 454: Don't need to add this index.html file if you're already adding `AuthSig_js_DEP`. Although given the CI/CD error, I think the `AuthSig_js` target has yet to be added.

James-Mart:
You need the dependency on the `AuthSig_js_DEP`. My comment was that you don't *also* need an explicit dependency on the `index.html` file.

And you also need to add the AuthSig_js target itself (so that theres a `*_DEP` for the package to depend on).

James-Mart (approved):

James-Mart:
Actually, when I click import, I get no feedback. The import actually worked (I can see the plugin call was successful if I look at the console logs in devtools).

image

I think it should have an "Account import successful! <a> Click here<a/> to return to homepage" modal.

services/system/AuthSig/ui/src/components/nav.tsx
```diff
@@ -0,0 +1,27 @@
...
+
+function HoverBorderGradientDemo() {
+  return (
+    <div className=" flex justify-center text-center">
+      <HoverBorderGradient
+        as="div"
+        containerClassName="rounded-full"
+        className="flex items-center space-x-2 bg-white text-black dark:bg-black dark:text-white"
+      >
+        <span>psibase</span>
```
James-Mart, line 12: Network name should get taken from the branding plugin

James-Mart, line 12: Network name should get taken from the branding plugin

James-Mart (changes requested):

James-Mart:
Now I don't see any account to import:
image

No errors in the console

services/system/Accounts/ui/src/AccountSelection.tsx
```diff
@@ -550,7 +525,7 @@ export const AccountSelection = () => {
                   ? "Loading..."
                   : isInvite
                   ? `Select an account to accept invite to `
-                  : `Select an account to login to `}
+                  : `Select an account to login to  `}
```
brandonfancher, line 528: nit, but if I'm not mistaken, the verb form of login is "log in," so there should be a space.

services/system/AuthSig/ui/src/components/theme-provider.tsx
brandonfancher, line 1: I notice we don't have a dark/light mode switch control... just an observation

brandonfancher (approved):
Tested and works nicely. Approving so as to not block merge. But I recommend addressing my (minor) comments.

James-Mart (approved):

Issue #965, opened 2024-12-17
auth sig account with key lookup failure


Velua:
When I attempt a GraphQL query against `auth-sig` I get an error `service 'r-auth-sig' aborted with message: N5Botan16Invalid_ArgumentE`

```
{
	accWithKey(pubkeyPem: "-----BEGIN PUBLIC KEY-----MFkwEwYHKoZIzj0CAQYIKoZIzj0DAQcDQgAEOOYtr14pQjE4fnZ/RAsqEqcmsnYH\nNdXjUz0Y96encrgnStgrc6fXq7M5Xkxvyus+jK8y9ZwTDyI284zK7Lx0tA==\n-----END PUBLIC KEY-----\n") {
		edges {
			node {
				account
			}
		}
 
	}
}
```

swatanabe:
I suspect a missing `\n` after `-----BEGIN PUBLIC KEY-----`

PR #975, opened 2025-01-04
Improve auth-sig query by key

James-Mart:
This seems to work, should close https://github.com/gofractally/psibase/issues/965

### DIFF
--- a/services/system/AuthSig/src/RAuthSig.cpp
+++ b/services/system/AuthSig/src/RAuthSig.cpp
@@ -31,7 +31,11 @@ namespace SystemService
                          optional<string>        before,
                          optional<string>        after) const
          {
-            std::erase(pubkeyPem, '\n');
+            std::size_t pos;
+            while ((pos = pubkeyPem.find("\\n")) != std::string::npos)
+            {
+               pubkeyPem.replace(pos, 2, "\n");
+            }
 
             auto idx =
                 AuthSig::Tables{AuthSig::service}.open<AuthSig::AuthTable>().getIndex<1>().subindex(
--- a/services/system/AuthSig/src/RAuthSig.cpp
+++ b/services/system/AuthSig/src/RAuthSig.cpp
@@ -31,6 +31,8 @@ namespace SystemService
                          optional<string>        before,
                          optional<string>        after) const
          {
+            std::erase(pubkeyPem, '\n');
+
             auto idx =
                 AuthSig::Tables{AuthSig::service}.open<AuthSig::AuthTable>().getIndex<1>().subindex(
                     keyFingerprint(SubjectPublicKeyInfo{parseSubjectPublicKeyInfo(pubkeyPem)}));


services/system/AuthSig/src/RAuthSig.cpp
```diff
@@ -31,6 +31,8 @@ namespace SystemService
                          optional<string>        before,
                          optional<string>        after) const
          {
+            std::erase(pubkeyPem, '\n');
```
swatanabe, line 34: I'm actually somewhat surprised this works. The new line (CR, LF, or CRLF) is [required](https://www.rfc-editor.org/rfc/rfc7468#page-5) for PEM, though many parsers are somewhat forgiving. The reason it failed, was that the query was not correct.

James-Mart: Changed my approach

James-Mart:
One should be able to just take the value of the returned public key from the `account` query and use it to find all related accounts using `accWithKey`. The issue is that the returned public key looks like:
```
{
    "data": {
        "account": {
            "account": "alice",
            "pubkey": "-----BEGIN PUBLIC KEY-----\nMFkwEwYHKoZIzj0CAQYIKoZIzj0DAQcDQgAE6u0G6+EAjvtXxQ7n5pPD0iPRzaXh\nZMeGCmkRj4kxfst5x6O1iDsW8IOaXrUyzfcUVIsRT0LpPfpJ1iGsPpZExA==\n-----END PUBLIC KEY-----\n"
        }
    }
}
```

And the newlines are escaped during serialization, which results in literal "\n" sequences that cause key parse failures. The new approach replaces any literal newline character sequences "\n" with an actual newline character before trying to parse the key.

swatanabe:
> One should be able to just take the value of the returned public key from the `account` query and use it to find all related accounts using `accWithKey`. The issue is that the returned public key looks like:
> 
> ```
> {
>     "data": {
>         "account": {
>             "account": "alice",
>             "pubkey": "-----BEGIN PUBLIC KEY-----\nMFkwEwYHKoZIzj0CAQYIKoZIzj0DAQcDQgAE6u0G6+EAjvtXxQ7n5pPD0iPRzaXh\nZMeGCmkRj4kxfst5x6O1iDsW8IOaXrUyzfcUVIsRT0LpPfpJ1iGsPpZExA==\n-----END PUBLIC KEY-----\n"
>         }
>     }
> }
> ```
> 
> And the newlines can be easily escaped during serialization, which results in literal "\n" sequences that cause key parse failures. The new approach replaces any literal newline character sequences "\n" with an actual newline character before trying to parse the key.

The GraphQL result is JSON. Why aren't the escapes processed during deserialization?

swatanabe:
Haha:
```
                  // * Block strings (""") not currently supported and escape processing not
                  //   currently done; we may have to revisit this if we add either mutation
```

swatanabe:
This should be fixed properly in the GraphQL parser. See the [specification](https://spec.graphql.org/October2021/#sec-String-Value).

PR #976, opened 2025-01-04
Fix cpp graphql schema generation including variants

### LINKED ISSUE #966: producers gql invalid schema
`ConsensusData` is printed in the schema twice for some reason

```
type Claim {
    service: String!
    rawData: [Int!]!
}
type Producer {
    name: String!
    auth: Claim!
}
type CftConsensus {
    producers: [Producer!]!
}
type BftConsensus {
    producers: [Producer!]!
}
union ConsensusData =
  | CftConsensus
  | BftConsensus
union ConsensusData =
  | CftConsensus
  | BftConsensus
type Query {
    producers: [Producer!]!
    nextProducers: [Producer!]!
    consensus: ConsensusData!
    nextConsensus: ConsensusData
    jointStart: Float
}
```

James-Mart:
Closes #966

### DIFF
--- a/libraries/psio/include/psio/graphql.hpp
+++ b/libraries/psio/include/psio/graphql.hpp
@@ -359,13 +359,16 @@ namespace psio
                              bool                                        is_query_root)
    {
       (fill_gql_schema((T*)nullptr, stream, defined_types, is_input, false), ...);
-      write_str("union ", stream);
-      write_str(generate_gql_partial_name((std::variant<T...>*)nullptr, is_input), stream);
-      write_str(" =", stream);
-      ((write_str("\n  | ", stream),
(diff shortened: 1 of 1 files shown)


libraries/psio/include/psio/graphql.hpp
```diff
@@ -359,13 +359,16 @@ namespace psio
                              bool                                        is_query_root)
    {
       (fill_gql_schema((T*)nullptr, stream, defined_types, is_input, false), ...);
```
swatanabe, line 361: This should probably be inside the condition.  It could cause infinite recursion in some corner cases.

James-Mart: Fixed

swatanabe (approved):

PR #978, opened 2025-01-06
Require table definitions outside of service module

cool-ant:
- moved processing of tables in service macro out into new macro (that also is applied to modules), named `service_tables`
- All table definitions must now be made in such a module, devoted to tables
- All table-related functionality persists in the new macro
- Clients simply need to add a `use` line to pull in the table definitions from the `service_tables` module.


James-Mart:
You should update 
`doc/psidk/src/development/services/rust-service/tables.md`
and the service that uses a table in 
`doc/psidk/src/development/services/rust-service/graphql.md`

cool-ant:
> doc/psidk/src/development/services/rust-service/graphql.md

right. Had this on my list and forgot it. Thanks.
NOTE: there's so much code in here that, for maintainability, we might consider moving all this code out to actual code files in a dedicated directory so we can run tests to confirm this code all compiles...

swatanabe (approved):

PR #982, opened 2025-01-09
Create and import server key on prod chain boot

brandonfancher:
## In this PR
When booting in Production mode, generate bp block and transaction signing key. Set these keys in server (config file as HSM for now) and in plugins. Import key and account to accounts plugin.

See commit messages for more granularity.

## Directive for this PR
1. Start `psinode` (no need for any `—key` flag because that’s part of the `psibase` boot interface, which is an alternative to using `psinode` and `XAdmin`)
2. Open up `XAdmin` interface
3. Choose `Production`
4. Generate a key (can be done invisibly, or with minor feedback)
5. **Not in this PR...fast-follow:** List HSMs on server. Choose the one you want to use and unlock it with your HSM PIN. “This is the device that will store your block-signing key.” 
   1. `GET` on `/native/admin/keys/devices`
   2. `POST` to `/native/admin/keys/unlock` (handle error that returns if incorrect PIN entered) Hex encoding of a `DER` (binary) encoded key. `R1`
   3. Context: I can call `softhsm2-util` to list the virtual HSMs in `psibase-contributor`.
6. When you click the final button in the wizard but before boot, it should call that API endpoint to import the key that was generated behind the scenes. `POST` to `/native/admin/keys`
7. Boot with public key
8. After boot, call `AuthSig` plugin function `importKey` with that same key. That will load the key into the plugin. Then you can sign with producer.
   1. `/root/psibase/services/system/AuthSig/plugin/wit/world.wit`
   2. Import function: `keyvault/import-key`

## Fast follow PR
- Add indication that Production is secure and generates keys (see comment below)
- List HSMs on server for user to choose from and unlock with PIN.


services/user/XAdmin/ui/src/pages/create-page.tsx
```diff
@@ -156,47 +167,64 @@ export const CreatePage = () => {
...
+                }
+                const desiredPackageIds = Object.keys(rows);
+                const desiredPackages = packages.filter((pack) =>
+                    desiredPackageIds.some((id) => id == getId(pack))
+                );
+                const requiredPackages = getRequiredPackages(
+                    packages,
+                    desiredPackages.map((pack) => pack.name)
+                );
+                bootChain(requiredPackages, bpName, (state) => {
```
swatanabe, line 184: The server's public key needs to be passed to boot.

brandonfancher: Ah, thank you

rust/psibase/src/boot.rs
```diff
@@ -276,10 +281,21 @@ pub fn js_create_boot_transactions(
     let expiration = TimePointSec::from(now_plus_120secs);
     let prod = js_err(ExactAccountNumber::from_str(&producer))?;
 
-    // Todo: Allow parameterization of compression level from browser
-    let compression_level = 4;
+    let initial_key = public_key_pem
+        .map(|k| -> Result<AnyPublicKey, JsValue> {
+            let data = pem::parse(k.trim()).map_err(|e| JsValue::from_str(&e.to_string()))?;
```
swatanabe, line 286: Should check that `data.tag()` is `"PUBLIC KEY"` and that the contents can be parsed as a `SubjectPublicKeyInfo`.

James-Mart: done

brandonfancher: James addressed this.

swatanabe:
```
[2025-01-14T21:36:13.531Z] [notice]: Node is not connected to any psibase networ
k. Visit 'http://x-admin.psibase.127.0.0.1.sslip.io:8080' for node setup.
```
This advice no longer works:
```Uncaught (in promise) ReferenceError: CryptoKey is not defined```

It's probably fine for production to mandate HTTPS, but dev should work with plain HTTP. I just get a blank page and had to open the web developer tools to figure out what went wrong.

**Comment by brandonfancher** [comment]
Yes, that is a limitation of generating the keys with inbuilt browser methods.

Are we okay to assume that users selecting `Production` should always boot in a secure environment? If so, I can make this work fine via plain HTTP when using the UI to boot a `Development` chain. And I can disable the `Production` option if we're not in HTTPS...with some indication to the user that HTTPS is required for booting `Production` via the UI.

Otherwise, I may need to bring in third-party libraries for key generation.

@swatanabe @James-Mart thoughts?

> ```
> [2025-01-14T21:36:13.531Z] [notice]: Node is not connected to any psibase networ
> k. Visit 'http://x-admin.psibase.127.0.0.1.sslip.io:8080' for node setup.
> ```
> 
> This advice no longer works: `Uncaught (in promise) ReferenceError: CryptoKey is not defined`
> 
> It's probably fine for production to mandate HTTPS, but dev should work with plain HTTP. I just get a blank page and had to open the web developer tools to figure out what went wrong.

**Comment by James-Mart** [comment]
> Are we okay to assume that users selecting Production should always boot in a secure environment?

We could easily transpile the AuthSig plugin and bundle it with X-Admin, which would allow you to generate keys in browser in HTTP context. However, I actually think it's a good idea to restrict the production node to HTTPS. So, I agree with the plan to just  disable the selection of the Production template with some kind of error message for the user.

**Comment by brandonfancher** [comment]
> > Are we okay to assume that users selecting Production should always boot in a secure environment?
> 
> We could easily transpile the AuthSig plugin and bundle it with X-Admin, which would allow you to generate keys in browser in HTTP context. However, I actually think it's a good idea to restrict the production node to HTTPS. So, I agree with the plan to just disable the selection of the Production template with some kind of error message for the user.

Sounds good. I'll make the adjustments.

**Comment by James-Mart** [comment]
In the fast follow, I think you should include some indication to the admin that a key is being generated for their account and imported into the browser on their current device.
-----------

image

Something like:

Block Producer Account: `myproducer`
Account type: `Secure` | `Insecure` (?) 

Hovering over / clicking the (?): "Creating a network with the Production template automatically generates a keypair for your account. The account is automatically imported into the accounts app in the current device's web browser."

File: services/user/XAdmin/ui/src/lib/bootChain.ts [hunk 3]
```diff
@@ -25,11 +28,26 @@ export const bootChain = async (
...
         // Something is wrong with the Vite proxy configuration that causes boot to intermittently (but often) fail
         // in a dev environment.
 
         const [boot_transaction, transactions] =
-            wasm.js_create_boot_transactions(producerName, fetchedPackages);
+            wasm.js_create_boot_transactions(
+                producerName,
+                fetchedPackages,
+                publicKeyPem,
+                4
```
James-Mart, line 49: I think I mentioned you could just hardcode the compression as 4, but now that I think about it, it's probably better to use 11 for the production template, 4 for the development template.

services/user/XAdmin/ui/src/hooks/useImportKey.ts
```diff
@@ -28,13 +28,11 @@ export const useImportKey = () =>
...
 
-            // TODO: This function call does not properly import the account into the accounts app
-            // TODO: When we add the new method here, ensure it also fires for Development boots
-            // void (await supervisor.functionCall({
-            //     method: "importAccount",
-            //     params: [account],
-            //     service: "accounts",
-            //     intf: "admin",
-            // }));
+            void (await supervisor.functionCall({
```
James-Mart: > When we add the new method here, ensure it also fires for Development boots

Was this confirmed?

services/user/XAdmin/ui/src/lib/bootChain.ts
brandonfancher: Okay. I'll address this.

brandonfancher: Are we going with 11 because of the presence of the public key, or specifically because it's Production? Not sure what "compression" is here.

brandonfancher: 7 seemed to strike the right balance with UI responsiveness. FYI.

James-Mart (approved):

PR #984, opened 2025-01-13
Add Workshop package

Velua:
This adds the `Workshop.psi` package with it's own UI which engages the registry. 

Currently, whatever account the user is authenticated as is assumed to be the account of the actual app, after staged transactions is ready we'll later adopt it for a better multi-user experience. 

The default header in sites for `Content-Security-Policy` in images has been updated to include `:data` so that image embedding from a string will work, subsequent changes will remove the idea of default headers and instead will be entirely configurable in the workshop app for developers to work with.


services/user/Workshop/ui/README.md
James-Mart, line 1: Any reason not to delete this?

services/user/Workshop/ui/src/components/metadata-form.tsx
```diff
@@ -0,0 +1,271 @@
...
+        className="flex flex-col gap-2"
+      >
+        <FormField
+          control={form.control}
+          name="name"
+          render={({ field }) => (
+            <FormItem>
+              <FormLabel>App Name</FormLabel>
+              <FormControl>
+                <Input placeholder="MonsterEOS" {...field} />
```
James-Mart, line 134: No one knows what MonsterEOS is, and it's a reference to EOS.
Probably should pick a different placeholder. `MonsterApp`?

services/user/Workshop/ui/src/components/nav.tsx
```diff
@@ -0,0 +1,27 @@
...
+
+function HoverBorderGradientDemo() {
+  return (
+    <div className=" flex justify-center text-center">
+      <HoverBorderGradient
+        as="div"
+        containerClassName="rounded-full"
+        className="flex items-center space-x-2 bg-white text-black dark:bg-black dark:text-white"
+      >
+        <span>psibase</span>
```
James-Mart, line 12: Should take name from branding app

services/user/Workshop/ui/vite.config.ts
James-Mart, line 1: Is this being used anymore?

Velua: Removed some stale config settings, me and @brandonfancher will go over the rest and figure out what makes sense and update all Vite projects.

brandonfancher: Yeah. I think we're going to need to do the opposite of what we've been doing. Instead of proxying from the vite dev server, we may want to proxy from the psibase config to the dev server for certain subdomains. Otherwise, our same-origin cross-subdomain OAuth-y stuff isn't going to work when trying to iterate quickly via the dev server, as it runs on another port.

services/user/Workshop/ui/public/vite.svg
brandonfancher, line 1: Let's remove this file and is reference in `index.html`

services/user/Workshop/ui/src/App.css
brandonfancher, line 1: This is boilerplate that can be deleted, right?

services/user/Workshop/ui/src/components/metadata-form.tsx
```diff
@@ -0,0 +1,271 @@
...
+        className="flex flex-col gap-2"
+      >
+        <FormField
+          control={form.control}
+          name="name"
+          render={({ field }) => (
+            <FormItem>
+              <FormLabel>App Name</FormLabel>
+              <FormControl>
+                <Input placeholder="MonsterApp" {...field} />
```
brandonfancher, line 134: I'd recommend a different default example here instead of a game. Maybe something related to voting or trading. Or a prediction market of some kind. 🤔

services/user/Workshop/ui/src/components/metadata-form.tsx
```diff
@@ -0,0 +1,271 @@
...
+        />
+        <div className="md:grid grid-cols-3 gap-2">
+          <FormField
+            control={form.control}
+            name="tosSubpage"
+            render={({ field }) => (
+              <FormItem>
+                <FormLabel>Terms of service</FormLabel>
+                <FormControl>
+                  <Input placeholder="/terms-of-service" {...field} />
```
brandonfancher, line 226: I was a little confused by the placeholder in these three fields. My initial intuition was that they were defaults that would actually get submitted if I didn't overwrite them. Instead, `/` was submitted for them. 

I could see a couple ways of getting around that:

* Make them required so the user has to explicitly input values, or
* Make the placeholder value submit as the default value, unless overridden by the user

services/user/Workshop/ui/src/components/theme-provider.tsx
brandonfancher, line 1: This reminds me...should there be a dark/light toggle?

services/user/Workshop/ui/src/components/ui/avatar.tsx
brandonfancher, line 1: Unused

services/user/Workshop/ui/src/components/ui/dialog.tsx
brandonfancher, line 1: Unused

services/user/Workshop/ui/src/components/ui/skeleton.tsx
brandonfancher, line 1: Unused

brandonfancher (changes requested):

services/user/Workshop/ui/src/components/metadata-form.tsx
Velua: Now explicit values are required.

services/user/Workshop/ui/src/components/theme-provider.tsx
Velua: Added

brandonfancher (approved):

PR #990, opened 2025-01-15
Fix registry package configuration

James-Mart:
Uses the conventions established in package-templates.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -427,7 +427,7 @@ cargo_psibase_package(
 )
 
 cargo_psibase_package(
-    OUTPUT ${SERVICE_DIR}/AppRegistry.psi
+    OUTPUT ${SERVICE_DIR}/Registry.psi
     PATH services/user/Registry
 )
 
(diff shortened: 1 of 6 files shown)


James-Mart:
CC @swatanabe I had a CI/CD error about the size of `DevDefault` being too large for `cargo-psibase` test, so I made a `TestDefault` package with a subset of packages. Should grow much slower than `DevDefault`.

swatanabe:
> CC @swatanabe I had a CI/CD error about the size of `DevDefault` being too large for `cargo-psibase` test, so I made a `TestDefault` package with a subset of packages. Should grow much slower than `DevDefault`.

The reason it causes a problem is that the rust tester doesn't call start_block during boot, and so tries to pack everything into one block.

James-Mart:
>  the rust tester doesn't call start_block during boot

Ah, I see. Though I like my solution regardless, yeah? Don't need all those DevDefault packages in every test chain.

Velua (approved):

PR #991, opened 2025-01-15
Align graphql and json timepointusec serialization

James-Mart:
Otherwise graphql returns a timepointusec in a form that serde is unable to deserialize into a rich object

### DIFF
--- a/rust/psibase/src/time.rs
+++ b/rust/psibase/src/time.rs
@@ -1,6 +1,5 @@
 use crate::{Pack, ToKey, ToSchema, Unpack};
-use async_graphql::ScalarType;
-use async_graphql::{InputObject, SimpleObject};
+use async_graphql::{scalar, InputObject, SimpleObject};
 use chrono::{DateTime, FixedOffset, Utc};
 use serde::{
     de::{Deserializer, Error as _},
@@ -8,6 +7,7 @@ use serde::{
     Deserialize, Serialize,
 };
 use std::ops::{Add, Sub};
+use std::str::FromStr;
 
 #[derive(
     Debug,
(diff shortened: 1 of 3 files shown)


rust/psibase/src/time.rs
```diff
@@ -72,32 +73,48 @@ impl<'de> Deserialize<'de> for TimePointSec {
...
     fracpack_mod = "fracpack",
     custom = "TimePointUSec"
 )]
 #[to_key(psibase_mod = "crate")]
-#[graphql(input_name = "TimePointUSecInput")]
 pub struct TimePointUSec {
     pub microseconds: i64,
 }
 
+impl TryFrom<String> for TimePointUSec {
```
swatanabe, line 88: Shouldn't this be `FromStr`?

rust/psibase/src/time.rs
```diff
@@ -72,32 +73,48 @@ impl<'de> Deserialize<'de> for TimePointSec {
...
+}
+
+impl ToString for TimePointUSec {
+    fn to_string(&self) -> String {
+        DateTime::<Utc>::from(*self).to_rfc3339()
+    }
+}
+
+#[async_graphql::Scalar]
+impl ScalarType for TimePointUSec {
```
swatanabe, line 104: Does `scalar!` not work? It purports to use `Serialize` and `Deserialize` which should guarantee consistency.

rust/psibase/src/time.rs
James-Mart: It's not quite clear to me why it should be one over the other. I guess `FromStr` is the more specific tool for the job...

James-Mart: Oh, for std::str::parse compatibility?

rust/psibase/src/time.rs
James-Mart: Oh, nice. Yeah that works great. Done

rust/psibase/src/time.rs
James-Mart: Done

swatanabe: I would use `From<String>` only for types that adopt the string directly into their state. `FromStr` is better when the string is parsed.

rust/psibase/src/time.rs
```diff
@@ -128,17 +132,14 @@ impl From<TimePointUSec> for DateTime<Utc> {
...
 }
 
 impl<'de> Deserialize<'de> for TimePointUSec {
     fn deserialize<D: Deserializer<'de>>(deserializer: D) -> Result<Self, D::Error> {
-        let s = <&str>::deserialize(deserializer)?;
-        Ok(DateTime::<FixedOffset>::parse_from_rfc3339(s)
-            .map_err(|e| D::Error::custom(e.to_string()))?
-            .to_utc()
-            .into())
+        let s = <String>::deserialize(deserializer)?;
```
swatanabe, line 141: There's no reason to make a copy.

James-Mart: fixed

swatanabe (approved):

PR #994, opened 2025-01-16
Use localhost instead of sslip

swatanabe:
localhost is reserved as a top level domain and browsers consider it and its subdomains secure, even over HTTP.

- To allow localhost:8080 to reach boot, all requests redirect to x-admin when the chain is not booted.
- To avoid a disruptive transition, I added the ability to have multiple host names.

TODO:
- [x] Config read needs to parse changes to config format (hosts instead of host)
- [x] Config editor needs to allow multiple hosts
- [x] psibase should not depend on DNS to resolve *.localhost

### DIFF
--- a/libraries/psibase_http/handle_request.cpp
+++ b/libraries/psibase_http/handle_request.cpp
@@ -48,6 +48,18 @@ namespace psibase::http
       return origin;
    }
 
+   std::pair<beast::string_view, beast::string_view> split_port(beast::string_view host)
+   {
+      if (auto pos = host.rfind(':'); pos != std::string::npos)
+      {
+         if (host.find(']', pos) == std::string::npos)
+         {
+            return {host.substr(0, pos), host.substr(pos)};
+         }
+      }
+      return {host, {}};
+   }
+
    beast::string_view deport(beast::string_view host)
    {
       if (auto pos = host.rfind(':'); pos != std::string::npos)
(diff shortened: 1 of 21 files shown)


services/user/XAdmin/ui/src/configuration/utils.ts
```diff
@@ -276,7 +273,7 @@ export const mergeConfig = (
         ...updated,
         p2p: mergeSimple(prev.p2p, updated.p2p, user.p2p),
         producer: mergeSimple(prev.producer, updated.producer, user.producer),
-        host: mergeSimple(prev.host, updated.host, user.host),
+        hosts: mergeSimple(prev.hosts, updated.hosts, user.hosts),
```
swatanabe, line 276: Should use mergeList

brandonfancher: I think a little more work will be needed there. `mergeArray` expects a list of objects.

brandonfancher: `mergeSimple` got the boot unstuck. But I believe I'm going to need to turn `hosts` into a list of objects anyway...for working with them in the configuration form. Still familiarizing myself with this part of the x-admin code base though.

brandonfancher (approved):
Working well for me

libraries/psibase_http/handle_request.cpp
```diff
@@ -695,8 +758,37 @@ namespace psibase::http
...
             if (bc.needGenesisAction)
-               return send(error(bhttp::status::internal_server_error,
-                                 "Node is not connected to any psibase network."));
+            {
+               std::string location;
+               if (send.is_secure())
+                  location += "https://";
+               else
+                  location += "http://";
+               l.lock();
```
James-Mart, line 767: Is RAII worth using here?

James-Mart (approved):

libraries/psibase_http/handle_request.cpp
swatanabe: It is RAII. `l` is a unique_lock.

PR #1011, opened 2025-01-27
avoid quoting all numbers in events in json output

James-Mart:
SQLITE appears only able to give us numbers as 64 bit, but `psio::to_json` quotes numbers >32 bits. This disrupts parsing json back into numbers when it was serialized from event data. This so this tries to downcast event data numbers before serializing if possible.


swatanabe:
We really need to decide whether a string is a valid representation of a number (and fix rust) or not (and fix javascript).

James-Mart:
Closing in favor of explicitly instructing serde for string->number deserialization, when needed. 
As seen in 
https://github.com/gofractally/psibase/pull/1013/commits/ba16fd9889dbe38c1c855a99d9fb08323121ad05

PR #1013, opened 2025-01-27
Simplify making graphql-pagination-style queries for event data

### LINKED ISSUE #901: Graphql support for events
There ought to be a way to easily query events that returns a graphql result. If it's simpler, the query itself could still be made using SQL syntax, but we would need to be able to map the response into something graphql compatible...

Thoughts?

James-Mart:
Closes #901

### DIFF
--- a/rust/psibase/src/graph_ql.rs
+++ b/rust/psibase/src/graph_ql.rs
@@ -394,8 +394,11 @@ impl<'de> Deserialize<'de> for SqlRow {
 
         let rowid = map
             .remove("rowid")
-            .and_then(|v| v.as_i64())
-            .ok_or_else(|| serde::de::Error::missing_field("rowid"))? as i32;
+            .and_then(|v| match v {
+                Value::String(s) => s.parse::<i32>().ok(),
+                _ => None,
+            })
+            .ok_or_else(|| serde::de::Error::missing_field("rowid"))?;
 
         let data = Value::Object(map.into_iter().collect());
 
--- a/rust/psibase/src/graph_ql.rs
+++ b/rust/psibase/src/graph_ql.rs
@@ -381,7 +381,7 @@ pub trait EventDb {
 
 #[derive(Debug)]
 struct SqlRow {
-    rowid: i32,
+    rowid: u64,
     data: Value,
 }
 
@@ -395,7 +395,7 @@ impl<'de> Deserialize<'de> for SqlRow {
         let rowid = map
             .remove("rowid")
             .and_then(|v| match v {
-                Value::String(s) => s.parse::<i32>().ok(),
+                Value::String(s) => s.parse::<u64>().ok(),
                 _ => None,
             })
             .ok_or_else(|| serde::de::Error::missing_field("rowid"))?;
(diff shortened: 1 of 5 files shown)


James-Mart:
Build should work when https://github.com/gofractally/psibase/pull/1011 is merged

rust/psibase/src/graph_ql.rs
```diff
@@ -374,3 +378,284 @@ pub trait NamedEvent {
 pub trait EventDb {
     fn db() -> DbId;
 }
+
+#[derive(Debug)]
+struct SqlRow {
+    rowid: i32,
```
swatanabe, line 384: The event id is u64. i32 will eventually overflow for a long running chain.

rust/psibase/src/graph_ql.rs
```diff
@@ -374,3 +378,284 @@ pub trait NamedEvent {
...
+    data: Value,
+}
+
+impl<'de> Deserialize<'de> for SqlRow {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: Deserializer<'de>,
+    {
+        let mut map: HashMap<String, Value> = HashMap::deserialize(deserializer)?;
```
swatanabe, line 393: Can this use `serde_json::Map` directly and avoid the extra conversion?

rust/test_contract_2/src/lib.rs
```diff
@@ -95,6 +132,36 @@ mod service {
         async fn event(&self, id: u64) -> Result<event_structs::HistoryEvents, anyhow::Error> {
             get_event(id)
         }
+
+        async fn get_add_events(
```
swatanabe, line 139: Nothing using `condition`?

rust/psibase/src/time.rs
```diff
@@ -64,8 +64,10 @@ impl Serialize for TimePointSec {
 
 impl<'de> Deserialize<'de> for TimePointSec {
     fn deserialize<D: Deserializer<'de>>(deserializer: D) -> Result<Self, D::Error> {
-        let s = <&str>::deserialize(deserializer)?;
-        Ok(DateTime::<FixedOffset>::parse_from_rfc3339(s)
+        // String was needed over <&str> because deserialization during serde_json::from_value
+        //   cannot provide a borrowed string.
```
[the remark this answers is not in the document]
James-Mart: You're right, nice! Done

rust/psibase/src/graph_ql.rs
James-Mart: Could have sworn I saw an i32 being used somewhere, because I also thought that seemed small.
Anyway, fixed

rust/psibase/src/graph_ql.rs
James-Mart: Yep, good call - done

rust/test_contract_2/src/lib.rs
James-Mart: I had tested it manually before... but you're right. I just added a bunch of tests to demonstrate that the other functionality is working as expected

rust/test_contract_2/src/lib.rs
```diff
@@ -138,17 +138,23 @@ mod service {
 
         async fn get_add_events(
             &self,
+            condition: Option<String>,
```
swatanabe, line 141: I hope no one uses this as an example to copy for real services.

swatanabe (approved):

PR #1018, opened 2025-01-28
docs fixes

James-Mart:
Primary fix: Update `index.hbs`, which became outdated with newer versions of mdbook (causing our nav menu to break).

Additional improvements:
 - ~~Uses a local version of mermaidjs, which is better for offline rendering~~
   - Reverted back to dynamically loaded mermaidjs, in order to minimize uploaded artifact size
 - Adds support for tab controls in mdbook documents


swatanabe:
This isn't working for me. I just updated all the mdbook tools:

||version|
|-|-|
|mdbook | v0.4.44|
|mdbook-mermaid | 0.14.1|
|mdbook-pagetoc | v0.2.0|
|mdbook-tabs | v0.2.1|
|Firefox|135.0|

```sh
cd doc/psidk && mdbook serve
```

The table of contents toggle is inert and I also can't expand the items in the TOC.

swatanabe:
It seems to work better when served from chain, but I'm still seeing a few problems:
- The TOC toggle hides the TOC but doesn't reclaim the space it occupied
- Hide TOC, then click next link: The TOC toggle now shifts the main text to make room for the TOC, but the TOC is not actually displayed.

swatanabe (approved):

PR #1019, opened 2025-01-28
Introduce transact plugin hooks

James-Mart:
## Transact plugin hooks

Adds hooks to the transact plugin, allowing third-party apps to "hook" into various functionality to dynamically modify the process of transaction construction and authorization.

For example:
* The invite plugin uses hooks to ensure that its users are able to sign and submit transactions as the `invited-sys` user with custom auth claims/proofs in order to create their first account.
* The auth-sig plugin uses hooks to add claims/proofs
* The staged-tx plugin will use hooks to wrap action into staged-tx proposals

### Transaction construction lifecycle with hooks

The transaction construction cycle, including hooks, is as follows:

* *start-tx*

* *add-action-to-transaction*
  * on-action-sender           - the hooked plugin can set the sender of the action, otherwise the logged-in user is used by default
  * on-action-auth-claims      - the hooked plugin can add claims for this action
  * save action details

* *finish-tx*
  * on-user-claim              - the user auth plugin adds the user claim
  * on-tx-transform            - the hooked plugin can transform the set of actions in the transaction
  * construct transaction
  * hash-transaction
  * on-user-auth-proof         - the user auth plugin adds the user proof
  * on-action-auth-proofs      - the hooked plugin can add proofs for their action claims
  * publish transaction

### Hook worlds

The transact plugin's wit file contains worlds for using various hooks. Users of these hooks simply include the world corresponding to their hook, and their toolchain will enforce that their plugin exports the necessary functionality to be compatible with the transact plugin's expectations.

```
world hook-action-auth {
    import hooks;
    export transact-hook-action-auth: interface {
        use types.{claim, proof, action};
        on-action-auth-claims: func(action: action) -> result<list<claim>, error>;
        on-action-auth-proofs: func(claims: list<claim>, transaction-hash: list<u8>) -> result<list<proof>, error>;
    }
}
```

## Support inline interfaces 

Allows the component parser (used by supervisor to understand plugins) to support plugins with interfaces defined inline in a world, rather than requiring them to be defined outside of a world and pulled in.

Better adherence to the [wit world](https://github.com/WebAssembly/component-model/blob/main/design/mvp/WIT.md#wit-worlds) spec.


James-Mart:
@brandonfancher adding you as a reviewer because of the updates to the supervisor / plugin loading code

services/user/Supervisor/ui/src/plugin/pluginLoader.ts
brandonfancher, line 1: Deleted code is tested code 🔥

services/user/Supervisor/ui/src/supervisor.ts
```diff
@@ -79,13 +79,29 @@ export class Supervisor implements AppInterface {
...
+                // Consider deleting the user from the accounts plugin db
+            } else {
+                // Current limitation: an auth service plugin must be called "plugin" ("<service>:plugin")
+                this.loader.trackPlugins([
+                    pluginId(account.authService, "plugin"),
+                ]);
+            }
+            await this.loader.processPlugins();
+            await this.loader.awaitReady();
+        }
```
brandonfancher, line 104: Are lines `102` and `103` supposed to run even when the user account is invalid?

Assuming they are not and assuming you're okay with early returns, this all could be:

```ts
// Phase 2: Loads plugins needed by the current user
const user = this.getLoggedInUser();
if (!user) return;

const account = this.getAccount(user);
if (account === undefined) {
    console.warn(`Invalid user account '${user}' detected. Automatically logging out.`);
    this.logout();
    // Consider deleting the user from the accounts plugin db
    return;
}

// Current limitation: an auth service plugin must be called "plugin" ("<service>:plugin")
this.loader.trackPlugins([
    pluginId(account.authService, "plugin"),
]);
await this.loader.processPlugins();
await this.loader.awaitReady();
```

brandonfancher (approved):

services/user/Supervisor/ui/src/supervisor.ts
James-Mart: Thanks, refactored

swatanabe (approved):

PR #1026, opened 2025-01-30
Add staged tx plugin

James-Mart:
This adds the staged-tx plugin to allow for any plugins on the front-end to leverage the staged-tx infrastructure.

The branding plugin is used as an example to show how easy it is to integrate with staged-tx.


services/system/StagedTx/plugin/src/lib.rs
```diff
@@ -0,0 +1,226 @@
...
+
+#[derive(Deserialize, Serialize)]
+struct StagedTxData {
+    details: StagedTxDetailsInner,
+}
+
+#[derive(Deserialize, Serialize)]
+struct StagedTxDetailsInner {
+    txid: String,
+    #[serde(rename = "proposeBlock")]
```
swatanabe, line 52: rename_all = "camelCase" is less repetitive.

services/system/StagedTx/plugin/src/lib.rs
```diff
@@ -0,0 +1,226 @@
...
+    let details = serde_json::from_str::<StagedTxDetails>(&details).unwrap();
+    let txid = details.data.details.txid;
+    if txid.len() % 2 != 0 {
+        return Err(ErrorType::InvalidTxId.into());
+    }
+    let txid: Vec<u8> = (0..txid.len())
+        .step_by(2)
+        .map(|i| u8::from_str_radix(&txid[i..i + 2], 16).unwrap())
+        .collect();
+    let txid: [u8; 32] = txid.try_into().map_err(|_| ErrorType::InvalidTxId)?;
```
swatanabe, line 106: Can't all of this be handled by using `Checksum256` instead of `String` for the field?

services/system/StagedTx/plugin/src/lib.rs
```diff
@@ -0,0 +1,226 @@
...
+            .into_iter()
+            .map(|group| propose_wrap(group, &label))
+            .collect();
+
+        Ok(Some(actions))
+    }
+}
+
+fn propose_wrap(mut actions: Vec<Action>, label: &str) -> Action {
+    let sender = actions[0].sender.clone();
```
swatanabe, line 186: The logic here is pretty convoluted. I don't like the way the actions are created with the wrong sender and then patched up later. Note that having the wrong sender does not just mean that an action will fail. It could make the action mean something else entirely. Could we provide a way to set the sender explicitly when adding an action to the transaction, and have this only wrap the actions in `propose`?

services/system/StagedTx/plugin/src/lib.rs
James-Mart: done

services/system/StagedTx/plugin/src/lib.rs
James-Mart: When I make a sql query for the event with `WHERE txid='HEXDATA'` when the type is `Checksum256`, I always get back an empty array.

services/system/StagedTx/plugin/src/lib.rs
James-Mart: I decided to go for it. Can always roll it back, but I don't think it's actually a real security problem. 
Now plugins can ask `get-current-user`. I changed it from `get-logged-in-user` to avoid confusion over which user is being returned, the one logged in to "my plugin's app" or the one logged in to the top-level app. I think "current user" is clearer.

services/system/StagedTx/plugin/src/lib.rs
swatanabe: I don't understand. Aren't the types here only used for deserializing? How do they affect the query?

swatanabe: Also, try `WHERE txid=X'HEXDATA'` (a blob literal)

services/system/Transact/plugin/src/helpers.rs
```diff
@@ -33,16 +33,15 @@ pub fn get_sender_app() -> String {
 
 pub fn get_action_sender(service: &str, method: &str) -> Result<String, Host::types::Error> {
     if let Some(plugin) = ActionSenderHook::get() {
-        ActionSenderHook::clear();
```
swatanabe, line 36: I think this interface creates a high risk of applying the sender to actions that it isn't intended for. Is there ever a case where the caller of `add_action_to_transaction` doesn't know the correct sender?

services/system/StagedTx/plugin/src/lib.rs
```diff
@@ -0,0 +1,223 @@
...
+impl Proposer for StagedTxPlugin {
+    fn set_propose_latch(account: Option<String>) -> Result<(), Error> {
+        if let Some(ref acc) = account {
+            validate_account(acc)?;
+        }
+
+        hook_tx_transform_label(account.as_deref());
+
+        CurrentSender::set(account);
+        hook_actions_sender();
```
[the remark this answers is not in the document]
swatanabe: Nvm. I got confused keeping track of which sender is which.

services/system/Transact/plugin/src/helpers.rs
James-Mart: This is the tradeoff of using a hook to set the sender so the actions are always added with the correct sender. The set action sender hook used to only apply to the next action added, but if we are using it for staged tx, then it needs to be persistent so that all actions added downstream of the propose latch have their sender set correctly.

The caller of `add_action_to_transaction` typically does not care who the sender is, they let transact manage it

swatanabe: In the case of `branding` "all actions added downstream" is wrong. It's only those specific actions that should use `branding` as the sender.

James-Mart: Well, there are no actions added downstream because branding doesn't call any other plugins. If it did, then we would want all those added downstream to be part of the staged-tx.

What should actually happen is that when the plugin who sets the latch pops off the callstack, the latch is unset. But without a mechanism to do that, I suppose it should be standard to call unlatch() at the end?
So branding should be:
```
latch()
add_action_to_tx()
unlatch()
```

Thoughts?

swatanabe: Making the sender effectively dynamically scoped would address my concerns here. rust has RAII, so the unlatch doesn't need to be explicit.

James-Mart: But it would be up to the plugin to build the raii wrapper for the latch. Or are you suggesting something that I'm missing?

services/system/StagedTx/plugin/src/lib.rs
James-Mart: This worked. Updated

services/system/Transact/plugin/src/helpers.rs
swatanabe: Oh right, it's importing wit, not a regular library.

James-Mart: I think in the future I might be able to define the RAII in the staged-tx plugin itself, using a [resource](https://github.com/WebAssembly/component-model/blob/main/design/mvp/WIT.md#item-resource).

But resources are currently not supported by our wasm component host.

James-Mart: Added the ability for `staged-tx` to unhook from `transact` by calling `set_propose_latch(None)` and added RAII to `branding` to automatically manage hook/unhook.
Added comment in staged-tx as a reminder to move the RAII into staged-tx when resources are supported.

services/user/Branding/plugin/src/lib.rs
```diff
@@ -17,14 +17,31 @@ use errors::ErrorType;
...
     fn set_network_name(name: String) {
+        let _latch = Latch::new();
+
         let packed_network_name_args = branding::action_structs::setNetworkName { name }.packed();
-        set_propose_latch(Some("branding")).unwrap();
         add_action_to_transaction("setNetworkName", &packed_network_name_args).unwrap();
     }
     fn set_logo(logo: Vec<u8>) {
-        set_propose_latch(Some("branding")).unwrap();
+        let _ = Latch::new();
```
swatanabe, line 43: Is this right? I thought _ doesn't bind anything nor does it extend the lifetime of the initializer.

James-Mart: You're correct, I noticed that and fixed it for `set_network_name` and forgot to update it for `set_logo`.
Fixed

services/system/Transact/plugin/src/helpers.rs
swatanabe: This works, but it relies on the plugins cooperating. I'm concerned about the security implications if they don't.

For example, let alice be the current user who has her own account, and also has signing power for bob.
```
mallory::evil()
tokens::transfer(carol, 100 PSI)
```

I would expect that mallory could derail the whole thing. I would also expect mallory to be able to add new actions that are sent to mallory's service. I would not expect mallory to be able to change the transfer to come from bob.

I don't think a resource completely solves the problem, because a plugin could choose to keep the resource alive longer than it's supposed to.

James-Mart: Okay, I just decided to make staged-tx privileged so it can import `get-active-app` from the host. That allows it to restrict `set_propose_latch` to only the active application or its plugin. There is therefore no problem if the latch is not explicitly unset because the transact hook will expire after the transaction context.

Furthermore, for security, I added some explicit (but temporary) checks in the transact plugin hooks to whitelist the apps that are allowed to hook. We can relax this once we have the client-side oauth stuff, but this keeps the implementation safe in the meantime.

services/system/Transact/plugin/src/lib.rs
```diff
@@ -54,11 +54,19 @@ impl Hooks for TransactPlugin {
     }
 
     fn hook_actions_sender() {
-        if let Some(sender) = ActionSenderHook::get() {
-            if sender != get_sender_app() {
+        let sender_app = get_sender_app();
+
+        if let Some(hooker) = ActionSenderHook::get() {
```
swatanabe, line 59: I'd find another name for the variable.

swatanabe (approved):

services/system/Transact/plugin/src/lib.rs
James-Mart: Lol, done

PR #1028, opened 2025-01-30
Key device selection and unlock during production boot

brandonfancher:
## In this PR
See commits for full list of what was done in this PR.


services/user/XAdmin/ui/src/components/forms/chain-mode-form.tsx
```diff
@@ -21,6 +21,7 @@ interface Props {
 
 export const ChainTypeForm = ({ form, next }: Props) => {
     const isSecure = window.location.protocol === "https:";
+    const isLocalhost = window.location.hostname.endsWith("localhost");
```
swatanabe, line 24: https://developer.mozilla.org/en-US/docs/Web/API/Window/isSecureContext

brandonfancher: Ah, I didn't know about this. Thank you!

services/user/XAdmin/ui/src/components/forms/key-device-form.tsx
```diff
@@ -0,0 +1,158 @@
...
+                </CardDescription>
+            </CardHeader>
+            <CardContent>
+                {isSuccess && !keyDevices.length ? (
+                    <div className="mb-4 flex max-w-[450px] items-center gap-3">
+                        <p className="text-2xl">⚠️</p>
+                        <p>
+                            No key devices were found. Please ensure one is
+                            available. If you'd like to boot without a key
+                            device, go back and select the "Development" option.
```
swatanabe, line 71: This could do better to express the reason: development does not have a key, so it doesn't need a secure place to store it (as opposed to having a key but not storing it securely).

swatanabe:
>     * [ ]  This still needs to be tested with a hardware yubikey or similar before merging

I don't really recommend a yubikey. It can be made to work, but it's not designed for this.

brandonfancher:
> I don't really recommend a yubikey. It can be made to work, but it's not designed for this.

Are there any other kinds of devices this should be tested with?

services/user/XAdmin/ui/src/hooks/useKeyDevices.ts
```diff
@@ -0,0 +1,56 @@
...
+import { queryKeys } from "@/lib/queryKeys";
+import { chain } from "@/lib/chainEndpoints";
+import { exportKeyToDER, generateP256Key } from "@/lib/keys";
+
+export const useKeyDevices = () =>
+    useQuery({
+        queryKey: queryKeys.keyDevices,
+        queryFn: async () => {
+            try {
+                return chain.getKeyDevices();
```
Velua, line 12: By returning this promise without `await` I think it's gonna skip over the `catch` logic you have. 
```suggestion
                return await chain.getKeyDevices();
```

services/user/XAdmin/ui/src/hooks/useKeyDevices.ts
```diff
@@ -0,0 +1,56 @@
...
+            } catch (e) {
+                console.error("Failed to fetch key devices", e);
+                throw "Failed to fetch key devices";
```
Velua, line 15: For the sake of consistency I think we should aim for all promises / hooks to throw an `Error` rather than a `string` so then when displaying them on the UI it's the same, otherwise we'll end up with more `e instanceof Error` and `typeof e == 'string' to try parse some kinda message to display.
```suggestion
                throw new Error("Failed to fetch key devices");
```

services/user/XAdmin/ui/src/components/forms/key-device-form.tsx
```diff
@@ -0,0 +1,158 @@
...
+                ) : null}
+                {isError ? (
+                    <div className="mb-4 flex max-w-[450px] items-center gap-3">
+                        <p className="text-2xl">⚠️</p>
+                        <p>
+                            There was an error fetching key devices. See the
+                            console for more information.
+                        </p>
+                    </div>
+                ) : null}
```
Velua, line 83: Can we just display the error message here? If it's a simple error the user might be able to recover themselves rather than reach out to a dev.

services/user/XAdmin/ui/src/components/installation-summary.tsx
```diff
@@ -0,0 +1,107 @@
...
+    hint?: string;
+}
+
+const Chip = ({ label, value, hint }: ChipProps) => (
+    <div className="w-60 truncate rounded-md border p-2 text-right">
+        <div className="text-sm text-muted-foreground">{label}</div>
+        {value}{" "}
+        {hint ? (
+            <Popover>
+                <PopoverTrigger>ℹ️</PopoverTrigger>
```
Velua, line 32: I think [this](https://lucide.dev/icons/info) comes across as more Shadcn like

services/user/XAdmin/ui/src/pages/create-page.tsx
```diff
@@ -99,13 +87,38 @@ export const CreatePage = () => {
...
+        previous,
+        currentStepNum,
+        currentStep,
+        maxSteps,
+    } = useStepper<StepKey>([
+        { step: Step.ChainType, form: chainTypeForm },
+        { step: Step.BlockProducer, form: blockProducerForm },
+        { step: Step.KeyDevice, form: keyDeviceForm, skip: isDev },
+        { step: Step.Confirmation },
+    ]);
```
Velua, line 119: Gives me `enum` vibes

```suggestion
    const Step = z.enum([
        "ChainType",
        "BlockProducer",
        "KeyDevice",
        "Confirmation",
        "Completion",
    ]);

    type StepKey = z.infer<typeof Step>;

    const {
        canNext,
        canPrev,
        next,
        previous,
        currentStepNum,
        currentStep,
        maxSteps,
    } = useStepper<StepKey>([
        { step: Step.Enum.ChainType, form: chainTypeForm },
        { step: Step.Enum.BlockProducer, form: blockProducerForm },
        { step: Step.Enum.KeyDevice, form: keyDeviceForm, skip: isDev },
        { step: Step.Enum.Confirmation },
    ]);
```

James-Mart:
I verified this works, and was able to use this interface to generate the key in the softhsm to store a keypair that gets used for block signing.

Worth noting that this still keeps the block and transaction signing keys the same. Do you want to separate them in this PR? Or in a followup?

brandonfancher:
> Do you want to separate them in this PR? Or in a followup?

Followup

services/user/XAdmin/ui/src/pages/create-page.tsx
brandonfancher: Yes, it works similar to an enum, but better. I try to avoid TS enums, as they're considered harmful by many. As for zod here, I get the value add from use cases with external data sources like API responses or user inputs. But I'm hesitant to zodify everything. And while we iterate early on, you'll see me opting for plain TS for dev speed and simplicity.

services/user/XAdmin/ui/src/hooks/useKeyDevices.ts
brandonfancher: Good catch. Thanks.

services/user/XAdmin/ui/src/components/forms/key-device-form.tsx
brandonfancher: Yeah, I can do that. I figure anyone booting a chain should know how to look at the console. But I don't feel strongly here either way.

PR #1029, opened 2025-02-01
Plugin tester - rich parameter editor

James-Mart:
Creates a rich parameter editor (as opposed to the raw JSON parameter input) for all plugins in the plugin tester, for the following `wit` types:

#### Primitives

* u8
* s8
* u16
* s16
* u32
* s32
* u64
* s64
* string
* char

#### Complex types

* enum
* flags
* variant
* tuple
* option
* list
* record

This has support for all parameter types listed in the [`Wit documentation`](https://github.com/WebAssembly/component-model/blob/main/design/mvp/WIT.md) except `resources` and `results`.  But `resources` are not yet supported by the plugin-host anyway, and `results` are generally only used for returns, not parameters.

A few screenshots:

`string`
image

`record` and `list<string>`
image

File input type for `list<u8>`
image

Other options for `list<u8>`
image


services/user/CommonApi/common/packages/plugin-tester/ui/src/components/ServiceInput.tsx
```diff
@@ -13,19 +13,27 @@ export function ServiceInput({
...
+        onKeyDown={handleKeyDown}
       />
       <input
-        style={{ padding: "0.5rem" }}
+        className="service-input"
         placeholder="Plugin name"
         value={plugin}
         onChange={(e) => onPluginChange(e.target.value)}
+        onKeyDown={handleKeyDown}
       />
```
brandonfancher, line 37: Normally we'd wrap the two inputs in a `form` component with an `onSubmit` prop:

```ts
const onSubmit = (e: React.KeyboardEvent) => {
    e.preventDefault();
    onLoad();
}

<form className="service-input-container" onSubmit={onSubmit}>
    // inputs without the `onKeyDown` prop
    <button type="submit" className="common-button">Load</button>
</form>
```

When `Enter` is pressed, the form submit handler fires.

services/user/CommonApi/common/packages/plugin-tester/ui/src/components/editors/RawParameterEditor.tsx
```diff
@@ -0,0 +1,34 @@
...
+
+interface RawParameterEditorProps {
+  values: Record<string, unknown>;
+  onChange: (values: Record<string, unknown>) => void;
+}
+
+export const RawParameterEditor: FC<RawParameterEditorProps> = ({
+  values,
+  onChange,
+}) => {
```
brandonfancher, line 11: Best practice for functional components is what you're doing elsewhere without `FC`:

`export const RawParameterEditor = ({ values, onChange }: RawParameterEditorProps) => { ... }`

brandonfancher: Before React 18, `React.FC` was not recommended, for various reasons. It's technically fine now. More a style preference than anything else.

services/user/CommonApi/common/packages/plugin-tester/ui/src/components/editors/RawParameterEditor.tsx
```diff
@@ -0,0 +1,34 @@
...
+    } catch {
+      console.error("Invalid JSON");
+    }
+  };
+
+  return (
+    <textarea
+      className="common-textarea"
+      value={JSON.stringify(values, null, 2)}
+      onChange={(e) => handleChange(e.target.value)}
```
brandonfancher, line 31: Doing this `onChange` means the user really cannot free type in the field. Every key pressed must result in good JSON or the change doesn't happen. I assume that's what you're going for, right? (Makes sense, I guess. Keeps it valid and in sync with the rich editor.) Alternatively, you can validate on submit.

brandonfancher (approved):
Looks good. Cool tester!

services/user/CommonApi/common/packages/plugin-tester/ui/src/components/ServiceInput.tsx
brandonfancher: It works as is, so whether you want to make that change or not is up to you.

James-Mart:
Will address your comments in a followup pr @brandonfancher

PR #1035, opened 2025-02-07
Reduce copies in cpp graphql lib

James-Mart:
This PR attempts to avoid making some of the copies that were previously made by the cpp graphql library.
It also adds better pagination controls to the block query in explorer.

Note that this isn't enough to fix the OOM with the first page of the explorer.


libraries/psibase/common/include/psibase/serveGraphQL.hpp
```diff
@@ -317,7 +317,7 @@ namespace psibase
       auto       add_edge = [&](const auto& it)
       {
          auto cursor = psio::to_hex(it.keyWithoutPrefix());
-         result.edges.push_back(typename Connection::Edge{proj(*it), std::move(cursor)});
+         result.edges.push_back(typename Connection::Edge{proj(it.view()), std::move(cursor)});
```
swatanabe, line 320: Construction of Edge creates a full copy

libraries/psibase/common/include/psibase/Table.hpp
```diff
@@ -257,6 +257,16 @@ namespace psibase
...
       T operator*() const { return psio::from_frac<T>(psio::prevalidated{*base}); }
 
+      /// get object view
+      ///
+      /// This returns a view of the object in the database. It does not cache; it returns a view of the
+      /// fresh object each time it's used.
+      psio::view<T> view() const
+      {
+         auto buffer = *base;
+         return psio::view<T>{psio::prevalidated{buffer.data()}};
```
swatanabe, line 267: U.B. Need to use `shared_view_ptr`.

libraries/psibase/common/include/psibase/serveGraphQL.hpp
James-Mart: reverted

libraries/psio/include/psio/graphql.hpp
```diff
@@ -1600,72 +1599,94 @@ namespace psio
...
+         std::vector<char> err_result(err_measure.written());
+         fixed_buf_stream  error_stream(err_result.data(), err_result.size());
+         write_error_json(error, error_stream);
+         return err_result;
+      }
+
+      // Write valid data
+      std::vector<char> result(measure.written());
+      fixed_buf_stream  out(result.data(), result.size());
+      write_data_json(value, query, variables, allow_unknown_members, out, error);
```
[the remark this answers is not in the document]
James-Mart: I couldn't think of how else to avoid copies then to know the size, and then read directly into the `vec<char>`. Is there a simpler way to do this?

services/user/Explorer/src/Explorer.cpp
```diff
@@ -7,14 +7,26 @@
 
 struct Query
 {
-   auto blocks() const
+   /// Retrieve blocks
```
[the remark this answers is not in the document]
James-Mart: Does it actually do anything? Or are the parameters added to the schema automatically these days?

Issue #749, opened 2024-06-17
Default app->app permissions ψauth page


James-Mart:
# Introduction

Plugins need to manage permissions related to sharing private user data between psibase apps. This can be a complicated task to do well, because permission structures can be highly complex.

The simplest permission configuration strategy is a simple configuration for whether the user trusts App A completely to use the plugin data and functionality of App B. This "full permissions" strategy may be too permissive in many circumstances, but it may still be a fairly common way for smaller apps to manage permissions, especially when the app is new.

Therefore, we need to build this "full permission" ψauth dialog and make it accessible from any plugin with limited configuration (probably through the URL querystring).  A basic client-side app-to-app permission structure oauth flow should be created to make it simpler for apps to do permissions management.

This depends on https://github.com/gofractally/psibase/issues/747

Issue #747, opened 2024-06-17
ψAuth - psibase oauth flow


James-Mart:
# Introduction

After various redesigns, it is still the case that our client-side infra requires the ability for apps to pop up windows in new tabs from their own domain. This is probably primarily used to set up sharing of private user data between apps.

# Tasks

- [x] Review and update specification
- [x] Implementation

PR #1038, opened 2025-02-10
Default Permissions Infra v0.1 Handles round 1 for #747, #749

cool-ant:
This is the first implementation of the "permissions OAuth" flow.
WASM doesn't support async calls yet, so this isn't as smooth as it eventually will be, but this gives us a version of this tech that works with some caveats and can easily be augmented when the async upgrades to WASM are completed.
Related issue (which I'm leaving open to include the eventual async work) #747, #749.

Permissions OAuth works on a per-method-call basis. So Caller calling Callee.plugin's xxxxx method is verified as permitted by Callee via a new Permissions Plugin.

To see this in action,
1) open any plugin in the plugin tester, e.g., http://branding.psibase.localhost:8079/common/plugin-tester and in the upper-most left text box that says "branding", change it to "auth-sig" and click "Load".

2) We'll trigger an action that requests a permission check via Auth-Sig's plugin's `setkey` method.
 2a) Click the "Keyvault" tab, then click `generate-keypair`, then the "Execute" button. Copy the public key (everything in the Response textbox other than the enclosing double quotes).
 2b) Click the "Actions" tab, then click "set-key", and paste the pub key into the "publicKey" field.
 2c) Due to an encoding issue in the tester, click on the "Raw" tab, delete the "public-key" line, and then remove the first backslash of the double-escaped newline characters in the pub key (so it looks like what you pasted). There are 4 double-escaped newline characters that need attention.
 2d) Finally, click the "Execute" button.

2) if a permissions setting hasn't already been saved (which would auto-approve/transparently-approve the request, a dialog will be displayed for the user to Approve/Disapprove the call (and the original call will fail with an instructional message in the browser console),
image

3) if user approves, the approval is saved by Permissions to local storage (with a key `clientdata:permissions:[callee]<-[caller]` in the Supervisor's subdomain to enable auto-approve going forward,
image

4) and finally repeating the original user action will now be autoapproved, since the approval is now stored in local storage, and you'll see a transaction happen in the console.
image


brandonfancher (commented):
Just a few comments. UI improvements can come in future PR.

- For me, the popup opened in a new tab. I'm using the Chromium-based Arc browser. It seemed to do the job, though.
- I noticed we cannot deselect "Remember this answer." Is that on purpose?
- Ultimately, the call I got permission for wasn't working. Not sure if it's related to this PR or something else.

Tomorrow (or whenever you're feeling better), can we walk through the functionality once more together?

Thanks!

James-Mart:
Uh oh. In testing this I noticed that the popup is detected as coming from branding.psibase (the parent window) instead of supervisor.psibase (the domain actually responsible)

image

Is this also what you saw @cool-ant ?

services/user/Permissions/plugin/Cargo.toml
```diff
@@ -0,0 +1,26 @@
+[package]
+name = "permissions"
+version.workspace = true
+rust-version.workspace = true
+edition.workspace = true
+publish = false
```
James-Mart, line 6: This should be `publish.workspace` now that `publish` is part of the workspace package.

services/system/AuthSig/plugin/src/lib.rs
```diff
@@ -134,10 +136,12 @@ impl KeyVault for AuthSig {
 
 impl Actions for AuthSig {
     fn set_key(public_key: Pem) -> Result<(), CommonTypes::Error> {
-        // TODO: check if sender authorizes caller app to set key
-        // Currently only an AuthSig app would be able to set a user's key
+        let caller = get_sender_app().app.unwrap();
+
         if !from_auth_sig_ui() {
-            return Err(Unauthorized("set_key").into());
+            if is_permitted(&caller)? {
```
James-Mart, line 142: if `!is_permitted`

James-Mart:
> r

I  made [a new issue](https://github.com/gofractally/psibase/issues/1074) about this. We do not need to hold up this PR from proceeding before we solve that issue. It can be a follow up improvement.

services/user/Permissions/plugin/src/db.rs
```diff
@@ -0,0 +1,68 @@
...
+impl CurrentAccessRequest {
+    pub fn set(caller: String, callee: String) -> Result<(), Error> {
+        let pr = serde_json::to_string(&PermissionRequest {
+            caller,
+            callee,
+            expiry_timestamp: TimePointSec::from(Utc::now()),
+        })
+        .map_err(|err| ErrorType::SerderError(err.to_string()))?;
+
+        Keyvalue::set(&PERM_OAUTH_REQ_KEY, &pr.as_bytes())
```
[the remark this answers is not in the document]
cool-ant: ah, never been exposed to it. Thanks James. super easy to use

James-Mart (approved):

brandonfancher (approved):

services/user/CommonApi/common/packages/wit/host-common.wit
```diff
@@ -101,8 +101,15 @@ interface client {
     my-service-origin: func() -> string;
 }
 
+interface web {
+    use types.{error};
+
+    popup: func(url-path: string) -> result<_, error>;
```
swatanabe, line 107: How would this work if we run a plugin in a non-browser context (like the CLI).

James-Mart: It would not. That's why it's in an interface called, `web`. Also it's probably not the final interface, just a first pass to start playing with the functionality

PR #1041, opened 2025-02-12
Fix failed transactions that emit events

swatanabe:
This would brick the chain because `putSequential` used a cached member variable which was not rolled back when the transaction aborted. Remove `BlockContext::databaseStatus` and just look it up as needed.


James-Mart (approved):
Tested and working :+1:

PR #1047, opened 2025-02-14
Update JS dependencies

swatanabe:
~~In Chainmail, the current version of `@milkdown/theme-nord` causes the build to fail. I pinned it to an older version that works. I did not investigate the reason for the failure.~~  It is not compatible with tailwind v3.


services/user/Chainmail/ui/package.json
brandonfancher, line 1: These updates have introduced UI regressions to the Chain Mail message editor. I imagine that has to do with the `@prosemirror-adapter/react`. And this isn't the first time I've had to fix regressions like this after milkdown- and prosemirror-related dependency updates.

I suggest we avoid updating Chainmail UI dependencies in this PR and let me do that separately.

Thoughts?

brandonfancher (changes requested):

services/user/Chainmail/ui/package.json
swatanabe: Works for me

brandonfancher (approved):

PR #1051, opened 2025-02-14
Add queries to SetCode and Sites

James-Mart:
### SetCode

```
type CodeRecord {
    codeNum: String!
    flags: [String!]!
    codeHash: String!
    vmType: Int!
    vmVersion: Int!
    codeHex: String!
}
type Query {
    code(account: String!): CodeRecord
}
```

### Sites

```
type SiteConfig {
    account: String!
    spa: Boolean!
    cache: Boolean!
    globalCsp: String!
}
type SitesContentRow {
    account: String!
    path: String!
    contentType: String!
    content: [Int!]!
    hash: String!
    contentEncoding: String
    csp: String
}
type SitesContentRowEdge {
    node: SitesContentRow!
    cursor: String!
}
type PageInfo {
    hasPreviousPage: Boolean!
    hasNextPage: Boolean!
    startCursor: String!
    endCursor: String!
}
type SitesContentRowConnection {
    edges: [SitesContentRowEdge!]!
    pageInfo: PageInfo!
}
type Query {
    getConfig(account: String!): SiteConfig
    getContent(account: String! gt: String ge: String lt: String le: String first: Float last: Float before: String after: String): SitesContentRowConnection!
}
```


services/system/SetCode/src/RSetCode.cpp
```diff
@@ -1,17 +1,40 @@
...
 
 namespace SystemService
 {
+   using namespace psibase;
+
+   struct Query
+   {
+      auto code(AccountNumber account) const
+      {
+         return kvGet<CodeRow>(CodeRow::db, codeKey(account));
```
swatanabe, line 17: Needing to make two separate queries is not very GraphQLish. The result should have a member that will follow the link to the CodeByHashRow.

services/system/SetCode/src/RSetCode.cpp
```diff
@@ -1,17 +1,40 @@
...
+      }
+
+      auto codeRaw(Checksum256 code_hash, uint8_t vm_type, uint8_t vm_version) const
+      {
+         return kvGet<CodeByHashRow>(CodeByHashRow::db,
```
swatanabe, line 22: I'm not sure that we want to expose `numRefs`.

services/system/SetCode/src/RSetCode.cpp
James-Mart: The reason I didn't do this is because I didn't want to pay for pulling the entire code into memory on every query even if it wasn't requested.

But since you suggested this, I tried using a method resolver for the `rawCode` field and it's working! I had no idea our gql implementation supported this.

rust/psibase/src/main.rs
```diff
@@ -985,7 +985,7 @@ async fn apply_packages<
                 // Remove out-dated files. This needs to happen before installing
                 // new files, to handle the case where a service that stores
                 // data files is replaced by a service that does not provide
-                // removeSys.
+                // remove.
```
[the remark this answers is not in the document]
James-Mart: deleted comment

PR #1053, opened 2025-02-15
Fix identity tests

James-Mart:
CC @swatanabe 
 - I updated TimePointSec to a gql scalar to align it with TimePointUSec


cool-ant (approved):
:shipit:

PR #1055, opened 2025-02-17
Use in memory action construction

James-Mart:
Avoids temporarily storing actions in localstorage, which fails for large actions (like sites:upload) due to restrictions on kv value size.

CC @Velua

### DIFF
--- a/services/system/Transact/plugin/src/lib.rs
+++ b/services/system/Transact/plugin/src/lib.rs
@@ -87,7 +87,7 @@ impl Hooks for TransactPlugin {
             }
         }
 
-        // Temporary whitelist until we have oauth
+        // TODO: Use `permissions` oauth instead of hardcoded whitelist
         if transformer != "staged-tx" {
             panic!(
                 "hook_tx_transform_label: {} is not whitelisted",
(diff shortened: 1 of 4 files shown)


services/system/Transact/plugin/src/lib.rs
```diff
@@ -166,7 +166,6 @@ impl Admin for TransactPlugin {
         if actions.len() == 0 {
             return Ok(());
         }
-        CurrentActions::clear();
```
swatanabe, line 169: Why doesn't it need to be cleared?

swatanabe: I think leaving it in is harmless and is more consistent and future-proof.

swatanabe (approved):

services/system/Transact/plugin/src/lib.rs
James-Mart: On the contrary, I was planning to add an in-memory storage interface to `clientdata`, use it for all of this hook data in `transact`, and then remove all of the manual clears (and remove all associated checks in `start_tx`). 

If we have a working in-memory storage library, I don't think users should have to re-verify it

PR #1056, opened 2025-02-17
make boot tx packing more consistent

James-Mart:


swatanabe (approved):
`TransactionBuilder` in `rpc.rs` should also be updated.

PR #1061, opened 2025-02-18
Reduce r-transact memory consumption

James-Mart:
Decreases the memory requirements of transact by pruning `rawData` out of the transaction trace.

### DIFF
--- a/services/system/Transact/include/services/system/RTransact.hpp
+++ b/services/system/Transact/include/services/system/RTransact.hpp
@@ -83,17 +83,8 @@ namespace SystemService
       std::vector<TraceClientInfo> clients;
    };
    PSIO_REFLECT(TraceClientRow, id, clients)
-   using TraceClientTable = psibase::Table<TraceClientRow, &TraceClientRow::id>;
 
-   // Holds the size of the last transaction trace
-   struct TraceSizeRow
-   {
-      psibase::Checksum256 id;
-      uint64_t             size;
-      auto                 primary_key() const { return psibase::SingletonKey{}; }
-   };
-   PSIO_REFLECT(TraceSizeRow, id, size)
-   using TraceSizeTable = psibase::Table<TraceSizeRow, &TraceSizeRow::primary_key>;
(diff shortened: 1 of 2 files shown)


services/system/Transact/include/services/system/RTransact.hpp
```diff
@@ -83,17 +83,27 @@ namespace SystemService
       std::vector<TraceClientInfo> clients;
    };
    PSIO_REFLECT(TraceClientRow, id, clients)
-
    using TraceClientTable = psibase::Table<TraceClientRow, &TraceClientRow::id>;
 
+   // Holds the size of the last transaction trace
```
swatanabe, line 88: I was expecting just a printf, which will show up in the psinode log (`debug` on success `warning` on failure). This query would not usable on a public endpoint because of the race.

swatanabe: The log on a successful callback is debug level, so it isn't shown by default. It's too noisy to be on by default.

swatanabe: I adjusted my filter to `Severity >= info or Channel = transaction` to see all the callbacks.

services/system/Transact/src/RTransact.cpp
```diff
@@ -56,9 +57,75 @@ std::optional<SignedTransaction> SystemService::RTransact::next()
    return result;
 }
 
+namespace
+{
+   ActionTrace pruneActionTrace(const ActionTrace& at);
```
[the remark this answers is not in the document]
James-Mart: done

services/system/Transact/include/services/system/RTransact.hpp
James-Mart: Done, and you're right - tested, printf works if I adjust logger

swatanabe (approved):
This brings the size down to a much more reasonable level:

| largest boot tx | before | after |
| - | - | - |
| trace size | 4859658 | 4856330 |
| memory usage |29229056 | 4915200 |

PR #1069, opened 2025-02-20
Fix registry account creation

James-Mart:
Found a bug testing `psibase.io` deployment yesterday that prevents staged-tx from going through in workshop.

The reason we didn't catch it previously was because even though the owner of workshop's created apps was set incorrectly, we were only testing while being logged in as the producer account (or we were on an insecure deployment).

This now correctly sets the owner of the new project accounts created in workshop to the intended owner.

### DIFF
--- a/services/system/AuthDelegate/include/services/system/RAuthDelegate.hpp
+++ b/services/system/AuthDelegate/include/services/system/RAuthDelegate.hpp
@@ -1,0 +1,17 @@
+#pragma once
+#include <psibase/Rpc.hpp>
+#include <psibase/Service.hpp>
+#include <psibase/serviceEntry.hpp>
+
+namespace SystemService
+{
+   class RAuthDelegate : public psibase::Service
+   {
+     public:
+      static constexpr auto service = psibase::AccountNumber("r-auth-dlg");
+
+      auto serveSys(psibase::HttpRequest request) -> std::optional<psibase::HttpReply>;
+   };
+   PSIO_REFLECT(RAuthDelegate,  //
+                method(serveSys, request))
+}  // namespace SystemService
--- a/services/system/AuthDelegate/plugin/src/lib.rs
+++ b/services/system/AuthDelegate/plugin/src/lib.rs
@@ -1,24 +1,58 @@
 #[allow(warnings)]
 mod bindings;
+use std::str::FromStr;
+
+use bindings::*;
 mod errors;
 use errors::ErrorType::*;
+use exports::auth_delegate::plugin::api::{Error, Guest as Api};
+use exports::transact_hook_user_auth::{Claim, Guest as HookUserAuth, Proof};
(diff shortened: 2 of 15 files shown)


services/system/AuthDelegate/src/AuthDelegate.cpp
```diff
@@ -78,6 +78,13 @@ namespace SystemService
 
    void AuthDelegate::setOwner(psibase::AccountNumber owner)
    {
+      auto table  = db.open<AuthDelegateTable>();
+      auto record = table.getIndex<0>().get(owner);
+      if (record.has_value())
+      {
+         check(record->owner != getSender(), "circular ownership");
```
swatanabe, line 85: What about a cycle of three or more accounts?

services/system/AuthDelegate/plugin/src/lib.rs
```diff
@@ -1,25 +1,59 @@
...
         _account_name: String,
         _transaction_hash: Vec<u8>,
-    ) -> Result<Option<Proof>, CommonTypes::Error> {
+    ) -> Result<Option<Proof>, Error> {
         Err(NotYetImplemented("get_proofs"))?
     }
 }
 
+impl Api for AuthDelegate {
+    fn set_owner(owner: String) -> Result<(), Error> {
```
swatanabe, line 35: This is unsafe without some kind of auth check.

swatanabe (changes requested):

services/system/AuthDelegate/src/AuthDelegate.cpp
James-Mart: Then they are bricked.
We had nothing before, so this is just intended to be a slight improvement for catching the simplest invalid scenario.

services/system/AuthDelegate/include/services/system/RAuthDelegate.hpp
```diff
@@ -0,0 +1,17 @@
+#pragma once
+#include <psibase/Rpc.hpp>
+#include <psibase/Service.hpp>
+#include <psibase/serviceEntry.hpp>
```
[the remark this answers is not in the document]
James-Mart: Whoops, it's not. Fixed

services/system/AuthDelegate/plugin/src/lib.rs
James-Mart: Of course, thanks - fixed

swatanabe (approved):

services/system/AuthDelegate/plugin/src/lib.rs
```diff
@@ -1,25 +1,76 @@
...
+        use psibase::services::auth_delegate::action_structs::setOwner as set_owner_action;
+
+        assert_caller_admin("new_account");
```
swatanabe, line 49: Did you mean `"set_owner"`?

PR #1078, opened 2025-02-26
Supervisor factory

brandonfancher:
When an app currently instantiates the supervisor, they're given a new, fresh instance. Having multiple instances of Supervisor causes problems. It has so far been up to apps to make sure they instantiate and use one and only one instance. 

This introduces a `SupervisorFactory` for apps to use which will always give them one and only one instance of the Supervisor. I chose this over the class instance singleton pattern, as forcing a class to return a single instance can cause maintenance and testability issues.

Looking for feedback on this approach before I switch all apps to using this.

I confirmed it fixes the errors we were seeing in XAdmin after booting.


Velua:
Looks good to me but can one pass the constructor params here? 
Then again we might be able to remove the options idea anyway...

brandonfancher:
Yeah, I'll enable passing of options if we're okay with this approach.

cool-ant (approved):

services/user/CommonApi/common/packages/common-lib/src/supervisor.ts
```diff
@@ -216,3 +235,5 @@ export class Supervisor {
         iframe.contentWindow.postMessage(message, this.supervisorSrc);
     }
 }
+
+export const getSupervisor = Supervisor.getInstance;
```
James-Mart, line 239: Any reason not to also have `export const supervisor = Supervisor.getInstance();`?

brandonfancher: Because then it eagerly instantiates the supervisor without any options. I tried multiple ways to make that work, but it snarls everything up. I imagine that with some more work on the bundling of common-lib, I could get it to work.

I could call the existing import `supervisor` instead of `getSupervisor` and then an app could just call `supervisor(...)` when ready if you prefer that.

PR #1088, opened 2025-03-04
Fix compilation with recent boost

swatanabe:


James-Mart:
I'm pretty sure I tried this and builds still failed with the latest boost. Will test again though

James-Mart (approved):
Tested and working

PR #1092, opened 2025-03-06
Shorten invite tokens

James-Mart:
Instead of packing a bunch of data in the invite link, most of it is now stored server-side. 
Including the encrypted private key of the invite keypair.

The link now only needs to contain 8 bytes for the seed used to generate the AES key that can decrypt the private key, and another 4 bytes for the ID of the invite (which is a randomly generated u32).

Old invite link:

> http://accounts.psibase.localhost:8070/?token=ACUBAAAMAAwAAAAUAAAAJgAAAAgAAABob21lcGFnZRIAAABodHRwczovL3BzaWJhc2UuaW_xAAAALS0tLS1CRUdJTiBQUklWQVRFIEtFWS0tLS0tCk1JR0hBZ0VBTUJNR0J5cUdTTTQ5QWdFR0NDcUdTTTQ5QXdFSEJHMHdhd0lCQVFRZ1V4WWQxaFRLS0Vza05BOUMKc1ZFTktNbSt5VDNlNUtuVFYyalkwa1dKV2xxaFJBTkNBQVFqQ2xtd3dtQXdCQ1U5MEkyYmY4bFJ2N3lvcFQ1egoxcU9Cd0pjMDk2SnBIbk5tUDlkWGVNVS9PZ1llZVZjQ2U1c2tyNVBtbjNyZzNPZUFxOWxUWDZTUAotLS0tLUVORCBQUklWQVRFIEtFWS0tLS0tCg%3D%3D

New invite link:
> http://accounts.psibase.localhost:8070/?token=AA4AAAAMAFO_zFsvGYDgeukYNQ%3D%3D


services/user/Invite/src/Invite.cpp
```diff
@@ -74,26 +78,28 @@ void Invite::createInvite(Spki inviteKey)
 
    if (settings.whitelist.size() > 0)
    {
-      bool whitelisted = find(settings.whitelist.begin(), settings.whitelist.end(), inviter)
-                         != settings.whitelist.end();
+      bool whitelisted = std::ranges::find(settings.whitelist, inviter) != settings.whitelist.end();
```
swatanabe, line 81: We're on a version of wasi-sdk that has `std::ranges::contains` now.

services/user/Base64/plugin/wit/world.wit
```diff
@@ -0,0 +1,11 @@
+package base64:plugin;
```
swatanabe, line 1: > This encoding may be referred to as "base64url".  This encoding should not be regarded as the same as the "base64" encoding and should not be referred to as only "base64" [RFC 4648](https://www.rfc-editor.org/rfc/rfc4648#section-5)

services/user/Rand/plugin/src/lib.rs
```diff
@@ -0,0 +1,22 @@
+#[allow(warnings)]
+mod bindings;
+use bindings::*;
+
+use exports::rand::plugin::api::Guest as Api;
+use rand::Rng;
+
+struct RandPlugin;
+impl Api for RandPlugin {
+    fn rand_bytes(nr_bytes: u32) -> Vec<u8> {
```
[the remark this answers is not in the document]
James-Mart: The intention of making all of these into separate plugins was to minimize the size of the caller's object code (present and future plugins that depend on this functionality). 

But you're right, for thin wrappers around wasi I think this is not necessary.

services/user/Invite/plugin/src/lib.rs
```diff
@@ -194,9 +201,9 @@ impl Inviter for InvitePlugin {
     fn generate_invite() -> Result<String, CommonTypes::Error> {
         let keypair = keyvault::generate_unmanaged_keypair()?;
 
-        let seed = rand::rand_bytes(8);
+        let seed = rand_bytes(8);
         let encrypted_private_key =
-            aes::with_password::encrypt(&seed, keypair.private_key.as_bytes());
+            aes::with_password::encrypt(&seed, keypair.private_key.as_bytes(), &keypair.public_key);
```
swatanabe, line 206: Is `keypair.public_key` PEM? I'm somewhat concerned that this requires the PEM encoding created by rust to be identical to the encoding created by Botan. DER is more reliable when you need a canonical encoding.

James-Mart: It is PEM. I can switch to der

James-Mart: Done

swatanabe (approved):

PR #1093, opened 2025-03-07
Add query for chain id

James-Mart:
Added query to transact service for the chain ID, and a plugin function to expose it to plugins

### DIFF
--- a/doc/psidk/src/development/front-ends/reference/http-requests.md
+++ b/doc/psidk/src/development/front-ends/reference/http-requests.md
@@ -27,7 +27,8 @@ The [common-api service](../../../default-apps/common-api.md) endpoints which st
 
 | Method | URL                              | Description                                                                                                              |
 | ------ | -------------------------------- | ------------------------------------------------------------------------------------------------------------------------ |
-| `GET`  | `/common/tapos/head`             | Returns [TaPoS](#tapos) for the current head block                                                                       | 
+| `GET`  | `/common/chainid`                | Returns the unique identifier for the chain (Chain ID)                                                                   |
+| `GET`  | `/common/tapos/head`             | Returns [TaPoS](#tapos) for the current head block                                                                       |
 | `GET`  | `/common/thisservice`            | Returns a JSON string containing the service associated with the domain. If it's the root domain, returns `"common-api"` |
 | `GET`  | `/common/rootdomain`             | Returns a JSON string containing the root domain, e.g. `"psibase.127.0.0.1.sslip.io"`                                    |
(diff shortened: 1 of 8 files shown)


services/system/Transact/src/RTransact.cpp
```diff
@@ -622,6 +623,18 @@ std::optional<AccountNumber> RTransact::getUser(HttpRequest request)
    return {};
 }
 
+struct Query
```
swatanabe, line 626: Is transact the right place to put this? Maybe `explore` or even `common` (`/common/tapos/head` is kind of related)

services/system/Transact/src/RTransact.cpp
```diff
@@ -622,6 +623,18 @@ std::optional<AccountNumber> RTransact::getUser(HttpRequest request)
...
+{
+   auto chainId() const
+   {
+      const auto status =
+          psibase::kvGet<psibase::StatusRow>(psibase::StatusRow::db, psibase::statusKey());
+      psibase::check(status.has_value(), "missing status record");
```
swatanabe, line 632: `getStatus`?

services/system/Transact/src/RTransact.cpp
James-Mart: Hah, I had it in explore and moved it here because I thought it was general enough to make sense here. What do you think? `/common/chainid`?

doc/psidk/src/development/front-ends/reference/http-requests.md
```diff
@@ -27,7 +27,8 @@ The [common-api service](../../../default-apps/common-api.md) endpoints which st
 
 | Method | URL                              | Description                                                                                                              |
 | ------ | -------------------------------- | ------------------------------------------------------------------------------------------------------------------------ |
-| `GET`  | `/common/tapos/head`             | Returns [TaPoS](#tapos) for the current head block                                                                       | 
+| `GET`  | `/common/chainid`                | Returns the unique identifier for the chain (Chain ID)                                                                   |
+| `GET`  | `/common/tapos/head`             | Returns [TaPoS](#tapos) for the current head block                                                                       |
 | `GET`  | `/common/thisservice`            | Returns a JSON string containing the service associated with the domain. If it's the root domain, returns `"common-api"` |
```
swatanabe, line 32: This needs to be updated too.

James-Mart: Done

services/system/Transact/src/RTransact.cpp
James-Mart: Done

swatanabe (approved):

PR #1095, opened 2025-03-10
Plugin tester fixes

James-Mart:
This fixes three issues:
1. String inputs couldn't have newline literals, and using "\n" was getting encoded incorrectly (as "\\n") by json stringify. Now the string input is a textarea so we can use actual newlines.
2. Parameter words that needed to be converted to camelCase were, in some cases, resulting in two entries in `paramValues` (one camelCase and one kebab-case). Now only the camelCase variant should be stored.
3. The embed code was serializing the parameters as an object instead of putting the values in an array.

Also added a readme.

### DIFF
--- a/services/user/CommonApi/common/packages/plugin-tester/ui/README.md
+++ b/services/user/CommonApi/common/packages/plugin-tester/ui/README.md
@@ -47,16 +47,6 @@ The Plugin Tester is a development tool designed to help developers test and int
 - `RpcUtils.ts`: Utilities for RPC calls
 - `StringUtils.ts`: String manipulation utilities
 
-## Data Flow
-
-1. The application loads the current service information on startup
-2. Users can specify a service and plugin to load
-3. The schema is fetched using the `usePluginSchema` hook
-4. Available functions are displayed in the `FunctionSelector`
-5. When a function is selected, parameter inputs are generated based on the function's schema
-6. Users can fill in parameter values using type-appropriate input components
-7. The `ExecutionTabs` component handles function execution and displays results
-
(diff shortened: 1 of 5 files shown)


brandonfancher:
@James-Mart There's one more issue that I believe to be a bug that you might want to address in this PR too. On the embed tab, the `params` are displayed as an object. They should be displayed, however, as an array. For example, the tester instructs the dev to use:

```
const response = await supervisor.functionCall({
  service: "accounts",
  plugin: "plugin",
  intf: "api",
  method: "getAccount",
  params: {
  "name": ""
}
});
```

But that doesn't work. It should be:

```
const response = await supervisor.functionCall({
                service: "accounts",
                plugin: "plugin",
                intf: "api",
                method: "getAccount",
                params: ["name"],
            });
```

James-Mart:
> @James-Mart There's one more issue that I believe to be a bug that you might want to address in this PR too. On the embed tab, the `params` are displayed as an object. They should be displayed, however, as an array. For example, the tester instructs the dev to use:
> 
> ```
> const response = await supervisor.functionCall({
>   service: "accounts",
>   plugin: "plugin",
>   intf: "api",
>   method: "getAccount",
>   params: {
>   "name": ""
> }
> });
> ```
> 
> But that doesn't work. It should be:
> 
> ```
> const response = await supervisor.functionCall({
>                 service: "accounts",
>                 plugin: "plugin",
>                 intf: "api",
>                 method: "getAccount",
>                 params: ["name"],
>             });
> ```

Done

brandonfancher (approved):

PR #1096, opened 2025-03-10
Make push_boot an admin API

swatanabe:


James-Mart:
```
[info] [127.0.0.1:53086]: Handled HTTP request: POST x-admin.psibase.localhost:8080/native/admin/push_boot 401 14
```
But I'm using default config and therefore have `admin-authz = rw:loopback`

swatanabe:
> ```
> [info] [127.0.0.1:53086]: Handled HTTP request: POST x-admin.psibase.localhost:8080/native/admin/push_boot 401 14
> ```
> 
> But I'm using default config and therefore have `admin-authz = rw:loopback`

I can't reproduce this

swatanabe:
> 
> I can't reproduce this. I tried both web and cli with a new database.

I also tried with the image-builders http.config

James-Mart (approved):
Resolved. 

Note: specifying admin-authz in the cli will hide any specified in the config file. Therefore if _any_ `admin-authz` options are specified on the cli, then _all_ `admin-authz` options must be.

PR #1099, opened 2025-03-11
Fix progress indicators

swatanabe:
Fix association between labels and transactions, which was broken by #1079

The web boot showed progress by assuming that each transaction installed an equal number of packages, which is very far from the truth. This PR adds a more accurate division of steps, not just package installation, and also divides the install steps into three ranges: completed, in progress, and not started.


James-Mart:
Nice, love it

James-Mart (approved):

Velua (approved):

PR #1100, opened 2025-03-11
Change brotli package to brotli-codec

James-Mart:
This better aligns with the cargo package organization that we establish in our package-template. And fixes a bug preventing the brotli plugin from loading in the plugin-tester


swatanabe:
> And fixes a bug preventing the brotli plugin from loading in the plugin-tester

What bug is this? As of current main (83a6380c0cc138d7b24c420f2cb49c5d08142128) I can load `psi-brotli` `plugin` in the plugin-tester.

James-Mart:
Actually there was no bug, I think when I tested I went to `brotli.psibase/common/plugin-tester` and saw it failed to load. But the problem was just that I was on the `brotli` namespace instead of `psi-brotli`.

So consider this just an improvement to the package name - I think `psi-brotli` is a worse name than `brotli-codec`

swatanabe (approved):

PR #1101, opened 2025-03-12
Add profiles & contacts

Velua:
This introduces the new homepage app `Contacts` and also adds the Profile service + plugin + UI.

# Contacts
Allows the user to create contacts with some simple properties to start with, will add addresses later which could be used to share shipping info. 

# Profile
Added new menu at the bottom left user button there's now a `Edit Profile` button where you can upload a bio, display name and an avatar. This is displayed for other users who add you, however their nickname of you will overwrite the display name you've added for yourself when displayed in Contacts. 

## Remaining tasks
- [x] Make the remaining chainmail components adopt the Two Column component. 
- [ ] Implement the contacts -> chainmail button on the contact detail page. 
- [ ] Implement the contacts -> tokens button on the contact detail page. 
- [x] Delete the templated exampleThing stuff in the Rust package, particularly service.


brandonfancher:
@Velua I added automatic prettier import sorting in the chainmail PR. To get it to work, you likely need to restart VSCode/Cursor. Once you do, you should see the import order change on save/format. I recommend doing that and formatting all the ts/tsx files in this PR...a format commit, if you will.

services/user/Homepage/ui/src/apps/contacts/types.ts
```diff
@@ -0,0 +1,16 @@
...
+export const LocalContact = z.object({
+    account: Account,
+    nickname: z.string().optional(),
+    email: z.string().optional(),
+    phone: z.string().optional(),
+});
+
+export const ProcessedContact = LocalContact.extend({
+    avatar: z.string(),
+});
```
brandonfancher, line 13: What would you think about switching to a convention for zod objects and starting them with lowercase `z`. So it would be `zLocalContact` and `zProcessedContact`. These aren't TypeScript types so it doesn't quite make sense to start them with a capital letter. And because the start with `z`, we'll always know we're dealing with a zod object.

Then we don't have to suffix TS types with `Type` as you're doing below. Types follow normal convention and start with a capital letter.

Thoughts?

services/user/Homepage/ui/src/components/form-profile.tsx
```diff
@@ -0,0 +1,116 @@
...
+    FormItem,
+    FormLabel,
+    FormMessage,
+} from "@/components/ui/form";
+
+const formSchema = z.object({
+    displayName: z
+        .string()
+        .min(2, "Display name must be at least 2 characters"),
+    bio: z.string().max(40, "Bio must be less than 40 characters"),
```
brandonfancher, line 20: Feels like this should be longer, no?

services/user/Homepage/ui/src/components/TwoColumnSelect.tsx
brandonfancher, line 1: I propose we standardize on the use of kebab-case for component (and hook) file names.

services/user/Homepage/ui/src/components/nav-user.tsx
```diff
@@ -111,14 +162,7 @@ export function NavUser() {
...
                                     <AvatarImage
-                                        src={
-                                            chainId && currentUser
-                                                ? generateAvatar(
-                                                      chainId,
-                                                      currentUser,
-                                                  )
-                                                : undefined
-                                        }
+                                        src={avatarSrc}
```
brandonfancher, line 165: An image that's not perfectly square gets stretched in the nav user menu. We should probably use `object-fit: cover;` here and below.

services/user/Homepage/ui/src/apps/contacts/page.tsx
```diff
@@ -0,0 +1,166 @@
...
+                            onChange={(e) => updateSearch(e.target.value)}
+                            className="pl-10"
+                        />
+                        <Search className="absolute left-8 top-1/2 h-4 w-4 -translate-y-1/2 text-muted-foreground" />
+                    </div>
+                    {!contactsData || contactsData.length === 0 ? (
+                        <div className="flex h-full items-center justify-center">
+                            <p className="text-muted-foreground">
+                                No contacts. Create one!
+                            </p>
```
brandonfancher, line 108: I agree that the first contact should be yourself.

services/user/Homepage/ui/src/components/nav-user.tsx
brandonfancher: Other places where this component is used should also `cover` (contacts list, contact page, etc.)

services/user/Homepage/ui/src/apps/contacts/components/contact-details.tsx
```diff
@@ -0,0 +1,266 @@
...
+
+    return (
+        <div>
+            {onBack && (
+                <div className="pl-4">
+                    <Button onClick={onBack} variant="outline" size="icon">
+                        <ArrowLeft />
+                    </Button>
+                </div>
+            )}
```
brandonfancher, line 91: I recommend reusing the same detail header we're using in chainmail's `message-detail.tsx`. It nicely aligns everything and would lend to a more consistent look and feel between the apps.

services/user/Homepage/ui/src/apps/tokens/components/not-logged-in.tsx
brandonfancher, line 1: We already have this component in `services/user/Homepage/ui/src/components/not-logged-in.tsx`. This one may be redundant. But do we even need either of these any more?

brandonfancher (changes requested):
This is awesome. Coming together nicely. If I understand you correctly, this is still in progress. Here's some feedback for you. I'm planning to do a deeper dive tomorrow after another round of review.

services/user/Homepage/ui/src/apps/tokens/components/not-logged-in.tsx
Velua: Removed.

services/user/Homepage/ui/src/apps/contacts/page.tsx
Velua: I've now got a useEffect which automatically creates one if it's not detected.

services/user/Homepage/ui/src/apps/contacts/components/contact-details.tsx
Velua: Adopted

services/user/Homepage/ui/src/apps/contacts/types.ts
Velua: I like it! Adopted with these types, can do another PR to pick up the pre-existing zod schemas later.

services/user/Homepage/ui/src/components/form-profile.tsx
Velua: Just copied X's, they've got 50 for the name and 160 for the bio.

services/user/Homepage/ui/src/apps/chainmail/inbox-page.tsx
```diff
@@ -1,28 +1,47 @@
...
+                    onBack={
+                        display === "right"
+                            ? () => setSelectedMessageId("")
+                            : undefined
+                    }
+                />
+            }
+            header={<MailboxHeader>Inbox</MailboxHeader>}
+            displayMode={display}
+        />
```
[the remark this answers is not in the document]
brandonfancher: The mail list (panel on the left) still appears to be missing its empty state.

services/user/Homepage/ui/src/apps/contacts/components/contact-details.tsx
brandonfancher: Cool. I see it now. We do have an alignment issue though:
image

I think the button in your version on the right hand side might be larger than the ones we use in Chain mail (for example, on a draft email).

services/user/Homepage/ui/src/apps/tokens/components/not-logged-in.tsx
brandonfancher: I see the homepage copy of this file has been deleted. But the one at `services/user/Homepage/ui/src/tokens/components/not-logged-in.tsx` still seems to be there. Is that in use?

services/user/Homepage/ui/src/components/TwoColumnSelect.tsx
brandonfancher: The new hooks in `services/user/Homepage/ui/src/apps/contacts/hooks` can be renamed to kebab-case too.

services/user/Homepage/ui/src/apps/contacts/page.tsx
brandonfancher: Awesome. I like this. But I would expect that editing that special "Myself" contact would open up the my-profile editor instead of treating it like I'm editing the profile information for someone else. I clicked on the "Myself" contact in the contact list and edited it, adding my name, email address and phone number. But when I went to "Edit Profile" from the user menu, none of that information was there. And it was only there that I could change my profile image. The "Myself" contact and my _profile_ should be one-in-the-same, IMHO.

services/user/Homepage/ui/src/components/nav-user.tsx
brandonfancher: Profile image is still stretched on the contact page:
image

brandonfancher (changes requested):
See the comments I unresolved above. Context added.

brandonfancher:
I feel like the "Edit Profile" dialog should go away after I successfully Save Profile.

services/user/Homepage/ui/src/apps/tokens/components/transfer-modal.tsx
```diff
@@ -75,7 +80,11 @@ export const TransferModal = ({
                                 <div>
                                     <img
                                         className="h-20 w-20   border"
```
brandonfancher, line 82: We should `object-cover` on this form too.

services/user/Homepage/ui/src/components/nav-user.tsx
```diff
@@ -201,32 +254,20 @@ export function NavUser() {
                                             <DropdownMenuSubContent>
                                                 {connectedAccounts.map(
                                                     (connectedAccount) => (
-                                                        <DropdownMenuItem
-                                                            disabled={
-                                                                isConnectingToAccount
-                                                            }
+                                                        <AccountMenuItem
```
brandonfancher, line 257: We should filter out the currently logged-in user's account from the "Switch Account" list. If you want to switch to a different account, doesn't make sense to have your current account as an option. Doing that in a separate PR is also justifiable.

brandonfancher (changes requested):

services/user/Homepage/ui/src/apps/contacts/components/contact-details.tsx
Velua: Fixed alignment issue, I can't find a size difference though

Velua:
> I feel like the "Edit Profile" dialog should go away after I successfully Save Profile.

Done

services/user/Homepage/ui/src/apps/chainmail/inbox-page.tsx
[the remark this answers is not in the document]
Velua: Ah right, I got tripped up looking at one the wrong component. 
Appears now.

services/user/Homepage/ui/src/apps/contacts/page.tsx
Velua: > The "Myself" contact and my profile should be one-in-the-same, IMHO.

I somewhat disagree, I like the simplicity of `Profile = Public`, `Contact = Private`, not Contact = Private unless it's yourself then some fields are actually public and other fields are private. 

I still might maintain a different name between them as well, e.g. I edit my contact with the name of "John Williamson" whereas my public name is "JW" and when I share my contact with others through chainmail or something to you, you can get my "John Williamson" version which comes with further sensitive information like my home address, email and mobile phone number. 

> I clicked on the "Myself" contact in the contact list and edited it, adding my name, email address and phone number. But when I went to "Edit Profile" from the user menu, none of that information was there

While it may be disappointing to see those fields are empty, I think it makes the segregation for the user black and white. 

> But I would expect that editing that special "Myself" contact would open up the my-profile editor instead of treating it like I'm editing the profile information for someone else.

That's actually kinda the thing I was aiming for, your contact is dumb and no different than others.

brandonfancher: Okay, I can see how a case can be made for it both ways. I'm fine with that.

brandonfancher (approved):

services/user/Profiles/plugin/src/lib.rs
James-Mart, line 1: FYI, all uses of 
`AccountNumber::from_str(current_user.as_str()).unwrap()`
could be replaced with
`AccountNumber::from(current_user.as_str())`

services/user/Profiles/plugin/src/lib.rs
```diff
@@ -0,0 +1,111 @@
...
+    fn remove_avatar() -> Result<(), Error> {
+        bindings::sites::plugin::api::remove("/profile/avatar")
+    }
+
+    fn get_contacts() -> Result<Vec<Contact>, Error> {
+        let current_user = get_current_user()
+            .expect("Failed to get current user status")
+            .expect("No user logged in");
+
+        let user_table = contact_table::ContactTable::new(AccountNumber::from_str(current_user.as_str()).unwrap());
```
James-Mart, line 53: I think your use of `AccountNumber` and `ContactEntry` in the database are db implementation details. If so, moving conversions to/from these types inside the table interface would significantly simplify the table interface and eliminate boilerplate.

services/user/Profiles/plugin/src/contact_table.rs
```diff
@@ -0,0 +1,98 @@
...
+#[derive(Pack, Unpack, Debug, Default)]
+pub struct ContactEntry {
+    pub account: AccountNumber,
+    pub nickname: Option<String>,
+    pub email: Option<String>,
+    pub phone: Option<String>,
+}
+
+impl ContactEntry {
+    pub fn new(
```
James-Mart, line 23: `From` would be more idiomatic rust than `new` here
```
impl From<Contact> for ContactEntry {
    fn from(contact: Contact) -> Self {
        Self {
            account: AccountNumber::from(contact.account.as_str()),
            nickname: contact.nickname,
            email: contact.email,
            phone: contact.phone,
        }
    }
}
```

services/user/Profiles/plugin/wit/world.wit
James-Mart, line 1: Consider splitting `contacts` from the `api` interface:

```
interface contacts {
    use types.{contact};

    set: func(contact: contact) -> result<_, error>; // Add / update
    remove: func(account: string) -> result<_, error>;
    get: func() -> result<list<contact>, error>;
}
```

services/user/Profiles/plugin/src/lib.rs
```diff
@@ -0,0 +1,111 @@
...
+
+        let current_user = get_current_user()
+            .expect("Failed to get current user status")
+            .expect("No user logged in");
+
+        let user_table = contact_table::ContactTable::new(AccountNumber::from_str(current_user.as_str()).unwrap());
+
+        user_table.remove(account_number)
+    }
+}
```
James-Mart, line 109: If you added a couple helpers (`user()` and `check_valid_caller()`), separated the `contacts` wit interface, and simplified your table interface, I think the *entire* contacts interface would simplify to something like:
```
impl Contacts for ProfilesPlugin {
    fn get() -> Result<Vec<Contact>, Error> {
        check_valid_caller()?;
        Ok(ContactTable::new(user()).get())
    }

    fn set(contact: Contact) -> Result<(), Error> {
        check_valid_caller()?;
        ContactTable::new(user()).set(contact)
    }

    fn remove(account: String) -> Result<(), Error> {
        check_valid_caller()?;
        ContactTable::new(user()).remove(account)
    }
}
```

services/user/Profiles/service/Cargo.toml
```diff
@@ -0,0 +1,27 @@
...
+
+[dependencies]
+psibase = { version = "0.15.0", path = "../../../../rust/psibase" }
+psibase_macros = { version = "0.15.0", path = "../../../../rust/psibase_macros" }
+async-graphql = "7.0.7"
+serde = { version = "1.0.208", features = ["derive"] }
+serde_json = "1.0.120"
+
+[package.metadata.component.target.dependencies]
+"sites:plugin" = { path = "../../Sites/plugin/wit/world.wit" }
```
James-Mart, line 23: You don't need a sites *plugin* dependency in the profiles *service*

services/user/Profiles/plugin/src/contact_table.rs
```diff
@@ -0,0 +1,98 @@
...
+    user: AccountNumber,
+}
+
+impl ContactTable {
+    pub fn new(user: AccountNumber) -> Self {
+        Self { user }
+    }
+
+    fn key(&self) -> String {
+        CONTACTS_KEY.to_string() + "." + &self.user.to_string()
```
James-Mart, line 48: You actually don't even need `CONTACTS_KEY` here, your entire key can just be the current `user`.

services/user/Profiles/plugin/src/lib.rs
```diff
@@ -0,0 +1,111 @@
...
+            display_name: profile.display_name.unwrap_or_default(),
+            bio: profile.bio.unwrap_or_default(),
+            avatar: profile.avatar,
+        }
+        .packed();
+
+        add_action_to_transaction("setProfile", &packed_profile_args).unwrap();
+    }
+
+    fn upload_avatar(file: File, compression_quality: u8) -> Result<(), Error> {
```
James-Mart, line 40: You can make the choice for the compression quality inside this plugin and then you don't need to pass it along to the GUI. I'd probably just use max compression (11).

James-Mart (changes requested):

services/user/Profiles/service/src/lib.rs
```diff
@@ -0,0 +1,57 @@
...
+
+    #[table(name = "ProfileTable", index = 0)]
+    #[derive(Default, Fracpack, ToSchema, SimpleObject, Serialize, Deserialize, Debug, Clone)]
+    pub struct Profile {
+        #[primary_key]
+        pub account: AccountNumber,
+
+        pub display_name: String,
+        pub bio: String,
+        pub avatar: bool,
```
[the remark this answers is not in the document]
Velua: My thinking was that by having this information volunteered it would spare a bunch of network requests from being made over and over again to check. 
If we checked would we throw an array of accounts to sites to check the paths?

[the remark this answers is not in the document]
James-Mart: For simplicity you could just request the profile pictures directly, but gracefully handle the when the pfp doesn't exist. It 
does mean you have to make as many requests as there are profile pictures in contacts, but all assets are cached so images won't be retransferred unless they change.

If you really want to avoid making more requests than necessary, then yeah you could add a gql query to `sites` that checks the existence of multiple files with one request, and then only actually request pfp for files that you know will exist. But tbh this seems like a bit too eager of an optimization to me at this point

services/user/Profiles/plugin/src/lib.rs
```diff
@@ -0,0 +1,102 @@
...
+            bio: profile.bio.unwrap_or_default(),
+            avatar: profile.avatar,
+        }
+        .packed();
+
+        add_action_to_transaction("setProfile", &packed_profile_args).unwrap();
+    }
+
+    fn upload_avatar(file: File) -> Result<(), Error> {
+        bindings::sites::plugin::api::upload(&file, 11)
```
James-Mart, line 94: Maybe this should take a custom type, like:
```
type avatar {
  content-type: string,
  content: list<u8>,
}
```

And then you construct a `sites:file` type internally, that way you force the path to be at "/profile/avatar" and don't leave it up to the caller.

services/user/Profiles/plugin/src/lib.rs
James-Mart, line 1: `setProfile`, `uploadAvatar`, `removeAvatar` still all also need `check_app_sender`

services/user/Profiles/plugin/src/lib.rs
James-Mart: Also consider setting a max image size in here, since this is an image that almost every user will want to upload, we should probably try to encourage these to be as small as possible. It's not necessarily authoritative because people could upload directly to sites to sidestep file size limits enforced in the plugin, but would still help in majority of cases where people set their avatar through using this plugin

services/user/Profiles/plugin/src/lib.rs
James-Mart: Any reason to continue to leave `ContactEntry` exposed outside `contact_table.rs`?

services/user/Profiles/plugin/src/contact_table.rs
```diff
@@ -0,0 +1,93 @@
...
+            .unwrap_or_default()
+    }
+
+    pub fn remove(&self, account: &str) -> Result<(), Error> {
+        let mut contacts = self.get_contacts();
+
+        let account_number = AccountNumber::from_str(&account)
+            .map_err(|_| ErrorType::InvalidAccountNumber(account.to_string()))?;
+
+        match contacts.iter().position(|c| c.account == account_number) {
```
James-Mart, line 85: This works, just FYI there's also `retain`, so if you didn't care about `ContactNotFound`, this `match` could just be 
```
contacts.retain(|c| c.account != account_number);
self.save_contacts(contacts)
```

services/user/Profiles/plugin/src/contact_table.rs
```diff
@@ -0,0 +1,93 @@
...
+        Self {
+            user: AccountNumber::from_str(&user).unwrap(),
+        }
+    }
+
+    fn key(&self) -> String {
+        self.user.to_string()
+    }
+
+    pub fn save_contacts(&self, contacts: Vec<ContactEntry>) -> Result<(), Error> {
```
James-Mart, line 49: `save_contacts` doesn't need to be `pub`

services/user/Profiles/plugin/src/lib.rs
```diff
@@ -0,0 +1,102 @@
...
+impl Api for ProfilesPlugin {
+    fn set_profile(profile: PluginProfile) {
+        let packed_profile_args = profiles::action_structs::setProfile {
+            display_name: profile.display_name.unwrap_or_default(),
+            bio: profile.bio.unwrap_or_default(),
+            avatar: profile.avatar,
+        }
+        .packed();
+
+        add_action_to_transaction("setProfile", &packed_profile_args).unwrap();
```
James-Mart, line 90: Slightly more maintainable if you use `profiles::action_structs::setProfile::ACTION_NAME` instead of hardcoding "setProfile"

services/user/Profiles/plugin/src/lib.rs
```diff
@@ -0,0 +1,102 @@
...
+fn check_app_sender() -> Result<(), ErrorType> {
+    match Client::get_sender_app().app {
+        Some(app) if app == "homepage" => Ok(()),
+        _ => Err(ErrorType::UnauthorizedApp()),
+    }
+}
+
+fn check_account_exists(account: &str) -> Result<(), ErrorType> {
+    get_account(account)
+        .map_err(|_| ErrorType::InvalidAccountNumber(account.to_string()))?
```
James-Mart, line 38: Don't need this `map_err`, you can just use `?`

Velua: It appears `get_account` on an error returns a `Error` not an `ErrorType`, not sure how to tackle this.

James-Mart:
```
fn check_app_sender() -> Result<(), Error> {
    match Client::get_sender_app().app {
        Some(app) if app == "homepage" => Ok(()),
        _ => Err(ErrorType::UnauthorizedApp().into()),
    }
}

fn check_account_exists(account: &str) -> Result<(), Error> {
    get_account(account)?
        .ok_or(ErrorType::NoAccountFound(account.to_string()).into())
        .map(|_| ())
}

fn user() -> Result<String, Error> {
    get_current_user()?.ok_or(ErrorType::NoUserLoggedIn().into())
}
```

services/user/Profiles/service/src/lib.rs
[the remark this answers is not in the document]
Velua: Removed the boolean concept, instead we do a fetch request for the image, should it fail then the identicon is rendered instead.

James-Mart (approved):
Looking great

PR #1107, opened 2025-03-17
Misc perms oauth v2 fixes

James-Mart:


cool-ant (approved):
I purposely removed the bool, heading toward the _or_ask() version of the function, but I'll just merge this and re-swap that when I create the 2nd method and rename both.
This all looks good. Thank you.

PR #1108, opened 2025-03-17
Improve invite app

James-Mart:
This PR was primarily about adding various invite service improvements:
- Use u32 id as invite primary key
- Separate invite query handling into r-invite
- Eliminate deprecated white/blacklist invite functionality
- Simplify event emission
- Add graphql historical invite event queries

In order to do so, I also added the same `EventQuery` functionality to cpp graphql libraries as was added to rust [here](https://github.com/gofractally/psibase/pull/1013).


services/user/Invite/src/Invite.cpp
```diff
@@ -41,78 +41,66 @@ void Invite::init()
...
-      bool whitelisted = std::ranges::contains(settings.whitelist, inviter);
-      check(whitelisted, onlyWhitelisted.data());
-   }
-   else if (settings.blacklist.size() > 0)
-   {
-      bool blacklisted = std::ranges::contains(settings.blacklist, inviter);
-      check(not blacklisted, noBlacklisted.data());
-   }
+   auto     idx      = inviteTable.getIndex<0>();
+   uint32_t inviteId = (idx.isEmpty()) ? 0 : (*(--idx.end())).inviteId + 1;
```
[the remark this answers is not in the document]
James-Mart: Fixed

services/user/Invite/include/services/user/InviteTables.hpp
```diff
@@ -9,33 +9,34 @@ namespace UserService
...
+         static constexpr std::string_view rejected       = "rejected";
+         static constexpr std::string_view deleted        = "deleted";
+         static constexpr std::string_view deletedExpired = "deletedExpired";
+      };
+
       /// An invite object
       struct InviteRecord
       {
+         /// Monotonically increasing ID of the invite
+         uint32_t inviteId;
```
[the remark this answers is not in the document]
James-Mart: I did not address this. I think it's not a bad idea but I'd like to address in a separate PR. The invite service has other improvements coming anyway

libraries/psibase/common/include/psibase/serveGraphQL.hpp
```diff
@@ -871,134 +874,316 @@ namespace psibase
...
+      from_json_object(stream,
+                       [&](std::string_view key)
+                       {
+                          if (key == "rowid")
+                          {
+                             std::string rowid_str;
+                             from_json(rowid_str, stream);
+                             auto [ptr, ec] = std::from_chars(
+                                 rowid_str.data(), rowid_str.data() + rowid_str.size(), s.rowid);
+                             check(ec == std::errc{}, "Invalid rowid");
```
swatanabe, line 1175: Does `from_json(s.rowid, stream)` not work?

swatanabe (approved):

libraries/psibase/common/include/psibase/serveGraphQL.hpp
James-Mart: Doesn't this way make more sense since the json has the `rowid` value quoted? Or would you expect `from_json` to still be able to extract that into a `uint64_t`?

James-Mart: Doesn't this way make more sense because the value of `rowid` is quoted? Or would you expect `from_json` to still be able to extract that into a `uint64_t`?

swatanabe: `from_json` for integer types can always parse a string.

PR #1109, opened 2025-03-18
Package template updates

James-Mart:
* use new EventQuery class for paginated event query
* use supervisor singleton
* remove explicit login call


James-Mart:
@brandonfancher Given that it's a singleton now, wouldn't it be more expected to just use `getSupervisor()` on any page that requires it, rather than import from main.tsx? 

If not I'll revert those changes, it just made sense to me in the moment

brandonfancher:
> @brandonfancher Given that it's a singleton now, wouldn't it be more expected to just use `getSupervisor()` on any page that requires it, rather than import from main.tsx? 

Yeah, for sure. Import at will!

brandonfancher (approved):
LGTM

PR #1114, opened 2025-03-25
Subgroups service

James-Mart:
A service for subgrouping algorithms. I added one I called, `gmp` or "Greatest Minimum Partition".

Given a population and a set of allowed group sizes, it generates a grouping that optimizes for maximizing the size of the minimum constructed partition.

Example output from my unit test, using fixed allowed group sizes (4,5,6) and varying the total population size from 10-30.

| Population | Groups (4,5,6)  |
|------------|-----------------|
|     10     | 5, 5            |
|     11     | 6, 5            |
|     12     | 6, 6            |
|     13     | 5, 4, 4         |
|     14     | 5, 5, 4         |
|     15     | 5, 5, 5         |
|     16     | 6, 5, 5         |
|     17     | 6, 6, 5         |
|     18     | 6, 6, 6         |
|     19     | 5, 5, 5, 4      |
|     20     | 5, 5, 5, 5      |
|     21     | 6, 5, 5, 5      |
|     22     | 6, 6, 5, 5      |
|     23     | 6, 6, 6, 5      |
|     24     | 6, 6, 6, 6      |
|     25     | 5, 5, 5, 5, 5   |
|     26     | 6, 5, 5, 5, 5   |
|     27     | 6, 6, 5, 5, 5   |
|     28     | 6, 6, 6, 5, 5   |
|     29     | 6, 6, 6, 6, 5   |
|     30     | 6, 6, 6, 6, 6   |

## New implementation

In the new implementation, I made it simpler for myself to define preference criteria for the solutions, so I added one more criteria.

When there are multiple solutions for a given population, prefer the new solution if:
* The minimum group size is greater than the old minimum group size (greatest minimum partition), or
* The solution with the minimum number of minimum-sized groups

In the old implementation, for example, population 14 with allowed_groups: [4,5,6] yielded the result: [6,4,4]
In the new implementation, the same parameters yield the result: [5,5,4]

The latter was chosen because it likely has better economic properties when used in the fractal democracy context.


services/user/Subgroups/service/src/lib.rs
```diff
@@ -0,0 +1,192 @@
+#[psibase::service(name = "subgroups")]
+pub mod service {
+    pub use psibase::fracpack;
+    use std::collections::HashMap;
+
+    /// # Description
+    ///
+    /// Recursive function that finds a valid partition of `n` into groups with sizes from `allowed`
+    /// Uses memoization to avoid redundant calculations
```
swatanabe, line 9: dynamic programming is more efficient

James-Mart: Not sure what you're looking for, but I would propose if efficiency is the only concern that we proceed since this has the desired properties, and we can optimize more later

swatanabe: I suspect it would also be simpler:

```rust
for i in 0..=n {
  groups[i] = allowed.iter().map(|g| append_solution(&groups[i - g], g)).max_by(choose_best_solution);
}
```

James-Mart: Hmm, okay, I think I understand and agree it's better. I'll give it a shot

James-Mart: Okay, that ended up being an easier refactor than I thought. I like this implementation much more.

services/user/Subgroups/service/src/lib.rs
```diff
@@ -1,98 +1,72 @@
...
-    /// # Description
-    ///
-    /// Recursive function that finds a valid partition of `n` into groups with sizes from `allowed`
-    /// Uses memoization to avoid redundant calculations
+    /// Function that determines which subgrouping solution is preferred based on specific criteria.
+    /// Returns true if the new solution is preferred over the old solution.
+    type PreferenceFunction = fn(new: &[u32], old: &[u32]) -> bool;
+
+    /// Preference function that compares based on the greatest minimum group size.
+    /// If minimum values are equal, it compares based on maximum group size.
```
swatanabe, line 10: What about sorting in ascending order and then comparing lexicographically.

swatanabe: Which has the same sort order as a tuple of the counts of groups of each size.

James-Mart: Elegant

services/user/Subgroups/service/src/lib.rs
```diff
@@ -0,0 +1,104 @@
...
+
+        criterion1 || criterion2
+    }
+
+    /// Finds a valid partition of `n` into groups of `allowed` sizes
+    ///
+    /// Parameters
+    /// * `n` - The number to partition (population)
+    /// * `allowed` - Vector of allowed group sizes
+    /// * `is_preferred` - Function that determines which solution is preferred
```
swatanabe, line 36: The requirements for `is_preferred` are
- It must define a strict weak ordering for partitions of each total size, up to `n`.
- Adding the same group to two partitions must not change the result of `is_preferred`

services/user/Subgroups/service/src/lib.rs
```diff
@@ -0,0 +1,104 @@
...
+    /// Achieves by comparison based on lexicographical ordering.
+    /// Returns true if new > old.
+    fn greatest_minimum_preference(new: &[u32], old: &[u32]) -> bool {
+        let mut new_sorted = new.to_vec();
+        let mut old_sorted = old.to_vec();
+
+        new_sorted.sort();
+        old_sorted.sort();
+
+        for (n, o) in new_sorted.iter().zip(old_sorted.iter()) {
```
swatanabe, line 19: https://doc.rust-lang.org/std/vec/struct.Vec.html#impl-Ord-for-Vec%3CT,+A%3E

services/user/Subgroups/service/src/lib.rs
James-Mart: Added these to the doc comment

services/user/Subgroups/service/src/lib.rs
James-Mart: Oh, nice. Simplified

swatanabe (approved):

services/user/Subgroups/service/src/lib.rs
```diff
@@ -0,0 +1,98 @@
...
+    /// * `allowed` - Vector of allowed group sizes
+    /// * `is_preferred` - Function that determines which solution is preferred
+    ///
+    /// Returns:
+    /// * `Some(Vec<u32>)` - A valid partition of `n` using allowed group sizes.
+    ///                      When >1 solutions exist for the same population size,
+    ///                      returns the solution preferred according to `is_preferred`
+    /// * `None` - No valid partition exists
+    fn partition(n: u32, allowed: &Vec<u32>, is_preferred: PreferenceFunction) -> Option<Vec<u32>> {
+        let mut solutions = vec![None; (n as usize) + 1];
```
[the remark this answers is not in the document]
James-Mart: I added this as a note into the code so maybe it can be addressed in the future

services/user/Subgroups/service/src/lib.rs
```diff
@@ -0,0 +1,98 @@
...
+    /// * `allowed_groups` - A vector of allowed group sizes
+    ///
+    /// Returns:
+    /// * `Some(Vec<u32>)` - A valid partition of the population into allowed group sizes.
+    /// * `None` - No valid partition exists
+    #[action]
+    fn gmp(population: u32, allowed_groups: Vec<u32>) -> Option<Vec<u32>> {
+        let mut allowed = allowed_groups.clone();
+        allowed.dedup();
+        allowed.sort_unstable();
```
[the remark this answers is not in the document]
James-Mart: Fixed

Issue #1074, opened 2025-02-22
Rework oauth mechanism


James-Mart:
Browsers detect the popup from the supervisor domain as coming from their parent window origin.

Original comment:
---
Uh oh. In testing this I noticed that the popup is detected as coming from the parent window instead of from supervisor.psibase (the domain actually responsible)

![image](https://github.com/user-attachments/assets/dcfb2675-8c7d-4bee-8322-5309b4e24933)

Is this also what you saw @cool-ant ?

_Originally posted by @James-Mart in https://github.com/gofractally/psibase/issues/1038#issuecomment-2673046181_
  

------

```mermaid
sequenceDiagram
    participant exchange
    participant tokens as tokens plugin
    participant permissions as permissions plugin
    participant supervisor as host:common
    participant /oauth as supervisor.psibase/oauth
    participant /permissions as permissions.psibase/permissions

    exchange->>tokens: transfer a token
    tokens->>permissions: isPermitted(exchange)?
    note over permissions: kvStore ID: { exchange, tokens, expiry... } 
    permissions->>supervisor: oauth request (/permissions)
    Note over supervisor: kvStore ID: { permissions, /permissions }
    supervisor->>exchange: needs redirect to<br>supervisor.psibase/oauth?id=id
    Note over exchange: user click to redirect<br>to supervisor.psibase/oauth?id=id
    Note over exchange,/oauth: REDIRECT
    /oauth-->>supervisor: <onload> Pass ID
    Note over supervisor: check ID, get {permissions, /permissions}
    supervisor-->>/oauth: Redirect to permissions.psibase/permissions
    note over /oauth,/permissions: REDIRECT
    /permissions-->>permissions: Store permissions
```

In this new strategy, the root window app is responsible to redirect to the page that handles the permissions. Or more precisely, they redirect to a supervisor subpage, which then reads its own localstorage and redirects again (double redirect) to the final oauth page. 

This design does not trigger the popup blocker, and continues to satisfy the constraint that  the root app does not get to know the details of any downstream oauth requests.

*If* we felt it was okay to drop the constraint that the root app does not get to know any details about downstream oauth requests, then it would be possible to simplify this design down to a single redirect directly from the root app to the permissions oauth page.

James-Mart:
V1.0 of this rework is now finished with #1117. Still to-do: make x-platform.

PR #1117, opened 2025-03-25
OAuth flow v0.1

cool-ant:
Addressing #1074 more completely


services/user/Permissions/ui/src/permissions/App.tsx
```diff
@@ -1,83 +1,110 @@
...
     const thisServiceName = "permissions";
-    const [isLoading, setIsLoading] = useState<boolean>(true);
-    const [params, setParams] = useState<any>({});
-    const [isValidPermRequest, setIsValidPermRequest] = useState<any>({});
-
-    const init = async () => {
+    const [isLoading, setIsLoading] = useState(true);
+    const [validPermRequest, setValidPermRequest] = useState<any>(null);
+    const [error, setError] = useState<string | null>(null);
+    const initApp = async () => {
```
brandonfancher, line 14: I'd wrap `initApp` in a `useCallback()` so it's not re-created on each render (unless its dependencies change). Or alternatively, move it into the `useEffect()` right above where you call it. That would have a similar result.

services/user/Supervisor/ui/src/oauth/App.tsx
```diff
@@ -0,0 +1,77 @@
...
+        urlParams.get("returnUrlPath") || "",
+    );
+    return newIframeUrl.toString();
+};
+
+export const App = () => {
+    const [iframeUrl, setIframeUrl] = useState<string | null>(null);
+    const [error, setError] = useState<string | null>(null);
+
+    const initApp = async () => {
```
brandonfancher, line 26: Same here. I'd wrap `initApp` in a `useCallback()` so it's not re-created on each render (unless its dependencies change). Or alternatively, move it into the `useEffect()` right above where you call it. That would have a similar result.

services/user/Supervisor/ui/src/oauth/App.tsx
```diff
@@ -0,0 +1,77 @@
...
+                <iframe
+                    src={iframeUrl}
+                    style={{
+                        width: "50%",
+                        height: "50vh",
+                        border: "none",
+                    }}
+                />
+            )}
+        </div>
```
brandonfancher, line 75: I recommend using `dvh` units here instead of `vh` units, in order to better support mobile browsers with dynamic viewport heights.

services/user/Permissions/ui/src/permissions/App.tsx
```diff
@@ -1,83 +1,110 @@
...
         await supervisor.onLoaded();
```
brandonfancher, line 15: Just an FYI: Instead of exporting this from `perms_main` and importing it into this file, you can call `const supervisor = getSupervisor();` in this file and you'll get the same instance now every time.

brandonfancher (changes requested):

services/user/Supervisor/ui/src/plugin/pluginHost.ts
```diff
@@ -215,16 +225,58 @@ export class PluginHost implements HostInterface {
     getAppUrl(app: string): string {
         return `${siblingUrl(null, app)}`;
     }
+    
+    private setActiveUserPrompt(subpath: string | undefined, upPayloadJsonStr: string): Result<string, RecoverableErrorPayload> {
+        let up_id = uuidv4();
```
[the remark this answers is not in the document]
cool-ant: nice

services/user/Supervisor/ui/src/oauth/db.ts
```diff
@@ -0,0 +1,65 @@
...
+            throw new Error("No active oauth request found");
+        }
+
+        const oauthReq = JSON.parse(new TextDecoder().decode(oauthReqBytes));
+        if (id != oauthReq.id) {
+            await this.delete();
+            throw new Error("Oauth request id mismatch");
+        }
+
+        let isExpired = (new Date().getUTCSeconds()
```
Velua, line 42: Are you sure you want [`getUTCSeconds`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/getUTCSeconds) ? I think you want [`getTime`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/getTime) instead which will return milliseconds, can do `Math.floor(new Date().getTime() / 1000)` for seconds. `getUTCSeconds` is only going to return `0-59`

services/user/Supervisor/ui/src/oauth/db.ts
```diff
@@ -0,0 +1,65 @@
...
+
+export const isTypeOauthRequest = (
+    oauthReqRes: OauthRequest | RecoverableErrorPayload,
+): oauthReqRes is OauthRequest => {
+    return (
+        "subdomain" in oauthReqRes &&
+        "subpath" in oauthReqRes &&
+        "expiry_timestamp" in oauthReqRes
+    );
+};
```
Velua, line 22: There's a lot of `as` use here, wondering if it's worth using zod
```suggestion
const zOauthRequest = z.object({
    id: z.string(),
    subdomain: z.string(),
    subpath: z.string(),
    expiry_timestamp: z.number(),
})

export const isTypeOauthRequest = (
    oauthReqRes: OauthRequest | RecoverableErrorPayload,
): oauthReqRes is OauthRequest => {
    return zOauthRequest.safeParse(oauthReqRes).success
};
```

Velua (changes requested):
I've marked this as Request changes given the `getUTCSeconds` thing to ensure it's checked.

services/user/Supervisor/ui/src/oauth/db.ts
cool-ant: Great catch. Thanks. It hit my brain like seconds-since-the-epoch, and I didn't catch that.

services/user/Supervisor/ui/src/oauth/db.ts
cool-ant: Done. Thanks for spearheading our zodification @Velua

services/user/Permissions/ui/src/permissions/App.tsx
```diff
@@ -1,83 +1,111 @@
...
-    const [isLoading, setIsLoading] = useState<boolean>(true);
-    const [params, setParams] = useState<any>({});
-    const [isValidPermRequest, setIsValidPermRequest] = useState<any>({});
-
-    const init = async () => {
+    const [isLoading, setIsLoading] = useState(true);
+    const [validPermRequest, setValidPermRequest] = useState<any>(null);
+    const [error, setError] = useState<string | null>(null);
+    const initApp = useCallback(async () => {
         await supervisor.onLoaded();
```
James-Mart, line 16: awaiting `onLoaded` is no longer necessary either. Anything that calls the supervisor impl internally already does the await onLoaded

services/user/Supervisor/ui/src/oauth/db.ts
```diff
@@ -0,0 +1,59 @@
...
+            throw new Error("Oauth request id mismatch");
+        }
+
+        let isExpired = (Math.floor(new Date().getTime() / 1000)
+            - oauthReq.expiry_timestamp.seconds)
+            >= OAUTH_REQUEST_EXPIRATION;
+        if (isExpired) {
+            await this.delete();
+            throw new Error("Oauth request expired");
+        }
```
Velua, line 42: This seems strange as the expiry_timestamp is more, `created_at`.
I think it would be more common to treat this more as a traditional expiry timestamp and then add the `OAUTH_REQUEST_EXPIRATION` when this is made rather than assessed. 

```suggestion
        let isExpired = (Math.floor(new Date().getTime() / 1000)
            >= oauthReq.expiry_timestamp.seconds);
           if (isExpired) {
            await this.delete();
            throw new Error("Oauth request expired");
        }
```

services/user/Supervisor/ui/src/plugin/pluginHost.ts
```diff
@@ -215,16 +227,58 @@ export class PluginHost implements HostInterface {
...
+    
+    private setActiveUserPrompt(subpath: string | undefined, upPayloadJsonStr: string): Result<string, RecoverableErrorPayload> {
+        let up_id = window.crypto.randomUUID?.() ?? Math.random().toString();
+        
+        // In Supervisor localStorage space, store id, subdomain, and subpath
+        const supervisorUP = SupervisorPromptPayload.parse({
+            id: up_id,
+            subdomain: this.self.app,
+            subpath: subpath,
+            expiry_timestamp: Math.floor(new Date().getTime() / 1000),
```
Velua, line 239: If the other comment is followed 
```suggestion
            expiry_timestamp: Math.floor(new Date().getTime() / 1000) + OAUTH_REQUEST_EXPIRATION,
```

Velua (changes requested):

services/user/Permissions/plugin/wit/world.wit
```diff
@@ -7,7 +7,8 @@ interface api {
 
 interface users {
     use host:common/types.{error};
-    is-permitted: func(caller: string) -> result<bool, error>;
+    is-auth-or-prompt: func(caller: string) -> result<bool, error>;
+    is-auth: func(caller: string) -> result<bool, error>;
 }
 
 world imports {
```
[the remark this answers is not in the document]
James-Mart: You want auth-sig to call `save-permission`? That's the only function in your `api` interface.

services/user/Permissions/ui/src/permissions/App.tsx
```diff
@@ -1,83 +1,111 @@
...
+            throw e;
         }
-        window.close();
+        ActivePermsOauthRequest.delete();
+        followReturnRedirect();
     };
     const deny = () => {
-        window.close();
+        ActivePermsOauthRequest.delete();
+        followReturnRedirect();
```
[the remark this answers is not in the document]
cool-ant: Yes, execution context will be interrupted on page nav. Great catch. Thanks.

services/user/Permissions/ui/src/permissions/db.ts
```diff
@@ -0,0 +1,36 @@
...
+            intf: "keyvalue",
+            method: "get",
+            params: [OAUTH_REQUEST_KEY]});
+
+        if (!oauthReqBytes) {
+            throw new Error("No active oauth request found");
+        }
+
+        const oauthReq = JSON.parse(new TextDecoder().decode(oauthReqBytes));
+        return oauthReq as PermsOauthRequest;
```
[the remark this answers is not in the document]
cool-ant: Weird. this wasn't in my search for " as " when I did the zodification. another good catch.

services/user/Supervisor/ui/src/plugin/pluginHost.ts
```diff
@@ -22,6 +25,15 @@ interface HttpResponse {
     body: string;
 }
 
+const SupervisorPromptPayload = z.object({
+    id: z.string().uuid(),
```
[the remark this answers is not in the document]
cool-ant: too bad we don't have a strong type system for randomUUID() and Math.random() to use so we don't have problems like this. ;)

services/user/Permissions/plugin/wit/world.wit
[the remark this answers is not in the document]
cool-ant: Ah. Sorry. the github window showed the wrong code, and I assumed I was looking at the code you were referenced. This makes more sense now. thx

services/user/Permissions/plugin/wit/world.wit
```diff
@@ -2,15 +2,15 @@ package permissions:plugin;
...
-    save-permission: func(caller: string, callee: string) -> result<_, error>; // only option in v1 is fullAccess
-}
-
-interface users {
-    use host:common/types.{error};
-    is-permitted: func(caller: string) -> result<bool, error>;
+    /// @description Check if the caller is permitted access to `callee`'s privileged methods
+    /// and if not already, prompt user to request such access.
+    /// @param caller The caller, requesting access
+    is-auth-or-prompt: func(caller: string) -> result<bool, error>;
```
James-Mart, line 8: Not necessary to address this comment in this PR (I won't hold up the PR for this), 

After a little more thought, I'm thinking maybe a better API would be:
* is-auth 
* prompt

So `is-auth` stays the same, but the other function, instead of `is-auth-or-prompt` just becomes a `prompt` function. The caller, then, is responsible for prompting at will: 
```rust
if permissions.is_auth(...) != Ok(true) {
    permissions.prompt(...)
}
```

This has the added benefit of allowing the caller to `prompt` regardless of the is-auth status (which allows the UI to allow the user to reconfigure their permissions even if already granted).

Thoughts?

cool-ant: Well...
In the not-Permissions scenario, the use has Supervisor::prompt_user() it can call directly, right?
For the Permissions scenario, I wouldn't call it just prompt cuz it's not that generic. I suppose prompt_auth()?
Lastly, we actually *don't* have a flow for removing an auth. Shall I add that to the Deny button? I think that's a super noninvasive way to get this done. No change at all to the current functionality on a first-time prompt, but it'll clear the auth grant if it has previously been made. The only corner case that leaves is the "shit, I didn't want to even go here, let me cancel, which of course the user always has the browser back button for.

services/user/Permissions/plugin/src/db.rs
```diff
@@ -1,72 +1,21 @@
...
+use crate::bindings::accounts::plugin::api as Accounts;
+use crate::bindings::clientdata::plugin::keyvalue as Keyvalue;
 
 pub struct AccessGrants;
 
 impl AccessGrants {
     pub fn set(caller: &str, callee: &str) {
-        Keyvalue::set(&format!("{callee}<-{caller}"), &"1".as_bytes())
+        let any_value = "1".as_bytes();
+        let user = Accounts::get_current_user()
```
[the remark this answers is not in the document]
James-Mart: Yeah, at the time of the prompt, you have to store the current user in localstorage as part of the payload that gets read by permissions page

services/user/Permissions/ui/src/permissions/db.ts
```diff
@@ -0,0 +1,40 @@
...
+import { z } from "zod";
+
+const supervisor = getSupervisor();
+
+const PermsOauthRequest = z.object({
+    id: z.string(),
+    user: z.string(),
+    caller: z.string(),
+    callee: z.string(),
+});
```
brandonfancher, line 12: @Velua and I agreed to start using a `z` prefix for zod schemas: `zPermsOauthRequest`. That way, if it's exported/passed around, it's easy to identify what is the type and what is the zod schema.

services/user/Permissions/ui/src/permissions/App.tsx
```diff
@@ -1,83 +1,125 @@
...
+            if (e instanceof Error) {
+                setError(e.message);
+            } else {
+                setError("Unknown error saving permission");
+            }
             console.error("error saving permission: ", e);
+            throw e;
         }
-        window.close();
+        await ActivePermsOauthRequest.delete();
```
brandonfancher, line 83: Should this be in the `try` block...to catch errors?

services/user/Permissions/ui/src/permissions/App.tsx
```diff
@@ -1,83 +1,125 @@
...
+            intf: "admin",
+            method: "delPerm",
+            params: [
+                validPermRequest?.user,
+                validPermRequest?.caller,
+                validPermRequest?.callee,
+            ],
+        });
+        await ActivePermsOauthRequest.delete();
+        followReturnRedirect();
```
brandonfancher, line 98: Will this throw errors that should be handled?

services/user/Supervisor/ui/src/oauth/main.tsx
```diff
@@ -0,0 +1,16 @@
+import React from "react";
+import ReactDOM from "react-dom/client";
+import { QueryClient, QueryClientProvider } from "@tanstack/react-query";
```
brandonfancher, line 3: Is `@tanstack/react-query` being used anywhere in this app? Or can it be removed?

services/user/Supervisor/ui/src/oauth/db.ts
```diff
@@ -0,0 +1,57 @@
+import { Result } from '../hostInterface';
+import { RecoverableErrorPayload } from '../plugin/errors';
+import { OAUTH_REQUEST_KEY } from '../constants';
+import { getSupervisor } from "@psibase/common-lib";
+import { z } from 'zod';
+
+export const supervisor = getSupervisor();
+
+export const OauthRequestSchema = z.object({
```
brandonfancher, line 9: `zOauthRequest`

services/user/Supervisor/ui/src/oauth/App.tsx
```diff
@@ -0,0 +1,75 @@
...
+        }
+
+        try {
+            const oauthReqUnchecked = await ActiveOauthRequest.get(oauth_id);
+            const oauthReq = OauthRequestSchema.parse(oauthReqUnchecked);
+            setIframeUrl(buildIframeUrl(oauthReq));
+        } catch (e) {
+            setError("OAuth request error: " + e);
+        } finally {
+            await ActiveOauthRequest.delete();
```
brandonfancher, line 41: Can this throw?

services/user/Supervisor/ui/src/oauth/db.ts
brandonfancher: Parsing with this schema as is will pass with empty strings. Is that intentional or do you want to make a min length required?

services/user/Supervisor/ui/src/plugin/pluginHost.ts
```diff
@@ -22,6 +25,15 @@ interface HttpResponse {
     body: string;
 }
 
+const SupervisorPromptPayload = z.object({
```
brandonfancher, line 28: `zSupervisorPromptPayload` and `.min(1)` instead of `.nonempty()`. According to the docs, the latter is for validating arrays; it seems to be deprecated for strings.

services/user/Supervisor/ui/src/plugin/pluginHost.ts
```diff
@@ -215,16 +227,66 @@ export class PluginHost implements HostInterface {
     getAppUrl(app: string): string {
         return `${siblingUrl(null, app)}`;
     }
+    
+    private setActiveUserPrompt(subpath: string | undefined, upPayloadJsonStr: string): Result<string, RecoverableErrorPayload> {
```
brandonfancher, line 231: I'm trying to understand this better. Are we not worried about awaiting these calls and any unhandled promise rejection errors in this method?

brandonfancher (changes requested):

services/user/Supervisor/ui/src/oauth/App.tsx
cool-ant: yes it can. great catch. 🙏

brandonfancher (approved):

services/user/CommonApi/common/packages/wit/host-common.wit
```diff
@@ -99,20 +99,18 @@ interface client {
     /// Gets the full domain origin of the service from which the
     /// currently running plugin was fetched.
     my-service-origin: func() -> string;
+    
+    /// `prompt-user` prompts the user for input. `prompt-user` will display the content at `subpath`, supplying `payload-json-str` as ?????
```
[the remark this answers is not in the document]
cool-ant: Yeah, new JS operator if you have *no* idea what the incoming data type is, leaving you completely confused..

PR #1118, opened 2025-03-26
update builder image to the latest

James-Mart:


swatanabe:
> Interesting
> 
> > Error: Unable to create cache directory

Maybe an update to sccache that needs some tweaks to the environment variables?

swatanabe (approved):

PR #1123, opened 2025-04-01
Fractal spec v0.1 proposal

James-Mart:
Addresses my own comments in #1122 . 

Other smaller misc updates: 
* shorter names
* isolating service/rpc/plugin

In general, this proposal is consistent with the idea that the first thing we should implement is just a service that tracks membership, facilitates evaluations, tracks reputation over time, and distributes a token based on reputation


services/user/fractally-world.wit
```diff
@@ -1,84 +1,83 @@
...
 
-interface fractally {
-    use fractally-types.{config};
-    set-config: func(config: config) -> result<void, error>;
-    get-fractal-info: func() -> result<fractal-info, error>;
-    // I propose the periodic subj evaluation process is managed via FSM with this method to nudge it when needed (rather than numerous such do-the-next-thing methods)
-    nudge: func() -> result<void, error>;
-}
+interface fractal {
+    use types.{config};
```
cool-ant, line 20: you don't define `config` above. Is there any config? Not sure after all the deletion that there's anything for fractal config

James-Mart: no longer applicable

services/user/fractally-world.wit
```diff
@@ -1,84 +1,83 @@
...
-        avatar: string;
-        created-at: timestamp;
-        status: enum<pre-evaluations, active, inactive>;
-        num-members: u64;
+
+    record member {
+        account: string, // primary key
+        name: string,
+        joined-at: timestamp,
+        reputation: u32,
```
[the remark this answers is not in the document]
James-Mart: Yeah joined-at we should keep in state so it doesn't get pruned

services/user/Fractal/doc/fractal-service.wit
```diff
@@ -0,0 +1,78 @@
...
+    @since(feature = scheduled-meetings)
+    record meeting-config {
+        // @velua, you spent plenty of time on this for the Psinq UI. Would you fill in the minimal set of details we determined were necessary for meeting schedule?
+    }
+
+    record fractal {
+        service: account, // Primary key
+        created-at: timepointsec,
+        name: string,
+        mission: string,
```
cool-ant, line 11: part of the profile service? I think this is the equivalent of "bio"

services/user/Fractal/doc/fractal-service.wit
```diff
@@ -0,0 +1,78 @@
+interface types {
+    @since(feature = scheduled-meetings)
+    record meeting-config {
+        // @velua, you spent plenty of time on this for the Psinq UI. Would you fill in the minimal set of details we determined were necessary for meeting schedule?
+    }
+
+    record fractal {
+        service: account, // Primary key
+        created-at: timepointsec,
+        name: string,
```
cool-ant, line 10: also part of profile service (and we proposed noting things that will come from the profile service as comments)

services/user/Fractal/doc/fractal-service.wit
```diff
@@ -0,0 +1,78 @@
...
+    // Titles
+    //
+    // candidate, u32(1) << 0; 
+    // @since(feature = member-fractals)
+    // ambassador, u32(1) << 1;
+    record member {
+        account: account, // primary key
+        @since(feature = invites)
+        inviter: option<account>,
+        // display name, bio, avatar - stored in `sites` service
```
cool-ant, line 25: I think this exact comment would work in fractal record above

cool-ant (approved):
just some minor changes. Otherwise, :shipit:

services/user/Fractal/doc/fractal-service.wit
```diff
@@ -0,0 +1,78 @@
...
+
+interface member {
+    join-waitlist: func(fractal: account);
+    config-candidacy: func(fractal: account, enable: bool);
+
+    @since(feature = invites)
+    invite: func(fractal: account, account: account);
+
+    @since(feature = petitions)
+    create-petition: func(fractal: account, subject: string, body: string);
```
cool-ant, line 67: we'll want this returning an `id` (stg-tx-id) so callers can then reference the created petition, right?

James-Mart: I don't think so, this action isn't what creates the stated tx. It will either need to accept a staged tx ID as a param, or it can just be included in the staged tx itself

PR #1124, opened 2025-04-02
Remove Events::setSchema in favor of Packages::setSchema

swatanabe:
- C++ package support installs schema
- C++ from_json can deserialize Schema


James-Mart (commented):
One missing
https://github.com/gofractally/psibase/blob/bfe3c75da55c0324e5d9621d0490229419a4d929/package-templates/basic-01/%7B%7Bproject-name%20%7C%20upper_camel_case%7D%7D/service/src/lib.rs#L40

James-Mart (approved):

PR #1138, opened 2025-04-09
Add Evaluations to fractal wit spec

cool-ant:


services/user/Fractal/doc/fractal-service.wit
```diff
@@ -20,10 +23,10 @@ interface types {
...
     record member {
         account: account, // primary key
-        @since(feature = invites)
+        @since(feature = fractal-invites)
         inviter: option<account>,
         // display name, bio, avatar - stored in `sites` service
-        joined-at: timestamp,
+        joined-at: timepointsec,
         titles: u32,
     }
```
cool-ant, line 31: If this contract handles multiple Fractals and therefore Members *of* Fractals, how is the association between Member and Fractal made? Is there James-visible magic I don't see? I think we're missing a `member.list<fractal>` field.

cool-ant: Ah, alternatively (and I'm guessing your intention), we could add `fractal: account` to `record member`. That way `member` indicates a member *in* a fractal, such that there could be numerous `record member` entries for the same chain account, one for each `fractal` they participate with, and
We can get the member list of a fractal by querying the `member` records where `fractal` == <fractal name>.

services/user/Fractal/doc/fractal-service.wit
```diff
@@ -1,11 +1,14 @@
...
     @since(feature = scheduled-meetings)
     record meeting-config {
         // @velua, you spent plenty of time on this for the Psinq UI. Would you fill in the minimal set of details we determined were necessary for meeting schedule?
     }
 
     record fractal {
-        service: account, // Primary key
+        account: account, // Primary key
         created-at: timepointsec,
         name: string,
```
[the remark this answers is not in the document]
James-Mart: I think not unique

services/user/Fractal/doc/fractal-service.wit
James-Mart: Yep, this
> we could add fractal: account to record member

James-Mart (approved):

PR #1140, opened 2025-04-11
update key import/export

James-Mart:
This base64url encodes the full private key when exporting it, so that the encoded string can be more easily decoded and used to to import keys into the cli. The cli is currently the only way to upgrade packages, so we need to be able to import account keys into it.


swatanabe:
If the goal is to be able to use the key in the CLI, wouldn't it make more sense to have a button that saves PEM to a file?

James-Mart:
> If the goal is to be able to use the key in the CLI, wouldn't it make more sense to have a button that saves PEM to a file?

Absolutely. I hope to recruit a front-end person to dramatically improve this UX.

Velua (approved):

PR #1141, opened 2025-04-11
Use JSON instead of hex packed arguments in postinstall scripts

swatanabe:


libraries/psio/include/psio/schema.hpp
```diff
@@ -1085,6 +1150,489 @@ namespace psio
...
+      };
+
+      template <typename To, typename From>
+      To checked_cast(From from)
+      {
+         if constexpr (std::is_signed_v<From>)
+         {
+            if constexpr (std::is_signed_v<To>)
+            {
+               if (from < std::numeric_limits<From>::min() || from > std::numeric_limits<To>::max())
```
James-Mart, line 1357: Should be `std::numeric_limits<To>::min()`

James-Mart (approved):

PR #1143, opened 2025-04-15
Allow tables to be defined in the service module

swatanabe:


rust/test_contract_2/src/lib.rs
```diff
@@ -19,16 +22,6 @@ mod tables {
...
-}
-
-#[psibase::service(tables = "tables")]
-#[allow(non_snake_case)]
-mod service {
-    use crate::tables::{Answer, AnswerTable};
-    use async_graphql::{connection::Connection, *};
-    use psibase::*;
-    use serde::{Deserialize, Serialize};
-    use serde_aux::field_attributes::deserialize_number_from_string;
```
cool-ant, line 31: Instead of changing this test to test the new functionality, we should add a new test so we test both formats.

cool-ant (changes requested):
just the one comment on the test. Other than that, this is really nicely done.

rust/test_contract_2/src/lib.rs
swatanabe: There are lots of places using the old format:
```
rust/test_psibase_macros/services/add-check-init/src/lib.rs:#[psibase::service(name = "addcheckinit", tables = "service_tables")]
rust/test_subjective/src/lib.rs:#[psibase::service(name = "subjective", tables = "tables")]
```

cool-ant (approved):
:shipit:

PR #1148, opened 2025-04-18
improve default vscode tasks/scripts

James-Mart:
For organization, I use bash scripts instead of scripting inline in the `tasks.json` file. This also allows the script to be called from anywhere (useful for multi-root workspaces)

### DIFF
--- a/.vscode/scripts/continue.sh
+++ b/.vscode/scripts/continue.sh
@@ -16,4 +16,4 @@ if [ -z "$HOST_IP" ] || [ -z "$PRODUCER" ] || [ -z "$PORT" ]; then
     exit 1
 fi
 
-psinode "$PROJECT_ROOT/db" -p "$PRODUCER" --admin-authz=rw:ip:"$HOST_IP" -l "$PORT" 
+psinode "$PROJECT_ROOT/db" -p "$PRODUCER" --admin-authz=rw:ip:"$HOST_IP" --admin-authz=rw:loopback -l "$PORT" 
--- a/.vscode/scripts/continue.sh
+++ b/.vscode/scripts/continue.sh
@@ -1,0 +1,19 @@
+#!/bin/bash
+
+# Continue/resume a chain
+# Same as launch but doesn't first delete the db.
+
+SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )"
+PROJECT_ROOT="$(dirname "$(dirname "$SCRIPT_DIR")")"
+
+HOST_IP=$1
+PRODUCER=$2
+PORT=$3
+
+if [ -z "$HOST_IP" ] || [ -z "$PRODUCER" ] || [ -z "$PORT" ]; then
+    echo "Error: Missing required parameters"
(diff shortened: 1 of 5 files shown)


.vscode/tasks.json.sample
```diff
@@ -1,42 +1,73 @@
...
-            "id": "testsuite",
-            "type": "promptString",
-            "default": "[mytestsuite]",
-            "description": "Enter test suite key (Example: \"[MyTestSuite]\""
+      }
+    },
+    {
+      "label": "Continue",
+      "type": "shell",
+      "command": ".vscode/scripts/continue.sh <YOUR_HOST_IP> myprod 8080",
```
swatanabe, line 21: Is `<YOUR_HOST_IP>` something that needs to be replaced manually? Won't that cause a problem with psibase-contributor just copying the file?

.vscode/scripts/continue.sh
```diff
@@ -0,0 +1,19 @@
...
+PRODUCER=$2
+PORT=$3
+
+if [ -z "$HOST_IP" ] || [ -z "$PRODUCER" ] || [ -z "$PORT" ]; then
+    echo "Error: Missing required parameters"
+    echo "Usage: $0 <host_ip> <producer> <port>"
+    exit 1
+fi
+
+psinode "$PROJECT_ROOT/db" -p "$PRODUCER" --admin-authz=rw:ip:"$HOST_IP" -l "$PORT"
```
swatanabe, line 19: Should this also have `loopback`, like the launch scripts?

.vscode/tasks.json.sample
James-Mart: It must be replaced manually, yes. I'll have to add a note about this in the psibase contributor docs. Sounds crazy, but afaict there is no way to reliably be able to know what that IP will be in a generic way (across host OS).

.vscode/scripts/continue.sh
James-Mart: Yep, good catch - fixed

.vscode/tasks.json.sample
swatanabe: Didn't we find that it was 127.0.0.1 on some systems? Would it work to use add the ip address iff `$HOST_IP` is defined?

James-Mart: I don't think it's ever 127.0.0.1.

On my windows/wsl2 setup it's currently `172.23.0.1` which can be queried from within the container using something like [this](https://github.com/gofractally/image-builders/blob/main/docker/psibase-contributor.Dockerfile#L82).

But on a Mac, I think http requests from the host are double-NATed, and nothing I try from within the container seems to be able to resolve to the IP address that we see in the psinode logs (Which is 192.168.65.1 or I think sometimes 192.168.65.2)

swatanabe: Even so, I think it would be better to rely on an env variable instead of manually editing the config file. Is the address stable across stopping/starting the container or rebooting the host?

James-Mart: Okay, I figured out a way to be able to continue to use $HOST_IP here. The drawback is that I need mac-os specific logic in `psibase-contributor`.

PR #1151, opened 2025-04-21
Add psibase upgrade subcommand

swatanabe:


doc/psidk/src/run-infrastructure/cli/psibase.md
```diff
@@ -222,6 +223,20 @@ Search for packages
...
 
+### upgrade
+
+`psibase` [`-a` *url*] `upgrade` [`--latest`] [*packages*\.\.\.]  
+
+Upgrade installed packages. This will not upgrade packages to a new major version unless `--latest` is used.
+
+- `--latest`
+
+  Don't limit upgrades to versions that are compatible with the current version.
```
James-Mart, line 234: This seems to me like slightly odd nomenclature. Is this a conventional use of the word, "latest"?

James-Mart: I think I'm used to seeing `--latest` as a selector for a specific version, sometimes versions are tagged with "latest". 
In this case it's a flag that means something like "remove the default limitation on what versions are installed".

Ultimately it's fine, I get why you named it that. Just struck me as slightly odd when I first read it.

James-Mart (approved):

PR #1153, opened 2025-04-22
Take wasip2 shims from jco

James-Mart:
This introduces some minor refactoring to the supervisor code. The most important change is that we now take the wasi shims directly from the jco dependency, rather than having a separately maintained shim file.


cool-ant (approved):

James-Mart:
Next attempt: unplug from unified yarn system

James-Mart:
Closing this in favor of https://github.com/gofractally/psibase/pull/1305
I moved most of the refactoring to that PR, since it has other supervisor-related changes.

But I left out the change that would take JCO shims from the local `jco` bundles, because I couldn't figure out how to do it anymore now that we're using yarn pnp. Perhaps I'll try again later, but for now the shims are still manually minified, copied into our sourcetree, and loaded using vite's `?raw` syntax.

PR #1164, opened 2025-05-02
fix erroneous dev-dependency in plugin

James-Mart:
Prevents tests written in plugin files from compiling


cool-ant:
> Prevents tests written in plugin files from compiling

and isn't needed? service refers back a dir the same way. Is this not needed only for plugin?

James-Mart:
I don't think it's needed for the plugin. 

`test_package` only has this `dev-dependency` for the service.

swatanabe:
> > Prevents tests written in plugin files from compiling
> 
> and isn't needed? service refers back a dir the same way. Is this not needed only for plugin?

The service needs a dev-dependency on the package so that `psibase::test_case` can install the package.

cool-ant (approved):

PR #1171, opened 2025-05-07
split module

Velua:


James-Mart (approved):
module names are conventionally lower_snake_case I believe

PR #1184, opened 2025-05-13
Fix mac boot tx expiry

James-Mart:
Increases tx expiry timeout from 2 to 4 minutes

### DIFF
--- a/rust/psibase/src/main.rs
+++ b/rust/psibase/src/main.rs
@@ -958,8 +958,7 @@ async fn get_package_registry(
 
 async fn boot(args: &BootArgs) -> Result<(), anyhow::Error> {
     let (client, _proxy) = build_client(&args.node_args.proxy).await?;
-    let now_plus_120secs = Utc::now() + Duration::seconds(120);
-    let expiration = TimePointSec::from(now_plus_120secs);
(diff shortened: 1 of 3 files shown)


James-Mart:
Reverted the compression change, because I realized the expiry timer is set before compression anyway, so increasing time spent compressing counts against the timer, which counteracts any tx speed up

rust/psibase/src/main.rs
```diff
@@ -958,8 +958,7 @@ async fn get_package_registry(
 
 async fn boot(args: &BootArgs) -> Result<(), anyhow::Error> {
     let (client, _proxy) = build_client(&args.node_args.proxy).await?;
-    let now_plus_120secs = Utc::now() + Duration::seconds(120);
-    let expiration = TimePointSec::from(now_plus_120secs);
+    let expiration = TimePointSec::from(Utc::now() + Duration::seconds(240));
```
swatanabe, line 961: tapos can only go up to 127 seconds in the future before it has to switch to the other form. We're not actually checking tapos during boot, though. The expiration serves no useful purpose during boot. Maybe we should just set it to zero and change transact to not check it.

James-Mart:
Closing this, we need a better solution. Captured in 
https://github.com/gofractally/psibase/issues/1185

PR #1187, opened 2025-05-15
Minimal workspace config with yarn 1

cool-ant:
This is a bare-bones PR that uses yarn v1 (no change) with workspaces set up to save node_modules space and the smallest number of vite.config.ts changes to get the UIs to build.
You should use the default yarn v1.22.22, and you should lock cargo-component to 0.20.0: `cargo install cargo-component@0.20.0 --locked`


services/system/Accounts/ui/package.json
```diff
@@ -51,6 +50,6 @@
     "tailwindcss": "^3.4.10",
     "typescript": "^5.5.3",
     "typescript-eslint": "^8.0.1",
-    "vite": "^5.4.1"
+    "vite": "4.5.2"
```
James-Mart, line 53: You have 
```
 "devDependencies": {
    "vite": "5.4.1"
  }
```
In `services/package.json`, but then here you install `4.5.2`. Intentional?

James-Mart (approved):

PR #1180, opened 2025-05-12
Fractals

Velua:
Adds `fractals` service, which allows for the creation of fractals, membership management, and evaluation scheduling.

PR #1188, opened 2025-05-15
Remove old fractal service

Velua:
Deletes the old version of the Fractal service, new one is coming in https://github.com/gofractally/psibase/pull/1180


James-Mart:
The `doc/` folder should probably be copied somewhere for reference

James-Mart (approved):

PR #1198, opened 2025-05-23
Add group finish results

Velua:
There's a better way of doing this but this will work for now.


James-Mart (approved):

James-Mart:
> There's a better way of doing this but this will work for now.

What better way?

PR #1208, opened 2025-05-29
Rust SDK - Allow table keys of objects that don't implement the Copy trait

cool-ant:
Primary key function returns a reference instead of copying


services/system/StagedTx/service/src/lib.rs
```diff
@@ -46,7 +46,7 @@ pub mod service {
     fn check_init() {
         let table: InitTable = InitTable::new();
         check(
-            table.get_index_pk().get(&()).is_some(),
+            table.get_index_pk().get(&&()).is_some(),
```
swatanabe, line 49: Is this (and similar changes) needed?

services/user/Chainmail/service/src/expanded_table.rs
```diff
@@ -321,8 +321,8 @@ mod service {
         type PrimaryKey = ();
         const SECONDARY_KEYS: u8 = 0u8;
         const DB: psibase::DbId = psibase::DbId::Service;
-        fn get_primary_key(&self) -> Self::PrimaryKey {
-            self.pk()
+        fn get_primary_key(&self) -> RawKey {
+            self.pk().to_key()
```
swatanabe, line 325: - This is incorrect. It's from a non-working version of the macro.
- Why is this file checked in?

services/system/StagedTx/service/src/lib.rs
cool-ant: nope. residual from earlier implementations. good catch.
reverted

swatanabe (approved):

James-Mart (approved):

James-Mart:
@cool-ant - You mentioned you didn't do anything with secondary keys. Just checking if I understand - does that mean it currently will fail to compile if I try to use e.g. a `Checksum256` in a secondary key?

swatanabe:
> @cool-ant - You mentioned you didn't do anything with secondary keys. Just checking if I understand - does that mean it currently will fail to compile if I try to use e.g. a `Checksum256` in a secondary key?

This was only a problem when the primary key was a field. secondary keys can only be defined as methods.

PR #1218, opened 2025-06-04
Rename RawNativeFunctions.hpp

swatanabe:


James-Mart:
`api` is pretty generic. What about "host-api" or "native-api"? ("Host" is the language used by the rest of the wasm world)

James-Mart:
Oops, closed by accident

PR #1226, opened 2025-06-06
Refactor accounts UI

cool-ant:
This JSX had gotten to hard too reason about. I broke out numerous components and moved hooks out to separate files in the hooks dir.
I tried to touch *no* functionality. Got pretty close to that.
Let me know if you see anything not working or acting differently than before. I haven't found any deviations.
AccountAvailability check is the place I changed the conditionals at all, but I checked it pretty rigorously in an attempt to ensure it wasn't a problem, but lmk.


services/system/Accounts/ui/src/TokenErrorUIs.tsx
```diff
@@ -0,0 +1,89 @@
...
+          Please ask the sender{" "}
+          <span className="text-primary">{inviteToken?.inviter}</span> for a new
+          one.
+        </CardDescription>
+      </CardHeader>
+    </Card>
+  );
+};
+
+export const InviteAlreadyAcceptedCard = (token: string) => {
```
Velua, line 42:
```suggestion
export const InviteAlreadyAcceptedCard = ({ token }: { token: string }) => {
```

A few of these components aren't accepting props properly, should look something more like this.

Velua (changes requested):

services/system/Accounts/ui/src/AccountSelection.tsx
```diff
@@ -197,13 +137,11 @@ export const AccountSelection = () => {
     return () => subscription.unsubscribe();
   }, [form]);
 
-  const [activeSearch, setActiveSearch] = useState("");
+  const [accountsToRender, setAccountsToRender] = useState<any[]>([]);
```
Velua, line 140: Any reason we can't type this?

services/system/Accounts/ui/src/AccountSelection.tsx
```diff
@@ -197,13 +141,14 @@ export const AccountSelection = () => {
     return () => subscription.unsubscribe();
   }, [form]);
 
-  const [activeSearch, setActiveSearch] = useState("");
+  const [accountsToRender, setAccountsToRender]: [
+    AccountType[],
+    (accounts: AccountType[]) => void,
+  ] = useState<AccountType[]>([]);
```
brandonfancher, line 147: Usually this would just be:

`const [accountsToRender, setAccountsToRender] = useState<AccountType[]>([]);`

The rest is inferred and typed properly.

services/system/Accounts/ui/src/AccountSelection.tsx
```diff
@@ -524,48 +306,21 @@ export const AccountSelection = () => {
...
-                          onAccountSelection(account.id);
-                        }}
-                        isSelected={selectedAccountId == account.id}
-                        key={account.id}
-                        name={account.account}
-                      />
-                    ))}
-              </div>
+              <AccountsList
+                isAcocuntsLoading={isAccountsLoading}
```
brandonfancher, line 319: `isAcocuntsLoading` -> `isAccountsLoading`

services/system/Accounts/ui/src/AccountSelection.tsx
brandonfancher: Another comment about this: You're duplicating accounts state here. Fetching data (like accounts) and stuffing it into state and then managing that state as a new list of accounts is an antipattern. Better to do what was done in the previous version: fetch the accounts, track the state of the active search, and filter the list of accounts-to-render from the full list of accounts on render.

services/system/Accounts/ui/src/AccountSelection.tsx
```diff
@@ -340,82 +212,44 @@ export const AccountSelection = () => {
...
+      if (!connectionToken) {
+        throw new Error(`Expected connection token for a login`);
+      }
+      login({
+        app: connectionToken.app,
+        origin: connectionToken.origin,
+        accountName: selectedAccount!.account,
+      });
+    }
+  };
```
brandonfancher, line 251: What happens to the UI if this function throws?

services/system/Accounts/ui/src/TokenErrorUIs.tsx
```diff
@@ -0,0 +1,89 @@
...
+  CardHeader,
+  CardTitle,
+} from "./components/ui/card";
+import { AlarmClockMinus, TriangleAlert } from "lucide-react";
+import dayjs from "dayjs";
+
+import { useDecodeInviteToken } from "./hooks/useDecodeInviteToken";
+import { useDecodeToken } from "./hooks/useDecodeToken";
+
+export const InviteExpiredCard = (token: string) => {
```
brandonfancher, line 13: Props should be a props object here too.

services/system/Accounts/ui/src/TokenErrorUIs.tsx
```diff
@@ -0,0 +1,89 @@
...
+            {inviteToken.actor}
+          </span>
+          .
+        </CardDescription>
+      </CardHeader>
+    </Card>
+  );
+};
+
+export const InviteRejectedCard = (token: string) => {
```
brandonfancher, line 67: And here.

services/system/Accounts/ui/src/TokenErrorUIs.tsx
```diff
@@ -0,0 +1,89 @@
...
+  return (
+    <Card className="w-[350px] mx-auto mt-4">
+      <CardHeader>
+        <div className="mx-auto">
+          <TriangleAlert className="w-12 h-12" />
+        </div>
+        <CardTitle>Invitation rejected.</CardTitle>
+        <CardDescription>
+          This invitation has been rejected
+          {inviteToken?.actor && inviteToken.actor !== "invite-sys"
```
brandonfancher, line 82: `{inviteToken?.actor !== "invite-sys" ? ...` is sufficient here.

services/system/Accounts/ui/src/AccountSelection.tsx
cool-ant: Whatever happened before this refactor. The code is functionally the same; just rearranged.
But good question.

Velua (approved):

brandonfancher (approved):

PR #1230, opened 2025-06-11
Unified Configs > Step 1 > Get Single App Updated

brandonfancher:
This gets configs in the Branding app (like eslint, prettier, tailwind, tsconfigs, shadcn) all updated and in the shape we want them in for all apps. And it updates them to modern versions.

First, we focus on one instance (Branding app) before we generalize. First, the concrete. Then, the abstract.

## PRs to follow
- Hoist one or more configs into the monorepo and get that working for 2 apps
- Hoist remaining configs
- Apply to all apps


services/user/Branding/ui/src/App.tsx
```diff
@@ -1,11 +1,12 @@
...
-import { Button } from "@shadcn/button";
-import { Label } from "@shadcn/label";
-import { Input } from "@shadcn/input";
+import { getSupervisor, siblingUrl } from "@psibase/common-lib";
 
-import { siblingUrl, getSupervisor } from "@psibase/common-lib";
-import { Nav } from "@components/nav";
+import { Button } from "@/components/ui/button";
+import { Input } from "@/components/ui/input";
+import { Label } from "@/components/ui/label";
```
cool-ant, line 7: Perhaps, given you're making this a model, we add a @/components/ui/index.js so we can import all 3 in one line?

cool-ant (approved):

services/user/Branding/ui/src/App.tsx
brandonfancher: I'm seeing articles discussing barrel files (those index.ts files that combine everything) and they're raising questions about whether they're an antipattern or not, or in what cases, etc. I want to learn more about that before we combine these and export them. But something to look into for sure.

Velua (commented):
Code looks good, will test shortly. 
Shared hooks would be neat one day!

Velua (approved):

PR #1235, opened 2025-06-14
Tailwind 4 vite plugin should only apply to upgraded UIs

brandonfancher:
Not sure why my prior few PRs' builds passed, but applying the tailwindcss 4 vite plugin to the top-level shared vite config was premature. All UIs need to be upgraded from tw3 to tw4 before we do that. In the meantime, we'll temporarily apply that plugin to the individual upgraded UI vite configs.


brandonfancher:
@Velua Once this is merged, we can merge `main` into the `eval-fractal` branch and that should clear up build issues on that PR.

PR #1236, opened 2025-06-16
Initial packages plugin

swatanabe:
This contains the core functionality for installing and updating packages. The UI is responsible for fetching packages as they can come from an off-chain source.

The UI is mainly so that I could test that the plugin works. Don't look at it too hard.

Not included here:
- Viewing and approving staged transactions


services/user/Sites/plugin/wit/world.wit
```diff
@@ -26,6 +26,8 @@ interface api {
     /// file (after compression if applied).
     upload: func(file: file, compression-quality: u8) -> result<_, error>;
 
+    upload-encoded: func(file: file, content-encoding: string) -> result<_, error>;
```
James-Mart, line 29: Doc comment

services/user/Packages/plugin/wit/world.wit
```diff
@@ -0,0 +1,57 @@
...
+        old: option<meta>,
+        new: option<list<u8>>,
+    }
+}
+
+interface queries {
+    use host:common/types.{error};
+    use types.{meta, package-source};
+    get-installed-packages: func() -> result<list<meta>, error>;
+    get-sources: func() -> result<list<package-source>, error>;
```
James-Mart, line 52: Would be good to add doc comments on these. Less important for `impl.wit` since that only contains private interfaces.

James-Mart (commented):
Doc comments in `wit` files will eventually be rendered on a plugin explorer UI ([for example](https://wa.dev/wasi:io)).

services/user/Packages/plugin/src/lib.rs
```diff
@@ -0,0 +1,510 @@
...
+
+impl Queries for PackagesPlugin {
+    fn get_installed_packages() -> Result<Vec<types::Meta>, CommonTypes::Error> {
+        Ok(get_installed_packages()?
+            .into_iter()
+            .map(|p| p.meta().into())
+            .collect())
+    }
+    fn get_sources() -> Result<Vec<types::PackageSource>, CommonTypes::Error> {
+        let owner = account!("root");
```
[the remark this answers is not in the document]
swatanabe: > Shouldn't this be get_logged_in_user instead of hardcoding it to root?

No. The logged in user is the proposer.

> If hardcoded to root, doesn't it imply that users of this plugin will only see packages available for installation from the repository source configured by the account, root?

Yes, and that's intended. I don't know how the UI will distinguish between installing my own packages to my own accounts and proposing a staged transaction to install as root. For now, the latter is more important.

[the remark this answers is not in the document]
swatanabe: I changed it to pass the account in where its needed. It's now the UI's problem.

services/user/Packages/plugin/wit/world.wit
James-Mart: You added comments on the types but not on these query functions...

James-Mart (approved):

PR #1237, opened 2025-06-16
Upload package data faster

swatanabe:
- The order in which we upload files to the staging area doesn't matter, so there's no need to wait for one to finish before starting the next. However, we do need to wait for them all to finish before actually installing.
- This exposes the need for http-server to use full-malloc, since onBlock may need to send many responses.

### DIFF
--- a/rust/psibase/src/main.rs
+++ b/rust/psibase/src/main.rs
@@ -2,7 +2,7 @@ use anyhow::{anyhow, Context};
 use chrono::{Duration, Utc};
 use clap::{Args, FromArgMatches, Parser, Subcommand};
 use fracpack::Pack;
-use futures::future::join_all;
+use futures::future::{join_all, try_join_all};
 use hmac::{Hmac, Mac};
 use hyper::service::Service as _;
 use indicatif::{ProgressBar, ProgressStyle};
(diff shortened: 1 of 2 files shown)


cool-ant (approved):
LGTM

James-Mart:
> This exposes the need for http-server to use full-malloc, since onBlock may need to send many responses.

@swatanabe  can you explain your reasoning here? Our docs say: "service-simple-malloc will not free memory until the transaction finishes." I guess that implies that full-malloc frees memory within a transaction. Presumably, what, after the service exits (e.g. an action dispatcher pops off the stack), then the memory is freed? How does that help with memory usage in the case of onBlock calling many responses?

swatanabe:
> > This exposes the need for http-server to use full-malloc, since onBlock may need to send many responses.
> 
> @swatanabe can you explain your reasoning here? Our docs say: "service-simple-malloc will not free memory until the transaction finishes." I guess that implies that full-malloc frees memory within a transaction. Presumably, what, after the service exits (e.g. an action dispatcher pops off the stack), then the memory is freed? How does that help with memory usage in the case of onBlock calling many responses?

Neither simple-malloc nor full-malloc ever returns memory to native. The difference is that full-malloc can re-use memory that has been freed. Under simple-malloc, `free` is a no-op.

PR #1240, opened 2025-06-23
Add wasm service for XAdmin

swatanabe:


services/system/Transact/src/RTransact.cpp
```diff
@@ -930,6 +943,12 @@ std::optional<HttpReply> RTransact::serveSys(const psibase::HttpRequest& request
          return reply;
       }
    }
+   else if (target == "/jwt_key" && request.method == "GET")
+   {
+      check(user.has_value(), "Unauthorized");
+      check(to<XAdmin>().isAdmin(*user), "Forbidden");
```
James-Mart, line 949: Since `XAdmin` service is a subjective service, can this inline action only be called in an RPC or otherwise subjective context?

services/user/XAdmin/query/src/XAdmin.cpp
```diff
@@ -0,0 +1,364 @@
...
+               else
+               {
+                  return {};
+               }
+            }
+         }
+         else if (req.method == "DELETE")
+         {
+            initRefCounts();
+            // MUST delete all data to avoid exposing node secrets to on-chain services
```
James-Mart, line 178: Should we create an issue to track that this wasn't yet implemented?

services/system/Transact/src/RTransact.cpp
swatanabe: Right. The chain theoretically could make it work by providing an `x-admin` service. Otherwise, `x-admin` won't be found in an objective context.

James-Mart (approved):

PR #1252, opened 2025-06-24
Allow setting the number of threads used for the run table

swatanabe:


James-Mart:
Should update `psinode.md` with the new option as well. 

Also, please explain in a little more detail what this parameter does. I think the current description makes one wonder whether transactions are being processed in parallel, or whether "actions posted by a service" are a specific/different kind of parallel action.

James-Mart (approved):

PR #1264, opened 2025-06-30
Resize database on startup to match database-cache-size

swatanabe:


libraries/triedent/include/triedent/ring_allocator.hpp
```diff
@@ -270,6 +273,48 @@ namespace triedent
       return make_update_free(freed);
    }
 
+   void ring_allocator::unsafe_resize(std::uint64_t new_size, auto&& move_object)
```
James-Mart, line 276: Would be nice to print a console message so that a node operator has an easy way to confirm that resizing happened

James-Mart: Running this locally and I can change the param and restart the node, but I'm not sure how to check if it's working

swatanabe: `ls -l data/`

James-Mart: What behavior do you expect if I boot fresh? The size I see if I add the sizes of everything in `/data/` is 448 MB on fresh boot, even though config file says 256 MiB. Stopping the node and continuing without changing it, keeps it at the same size.

Stopping, and continuing with `--database-cache-size "512 MB"`, now the sum total size is `796 MB`

James-Mart: Okay that's much closer. Still not bang on the configured size (488 MB instead of 500)

James-Mart (approved):

PR #1266, opened 2025-07-01
Add flags to rust tokens service

James-Mart:
I wrote some helpers to simplify implementation of bitset integers in tables in rust services.

* `Flags` helper class that simplifies getting/setting individual bits from an integer being used as a bitset.
* `define_flags` macro that creates a helper struct that defines:
    * Named constants for each bit in the bitset - for linking from service, its plugin, and third-party services.
    * A struct that can be returned from a custom gql query resolver to allow for querying individual flags by name

I implemented in the rust tokens service, since it was the primary impetus for this in the first place.


rust/psibase/src/flags.rs
```diff
@@ -0,0 +1,143 @@
...
+        Self::validate_index(index);
+        (self.value & (T::from(1) << index)) != T::from(0)
+    }
+
+    pub fn value(&self) -> T {
+        self.value
+    }
+}
+
+/// Macro to generate flag types with constants, get() method, and GraphQL schema
```
swatanabe, line 63: There are four different types being used: u8 (index), u8 (flags), Flags&lt;u8&gt;, and TokenFlags. I think there should only be two, one for the index and one for the bitset.

```
pub struct TokenFlags(u8);
impl TokenFlags {
    pub const UNBURNABLE: TokenFlags = TokenFlags(0);
    pub const UNTRANSFERABLE: TokenFlags = TokenFlags(1);
    pub const UNRECALLABLE: TokenFlags = TokenFlags(2);
}
impl FlagsType for TokenFlags {
    type BitsetType = u8;
    type JsonType = TokenFlagsJson;
}
struct TokenFlagsJson {
    unburnable: bool,
    untransferable: bool,
    unrecallable: bool,
}
impl From<Flags<TokenFlags>> for TokenFlagsJson;
impl From<TokenFlagsJson> for Flags<TokenFlags>;
```

rust/psibase/src/flags.rs
```diff
@@ -0,0 +1,143 @@
...
+/// define_flags!(StructName, u8, {
+///     first_flag,
+///     second_flag,
+///     third_flag,
+/// });
+/// ```
+///
+/// This generates:
+///
+/// - A struct with the given name and bool fields (snake_case names)
```
swatanabe, line 79: We've been using camelCase for JSON. I believe this was originally because that's the graphQL convention. async_graphql handles the conversion automatically, but serde need `rename`.

rust/psibase/src/flags.rs
```diff
@@ -0,0 +1,177 @@
...
+///
+/// `Flags` represents a bitset structure that works with flag types implementing `FlagsType`.
+///
+/// Provides a convenient way to manage an integer that represents a set of boolean flags
+/// using strongly-typed indices.
+///
+/// # Type Parameter
+///
+/// * `F` - The flag type that implements `FlagsType`
+pub struct Flags<F: FlagsType> {
```
swatanabe, line 27: I think Flags should be Packable and stored directly rather being cast to and from u8. It would also be nice if the representation is compatible with C++ `Bitset`.

James-Mart:
@swatanabe this will not pass CI/CD because it's targeting merge into `oxidise-tokens` which is still WIP

swatanabe (approved):

PR #1271, opened 2025-07-02
Remove awaiting for state change

Velua:
Now that `supervisor.functionCall` calls to trigger transactions are awaited before they're resolved, we no longer need to wait lengths of time before side effects are reflected.


James-Mart (approved):

brandonfancher (approved):
🙌

PR #1280, opened 2025-07-03
Fix CORS

swatanabe:
A condition was backwards and not propagated.

### DIFF
--- a/libraries/psibase_http/handle_request.cpp
+++ b/libraries/psibase_http/handle_request.cpp
@@ -730,7 +730,7 @@ namespace psibase::http
 
             if (req.method() == bhttp::verb::options)
             {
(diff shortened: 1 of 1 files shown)


James-Mart:
Looks like this fixed loading various plugins, but I am still seeing CORS errors on `/graphql`  requests:

```
Access to XMLHttpRequest at 'http://accounts.psibase.localhost:8080/graphql' from origin 'http://supervisor.psibase.localhost:8080' has been blocked by CORS policy: Response to preflight request doesn't pass access control check: No 'Access-Control-Allow-Origin' header is present on the requested resource.
```

swatanabe:
Right, that's an `OPTIONS` request that isn't handled by wasm...

James-Mart (approved):
Tested working

PR #1283, opened 2025-07-04
Bump version

James-Mart:


CMakeLists.txt
```diff
@@ -319,12 +319,10 @@ include(libraries/psibase/sdk/pack_service.cmake)
 
 set(SERVICE_DIR ${CMAKE_CURRENT_BINARY_DIR}/share/psibase/packages)
 
-set(PSIBASE_VERSION "0.17.0")
```
swatanabe, line 322: I think we can still keep a named constant, and reduce the amount of churn using something like this:
```
set(PSIBASE_VERSION "0.17.0")
set(PSIBASE_VERSION_BASE "0.17")
```

Then the version for a patch update will be `"${PSIBASE_VERSION_BASE}.1"`. On a release, we can replace all occurrences of `PSIBASE_VERSION_BASE`, which is more reliable than searching for a magic string.

James-Mart: Done

PR #1286, opened 2025-07-06
Add config app

Velua:
Config UI replaces role of `Branding` and `Producers`, subsequent updates will add further network administration features.


services/user/Config/plugin/src/lib.rs
```diff
@@ -0,0 +1,65 @@
...
+use crate::bindings::exports::config::plugin::producers::Producer;
+use staged_tx::plugin::proposer::set_propose_latch;
+
+use psibase::services::producers::SERVICE as PRODUCERS_SERVICE;
+
+struct ProposeLatch;
+
+impl ProposeLatch {
+    fn new(app: psibase::AccountNumber) -> Self {
+        set_propose_latch(Some(&app.to_string())).unwrap();
```
James-Mart, line 19: Technically I think this raii isn't really needed anymore since the propose latch can now only be set by the top-level app.

services/user/Config/plugin/wit/world.wit
James-Mart, line 1: * Doc comments

My OCD would also love if you made the spacing less odd

services/user/Config/ui/src/components/settings-dropdown.tsx
James-Mart, line 1: Shouldn't there be some kind of "hoisted" UI components you can use instead of implementing them again? Or otherwise what is the hoisting for?
 
CC @brandonfancher

James-Mart:
Also please add a brief pr description (links to this PR will show up in our changelog published with the new release)

services/user/Config/plugin/Cargo.toml
```diff
@@ -0,0 +1,33 @@
...
+[lib]
+crate-type = ["cdylib"]
+
+[package.metadata.component]
+package = "config:plugin"
+
+[package.metadata.component.target]
+world = "impl"
+
+[package.metadata.component.target.dependencies]
```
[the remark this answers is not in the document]
Velua: Removed

James-Mart (approved):

services/user/Config/ui/src/components/settings-dropdown.tsx
brandonfancher: We have all our configs and shared shadcn UI components hoisted. That was big savings. But we don't yet have any of our custom UI components hoisted. That said, it's now just a matter of putting this component in our shared location and referencing it from there. It's all set up for that now.

Issue #1274, opened 2025-07-02
Move setting CORS header to wasm


James-Mart:
Currently, native sets the CORS allowance header on every request. This should move to wasm so we have better control of the logic, as there are some resources for which we do not want to allow cross-origin requests (e.g. `[set-cookie](https://github.com/gofractally/psibase/issues/1273)`)

swatanabe:
> I'm thinking that, before the response is returned to the client, wasm can check if the request added any CORS headers, and if not, it adds the default (allow CORS). This allows specific requests to override the default permissive behavior.

The problem is that the way to disable CORS is not to send any CORS headers. There isn't a value that indicates to refuse access.

PR #1291, opened 2025-07-08
Add set-auth-cookie endpoint

cool-ant:
This PR implements the `set-auth-cookie` endpoint, used to set query tokens for authenticated queries.
Building on the call to login() to get the query token (previous PR), this PR adds the server endpoint that can set an HttpOnly cookie with the query token as a value.
The Accounts app still prints the query token to the console.
A localhost/development mode has also been included: the endpoint will alter the name of the cookie if the host is localhost, allowing the cookie to function in a localhost context (__HOST-SESSION cookie has requirements localhost can't fulfill).
Stripped out the old common/set-auth-cookie infra, in prep for Accounts plugin doing the cookie setting via the new post() function, which allows privileged apps to post cross-domain.

Testing this is not for the faint of heart, but the simplest approach would be
1) Replace the getNetworkName() function with this version, which does an authenticated query
```
const getNetworkName = async () => {
        try {
            // Direct GraphQL call (testing authentication)
            const response = await fetch("/graphql", {
                method: "POST",
                headers: { 
                    "Content-Type": "application/json"
                },
                body: JSON.stringify({ 
                    query: "query { networkName }" 
                }),
            });
            
            const result = await response.json();
            console.log("Direct GraphQL response:", result);
            
            if (result.errors) {
                console.error("GraphQL errors:", result.errors);
                setNetworkName("Error: " + result.errors[0].message);
                return;
            }
            
            if (result.data && result.data.networkName) {
                setNetworkName(result.data.networkName);
            } else {
                setNetworkName("No data returned");
            }
        } catch (e: unknown) {
            console.error("Error getting network name:", e);
            setNetworkName("Error: " + (e instanceof Error ? e.message : String(e)));
        }
    };
```

2) and modding the Branding service's get_network_name() function to check/output the user that comes as a param to serveSys()
```
#[action]
    #[allow(non_snake_case)]
    fn serveSys(request: HttpRequest, _socket: Option<i32>, user: Option<AccountNumber>) -> Option<HttpReply> {
        // Debug logging
        println!("🔍 Branding serveSys called");
        println!("  Method: {}", request.method);
        println!("  Target: {}", request.target);
        println!("  User: {:?}", user);
        
        // Check if this is a GraphQL request
        if request.target.starts_with("/graphql") && request.method == "POST" {
            println!("🔒 GraphQL request detected");
            println!("  Content-Type: {}", request.contentType);
            println!("  Body: {}", String::from_utf8_lossy(&request.body));
            
            // For GraphQL requests, check if user is authenticated
            if user.is_none() {
                println!("❌ No user found - returning GraphQL error");
                return Some(HttpReply {
                    status: 200,
                    contentType: "application/json".into(),
                    body: r#"{"data":null,"errors":[{"message":"Authentication required for GraphQL requests","extensions":{"code":"UNAUTHENTICATED"}}]}"#.as_bytes().to_vec().into(),
                    headers: vec![],
                });
            } else {
                println!("✅ User authenticated: {:?}", user);
            }
        }
        
        None.or_else(|| serve_graphql(&request, Query))
    }
```

and lastly,
3) launch your browser so it doesn't block this secure cookie from being set while in running locally without https, e.g., in FireFox, open the URL "about:config", search for "network.strict.transport.security", add it as a bool (since it likely doesn't already exist), and toggle it to false. Chrome also has a launch arg that will accomplish the same.


services/system/HttpServer/src/HttpServer.cpp
```diff
@@ -95,10 +95,26 @@ namespace SystemService
...
       {
          for (const auto& header : result.headers)
          {
+            // Special case: Allow Set-Cookie header only for /common/set-cookie endpoint
+            if (header.name == "Set-Cookie")
+            {
+               if (requestTarget != "/common/set-cookie")
+               {
+                  abortMessage("service " + service.str() +
+                               " attempted to set Set-Cookie header on unauthorized endpoint " +
```
James-Mart, line 111: Why should only the `/common/set-cookie` endpoint be allowed to set cookies?

services/user/CommonApi/src/CommonApi.cpp
```diff
@@ -79,6 +84,23 @@ namespace SystemService
...
+            psio::json_token_stream jstream{request.body.data()};
+            auto                    params = psio::from_json<cookie_data>(jstream);
+
+            // Create response headers with host-only __Host-SESSION cookie
+            std::vector<HttpHeader> headers;
+            std::string             cookieValue = "__Host-SESSION=" + params.accessToken +
+                                      "; Path=/; HttpOnly; Secure; SameSite=Strict";
+            headers.push_back({"Set-Cookie", cookieValue});
+
+            return HttpReply{.status      = HttpStatus::ok,
```
James-Mart, line 99: This is missing the check for the origin to determine if it's a development chain, in which case the cookie should be such that it can be set without configuring the browser in a special way.

services/system/HttpServer/src/HttpServer.cpp
cool-ant: This was the pre-existing situation; we didn't allow WASMs to set security-related anything, including cookies. I'm just maintaining that, according to discussions we had on this topic. this code opens that window only as far as necessary.

libraries/psibase/common/include/psibase/Rpc.hpp
```diff
@@ -70,6 +70,10 @@ namespace psibase
 
       /// Removes a cookie
       void removeCookie(std::string_view name);
+
+      /// Checks if the request origin indicates a development chain
+      /// by looking for "localhost" in the origin header domain
+      bool isDevChainOrigin() const;
```
James-Mart, line 76: I wouldn't make this part of the public API for `HttpRequest`, it should just be part of the unnamed namespace in `CommonApi.cpp`.

libraries/psibase/common/src/Rpc.cpp
```diff
@@ -141,4 +141,64 @@ namespace psibase
...
+   bool HttpRequest::isDevChainOrigin() const
+   {
+      // Check the host field first (always present)
+      if (host.find("localhost") != std::string::npos)
+      {
+         return true;
+      }
+      
+      // Also check origin header if present (for cross-origin requests)
+      for (const auto& header : headers)
```
James-Mart, line 154: What about `HttpRequest::getCookie`?

services/system/HttpServer/src/HttpServer.cpp
James-Mart: I know we discussed that eventually the `/common/set-cookie` endpoint will not have permissive CORS settings (some time after #1274 ), so it can only be called by an app on itself, or by the supervisor.

But I don't think we need to restrict only this endpoint to set cookies.

services/user/CommonApi/src/CommonApi.cpp
James-Mart: The cookie names should probably be consistent, e.g.:
`__HOST-SESSION` and `SESSION`
or 
`__HOST-authed-query` and `authed-query`

services/system/Transact/src/RTransact.cpp
```diff
@@ -1421,6 +1421,7 @@ std::optional<AccountNumber> RTransact::getUser(HttpRequest request)
 {
    std::vector<char>            key = getJWTKey();
    std::optional<AccountNumber> result;
+   bool                         isDevChain = request.isDevChainOrigin();
```
James-Mart, line 1424: Can't RTransact just check for the presence of *auth-cookie* or __HOST-*auth cookie*? Then it wouldn't need to care about whether the origin is localhost or not.

services/system/HttpServer/src/HttpServer.cpp
cool-ant: I'd propose until we see that use case or do some research, we just leave this door closed. In fact, how about we just create a separate issue to investigate this?

libraries/psibase/common/src/Rpc.cpp
```diff
@@ -141,4 +141,64 @@ namespace psibase
          ++iter;
       }
    }
+
+   bool HttpRequest::isDevChainOrigin() const
+   {
+      // Check the host field first (always present)
+      if (host.find("localhost") != std::string::npos)
```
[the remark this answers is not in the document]
cool-ant: The parsing code handles things properly, but I overlooked applying it to both origin and host. That's done now.

libraries/psibase/common/src/Rpc.cpp
```diff
@@ -141,4 +141,64 @@ namespace psibase
...
+      {
+         return true;
+      }
+      
+      // Also check origin header if present (for cross-origin requests)
```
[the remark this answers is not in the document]
cool-ant: You're right. I set it to None and added the Domain attrib, which we'll need to avoid CSRF attacks.

libraries/psibase/common/src/Rpc.cpp
```diff
@@ -82,22 +82,31 @@ namespace psibase
...
+      }
+      return {};
+   }
+
+   std::optional<std::string_view> HttpRequest::getHeader(std::string_view name) const
    {
       for (const auto& header : headers)
       {
-         if (std::ranges::equal(header.name, std::string_view{"cookie"}, {}, ::tolower))
+         if (std::ranges::equal(header.name, name, {}, ::tolower))
```
swatanabe, line 107: `getHeader` should not assume that the name passed in is lower case. You can use `header.matches`.

services/user/CommonApi/src/CommonApi.cpp
```diff
@@ -79,6 +84,30 @@ namespace SystemService
...
+         {
+            request.body.push_back(0);
+            psio::json_token_stream jstream{request.body.data()};
+            auto                    params = psio::from_json<cookie_data>(jstream);
+
+            std::vector<HttpHeader> headers;
+            bool                    isDevChain = psibase::isDevChain(request);
+            std::string             cookieName = "__Host-SESSION";
+            std::string             cookieAttribs =
+                "Path=/; HttpOnly; Secure; SameSite=None; Domain=" + request.host;
```
swatanabe, line 97: `Domain` is not allowed with `__Host`

services/user/CommonApi/src/CommonApi.cpp
```diff
@@ -79,6 +84,30 @@ namespace SystemService
...
+            if (isDevChain)
+            {
+               cookieName    = "SESSION";
+               cookieAttribs = "Path=/; Domain=" + request.host;
```
swatanabe, line 101: `Domain` causes the cookie to be sent for subdomains, which we don't want.

services/system/Transact/src/RTransact.cpp
```diff
@@ -1443,12 +1444,23 @@ std::optional<AccountNumber> RTransact::getUser(HttpRequest request)
...
    }
-   if (auto token = request.getCookie("__Host-SESSION"))
+
+   std::string_view cookieName = "__Host-SESSION";
+   if (isDevChain)
+   {
+      cookieName = "SESSION";
+   }
+
+   if (auto token = request.getCookie(cookieName))
```
swatanabe, line 1454: It's possible to have multiple values for a cookie (`__Host` prevents this, so it was fine before).

libraries/psibase/common/src/Rpc.cpp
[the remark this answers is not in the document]
cool-ant: huh, good catch. I worked a lot with this function and did not get the final version right. I've tested this (updated) code as thoroughly as I can think to do so.

services/system/HttpServer/src/HttpServer.cpp
```diff
@@ -93,12 +94,17 @@ namespace SystemService
       constexpr std::string_view allowedHeaders[] = {"Cache-Control",            //
                                                      "Content-Encoding",         //
                                                      "Content-Security-Policy",  //
-                                                     "ETag"};
+                                                     "ETag",                     //
+                                                     "Set-Cookie"};
```
James-Mart, line 98: afaict you can eliminate all changes in this file except adding this Set-Cookie allowance

services/system/Transact/src/RTransact.cpp
cool-ant: I don't see any evidence of multiple values for cookies (sort of custom encoding of such). RFC 6265 seems clear about that and RFC 6265big doesn't seem to have any reference to the contrary. Could you point me at what you're looking at?

cool-ant: Ok, this PR used the pre-existing getCookie(), and given the scope of this PR, I've created an issue to address this issue by upgrading getCookie() to be spec compliant #1337

services/system/Transact/src/RTransact.cpp
```diff
@@ -1446,12 +1447,23 @@ std::optional<AccountNumber> RTransact::getUser(HttpRequest request)
             return result;
       }
    }
-   if (auto token = request.getCookie("__Host-SESSION"))
+
+   std::string_view cookieName = isDevChain ? "SESSION" : "__Host-SESSION";
+
+   auto tokens = request.getCookie(cookieName);
+   if (!tokens.empty())
```
swatanabe, line 1454: This is redundant with a loop over `tokens`

services/user/CommonApi/src/CommonApi.cpp
```diff
@@ -79,6 +105,29 @@ namespace SystemService
...
+            request.body.push_back(0);
+            psio::json_token_stream jstream{request.body.data()};
+            auto                    params = psio::from_json<cookie_data>(jstream);
+
+            std::vector<HttpHeader> headers;
+            bool                    isDevChain = psibase::isDevChain(request);
+            std::string             cookieName = "__Host-SESSION";
+            
+            std::string cookieAttribs =
+                "Path=/; HttpOnly; Secure; SameSite=Strict; Expires=" + getCookieExpiration();
```
swatanabe, line 119: `Max-Age` is simpler than `Expires`, because it can just be a literal. It also doesn't assume that the client and server have synchronized clocks.

services/user/CommonApi/src/CommonApi.cpp
```diff
@@ -79,6 +92,29 @@ namespace SystemService
...
+            request.body.push_back(0);
+            psio::json_token_stream jstream{request.body.data()};
+            auto                    params = psio::from_json<cookie_data>(jstream);
+
+            std::vector<HttpHeader> headers;
+            bool                    isLocalhost = psibase::isLocalhost(request);
+            std::string             cookieName  = "__Host-SESSION";
+
+            std::string cookieAttribs =
+                "Path=/; HttpOnly; Secure; SameSite=Strict; Max-Age=" + getCookieMaxAge();
```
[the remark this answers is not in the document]
cool-ant: wow.  great catch! thx

James-Mart (approved):

PR #1297, opened 2025-07-09
wrap memo

Velua:


Velua:
I'm getting the error 

```
    Creating component /root/psibase/services/user/Tokens/target/wasm32-wasip1/release/plugin.wasm
error: failed to decode world from module

Caused by:
    0: module was not valid
    1: module requires an import interface named `env`
make[2]: *** [CMakeFiles/Tokens_ext.dir/build.make:89: Tokens_ext-prefix/src/Tokens_ext-stamp/Tokens_ext-build] Error 1
make[1]: *** [CMakeFiles/Makefile2:2001: CMakeFiles/Tokens_ext.dir/all] Error 2
```

Any ideas?

rust/psibase/src/services/tokens/mod.rs
```diff
@@ -100,17 +93,17 @@ mod service {
     }
 
     #[action]
-    fn credit(tokenId: TID, receiver: AccountNumber, amount: Quantity, memo: String) {
+    fn credit(tokenId: TID, receiver: AccountNumber, amount: Quantity, memo: Memo) {
```
James-Mart, line 96: This doesn't work as well as it does in C++. In C++, the `Memo` type will fail to unpack if the constraint is violated (>80 bytes). In rust, it only works if you construct a `Memo` with `new()`, which won't happen when unpacking.

Therefore, you should impl `validate()` on `Memo`, and every action that takes a memo needs to also call `memo.validate()`

Velua: @James-Mart I thought by making `contents` a private field, there would be no other option in constructing it unless it was with `Self::new` , is that not the case?

Velua: Should it circumvent construction do I wanna panic or return a `Result`?

PR #1298, opened 2025-07-09
Rename max supply in tokens

Velua:
@James-Mart can I get away with this or is a 17 char variable name gonna be an issue? 

`max_supply` is a bit funny now as we're not restricting **supply** but **issued_supply** 

So a case might be max_supply is 1,000, and minting is failing despite supply being only 600 in the event that someone burned 400.


James-Mart:
variable names can be whatever length, this is fine.

James-Mart (approved):

PR #1299, opened 2025-07-09
Fix testnet crash

swatanabe:
We were assuming that `callPreverify` always returns a token for each signature, but that was only true when we found a `preverify` callback. Make the implementation uphold that postcondition consistently.

This caused a non-producing node to fail whenever it received a block with a signed transaction.

### DIFF
--- a/libraries/psibase/native/src/BlockContext.cpp
+++ b/libraries/psibase/native/src/BlockContext.cpp
@@ -415,13 +415,16 @@ namespace psibase
                PSIBASE_LOG(trxLogger, debug) << "preverifyTransaction succeeded";
                if (result)
                {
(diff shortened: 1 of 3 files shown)


James-Mart (approved):

libraries/psibase/native/src/BlockContext.cpp
```diff
@@ -416,14 +415,12 @@ namespace psibase
                PSIBASE_LOG(trxLogger, debug) << "preverifyTransaction succeeded";
                if (result)
                {
-                  for (const auto& token : *result)
+                  for (const auto& [token, out] : std::views::zip(*result, tokens))
```
James-Mart, line 418: ` error: 'zip' is not a member of 'std::views'`

James-Mart (approved):

PR #1308, opened 2025-07-11
Enable clientdata empty records

James-Mart:


swatanabe:
Some explanation for the reason for the change would be appreciated. c.f. https://github.com/gofractally/psibase/pull/799#discussion_r1717248300

James-Mart:
> Some explanation for the reason for the change would be appreciated. cf. https://github.com/gofractally/psibase/pull/799#discussion_r1717248300

Two main reasons:
1) It was bad dx, in most places we were calling set, we were just immediately doing unwrap() or expect() anyways.
2) It actually had the reverse effect that I thought. Rather than prevent empty records, it forces valid use-cases for empty records to actually be non-empty just to appease the api, see [this anti-pattern](https://github.com/gofractally/psibase/blob/main/services/user/Permissions/plugin/src/db.rs#L7)

swatanabe (approved):

PR #1311, opened 2025-07-11
Publish packages faster

swatanabe:
Similar to #1237. We don't need to wait for one package to finish uploading before sending the next one.


cool-ant (approved):
Looks good!

PR #1312, opened 2025-07-14
Adding .tmp/ to excluded dirs when rebuilding

cool-ant:


brandonfancher:
Oh, wait...nevermind. I see that that targets directories too. Looks like`.tmp` works. I'd just remove the last `|| file === '.tsbuildinfo'`.

brandonfancher (approved):
Approving with assumption `|| file === '.tsbuildinfo'` will be removed.

cool-ant:
> `tsconfig.app.tsbuildinfo` and `tsconfig.node.tsbuildinfo` both reside in the `.tmp` directory, so you may not need the last two checks. Sorry if my previous comments were confusing.

of course. derp. Thanks. removed them.

PR #1314, opened 2025-07-14
load all connected accounts when loading an app

James-Mart:
Previously, the supervisor would only load the auth service for the account currently logged-in to the app responsible for calling the supervisor. This means you can't actually sign for anything until you're logged in. This worked fine before, but with authenticated queries, we have a new requirement that a tx must be signed by a user (requiring download of their auth service) *before* they are considered logged-in. In order to accomplish this, the supervisor now automatically downloads the auth service plugin for every account connected to the top-level app.

### DIFF
--- a/services/system/Accounts/plugin/accounts/src/interfaces/api.rs
+++ b/services/system/Accounts/plugin/accounts/src/interfaces/api.rs
@@ -32,15 +32,7 @@ impl API for AccountsPlugin {
         let acct_num = AccountNumber::from_exact(&name).map_err(|err| InvalidAccountName(err.to_string()))?;
 
         let query = format!(
-<<<<<<< Updated upstream
-            "query {{ getAccount(account: \"{}\") {{ accountNum, authService, resourceBalance {{ value }} }} }}",
-=======
-<<<<<<< Updated upstream
-            "query {{ getAccount(account: \"{}\") {{ accountNum, authService, resourceBalance }} }}",
-=======
             "query {{ getAccount(accountName: \"{}\") {{ accountNum, authService, resourceBalance {{ value }} }} }}",
->>>>>>> Stashed changes
->>>>>>> Stashed changes
             acct_num 
         );
 


services/user/Supervisor/ui/src/supervisor.ts
```diff
@@ -87,22 +92,38 @@ export class Supervisor implements AppInterface {
...
-            // Consider deleting the user from the accounts plugin db
-            return;
+        const addtl_plugins: QualifiedPluginId[] = [];
+        for (const service of auth_services) {
+            if (!service) continue;
+            addtl_plugins.push(pluginId(service.authService, "plugin"));
         }
+        this.loader.trackPlugins(addtl_plugins);
 
-        // Current limitation: an auth service plugin must be called "plugin" ("<service>:plugin")
```
cool-ant, line 104: Worth maintaining this comment...

services/system/Accounts/src/RAccounts.cpp
```diff
@@ -22,8 +22,19 @@ namespace SystemService
...
+      auto getAccounts(std::vector<AccountNumber> accounts) const
+      {
+         std::vector<std::optional<Account>> result;
+         for (auto account : accounts)
+         {
+            Accounts::Tables tables{Accounts::service};
+            result.push_back(tables.open<AccountTable>().get(account));
+         }
+         return result;
+      }
```
cool-ant, line 35: This reads like a function that takes accounts and returns account.
maybe simply rename the param to account_nums? Anything here to distinguish input frmo output I think would help a ton with readability

James-Mart: The thing that devs interact with is the schema, does that seem clearer?

```
type ResourceLimit {
    value: String!
}
type Account {
    accountNum: String!
    authService: String!
    resourceBalance: ResourceLimit
}
type Query {
    getAccount(account: String!): Account
    getAccounts(accounts: [String!]!): [Account]!
}
```

James-Mart: Changed parameter names

Issue #1185, opened 2025-05-14
Mac boot tx expiry


James-Mart:
When booting psibase on a mac, it takes a lot longer.

Some people started getting tx expiration on boot, even with it nearly at its maximum (currently at 120 sec, tapos max = 127).

As @swatanabe mentioned [here](https://github.com/gofractally/psibase/pull/1184#discussion_r2087399501), a better solution is not to continue to increase this expiry, but instead just set it to zero during boot and change transact to not check it.

PR #1334, opened 2025-07-23
ignore tx expiration until after boot

James-Mart:
#1185

### DIFF
--- a/services/system/Transact/src/RTransact.cpp
+++ b/services/system/Transact/src/RTransact.cpp
@@ -122,6 +122,17 @@ namespace
       return to<Transact>().checkFirstAuth(id, *trx.transaction);
    }
 
+   bool isBooted()
+   {
+      auto status =
+          Transact::Tables(Transact::service).open<TransactStatusTable>().get(std::tuple{});
+      if (!status)
+         return false;
+      if (!status->bootTransactions)
+         return false;
+      return status->bootTransactions->empty();
+   }
+
 }  // namespace
 
 std::optional<SignedTransaction> SystemService::RTransact::next()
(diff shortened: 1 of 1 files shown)


James-Mart:
Looking at also modifying
`Transact::startBlock`
`Transact::checkTapos`

... Is that right? any others?

services/system/Transact/src/RTransact.cpp
```diff
@@ -122,6 +122,17 @@ namespace
...
 
+   bool isBooted()
+   {
+      auto status =
+          Transact::Tables(Transact::service).open<TransactStatusTable>().get(std::tuple{});
+      if (!status)
+         return false;
+      if (!status->bootTransactions)
+         return false;
+      return status->bootTransactions->empty();
```
swatanabe, line 133: bootTransactions should never be empty.

services/system/Transact/src/RTransact.cpp
```diff
@@ -122,6 +122,17 @@ namespace
...
    }
 
+   bool isBooted()
+   {
+      auto status =
+          Transact::Tables(Transact::service).open<TransactStatusTable>().get(std::tuple{});
+      if (!status)
+         return false;
+      if (!status->bootTransactions)
+         return false;
```
swatanabe, line 132: Should return true. Also, I suspect that it's more reliable to check `status->enforceAuth`.

services/system/Transact/src/RTransact.cpp
```diff
@@ -781,15 +792,18 @@ void RTransact::onBlock()
       }
    }
 
-   PSIBASE_SUBJECTIVE_TX
+   if (isBooted())
```
swatanabe, line 795: It might be better to just exclude the bootTransactions. Other transactions can expire normally.

swatanabe:
> We don't want to allow duplicate transaction IDs. Therefore, we'll need the boot transactions to be rejected after boot finishes, regardless of expiration. The easiest way is to require the boot transactions to have an expiration of 0, or rather the minimum representable time.

PR #1338, opened 2025-07-25
BFT: don't get stuck when there is a view change before we unlock our keys

swatanabe:
Also add a timer loop, when sending a view change fails, because the hardware device might have transient failures and we won't be notified when it's ready again.

### DIFF
--- a/programs/psinode/main.cpp
+++ b/programs/psinode/main.cpp
@@ -1914,7 +1914,7 @@ void run(const std::string&              db_path,
                            });
       };
 
-      http_config->new_key = [&chainContext, &prover, &db_path, &runResult, getPKCS11Slot](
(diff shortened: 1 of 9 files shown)


programs/psinode/main.cpp
```diff
@@ -1948,6 +1948,7 @@ void run(const std::string&              db_path,
                          result = std::make_shared<PKCS11Prover>(pos->second.session, key.service);
                       }
                       pos->second.sessionProver->add(result);
+                      node.on_key_update();
```
James-Mart, line 1951: This not needed in the `key.service == AccountNumber{"verify-sig"}` path?

swatanabe: No. That's the path that stores the key in the config file, which is handled at at L2007.

PR #1342, opened 2025-07-29
Set cookie from plugin

cool-ant:
Accounts app was setting query-token. That's now incorporated into the Accounts plugin's login_direct() function for better encapsulation.

This PR is a follow on to #1291. Once that's merged, the commit history on this PR will be *very* short

### DIFF
--- a/services/system/Accounts/plugin/accounts/src/interfaces/admin.rs
+++ b/services/system/Accounts/plugin/accounts/src/interfaces/admin.rs
@@ -3,11 +3,15 @@ use crate::plugin::AccountsPlugin;
 use crate::bindings::accounts::account_tokens::api::*;
 use crate::bindings::exports::accounts::plugin::admin::{AppDetails, Guest as Admin};
 use crate::bindings::exports::accounts::plugin::api::Guest as API;
+use crate::bindings::host::common::admin as HostAdmin;
 use crate::bindings::host::common::client as Client;
+use crate::bindings::host::common::types::{BodyTypes, PostRequest};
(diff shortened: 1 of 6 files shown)


James-Mart:
> Once that's merged, the commit history on this PR will be very short

You can target your other PR from this one instead of `main`. When the other PR merges, the target of this one will automatically change to `main`. That way this one always has a focused diff

cool-ant:
> > Once that's merged, the commit history on this PR will be very short
> 
> You can target your other PR from this one instead of `main`. When the other PR merges, the target of this one will automatically change to `main`. That way this one always has a focused diff.
> 
> (And this one is probably a draft until that one merges, right?)

right, and it's now merged. and...
image

services/system/Accounts/plugin/accounts/src/interfaces/admin.rs
```diff
@@ -40,10 +49,19 @@ impl Admin for AccountsPlugin {
             TransactAuthApi::get_query_token(&Client::my_service_account(), &user).unwrap();
```
James-Mart, line 49: Temporarily, as a sanity check to simplify auditing authenticated queries, I think you should add in a `println` here, so we see the retrieved token. 

This was previously done in the GUI code, but that part is now being deleted by this pr.

James-Mart (approved):

services/system/Accounts/plugin/accounts/src/interfaces/admin.rs
cool-ant: as you said in Slack, I've done this in the next PR

PR #1345, opened 2025-07-31
Simplify host api

James-Mart:
This PR simplifies the host API for plugins.

It aligns the language with the service APIs - `get-sender` instead of `get-sender-app` and `get-receiver` instead of `my-service-account` (`getSender()` and `getReceiver()` are already standard functions from the C++ service host api).

It also eliminates the `OriginationData` type, which was not helpful. Plugins almost always just care about the sender service. In rare cases where the URL for a particular service is required, the plugin can use `get-app-url`.

### DIFF
--- a/services/system/Accounts/ui/src/hooks/useLoginDirect.ts
+++ b/services/system/Accounts/ui/src/hooks/useLoginDirect.ts
@@ -1,7 +1,7 @@
 import { useMutation } from "@tanstack/react-query";
 import { z } from "zod";
 
-import { getSupervisor } from "@psibase/common-lib";
+import { getSupervisor, siblingUrl } from "@psibase/common-lib";
 
 const LoginParams = z.object({
     app: z.string(),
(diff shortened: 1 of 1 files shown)


services/system/Accounts/plugin/account_tokens/wit/world.wit
```diff
@@ -11,7 +9,15 @@ interface types {
...
 
-    type connection-token = origination-data;
+    /// Data related to the origination of the current call
+    record connection-token {
+        /// The account name of the psibase app responsible for the call.
+        /// May be empty if the call comes from an external web app.
+        app: option<string>,
+
+        /// The domain origin responsible for the call.
+        origin: string
```
swatanabe, line 19: Is the origin still used anywhere?

James-Mart: No, I don't believe so.

James-Mart: This entire connection token object will be removed at some point soon in favor of accounts directly using `host:prompt` to trigger the redirect.

swatanabe (approved):

PR #1346, opened 2025-07-31
Track WasmConfig used for verify services in the block header

swatanabe:
We were using either the default value or the value from the current head block inconsistently, which could cause a consensus failure.

### DIFF
--- a/services/psibase_tests/include/services/test/VerifyWithCall.hpp
+++ b/services/psibase_tests/include/services/test/VerifyWithCall.hpp
@@ -13,5 +13,5 @@ namespace TestService
                                       psibase::Claim       claim,
                                       std::vector<char>    proof);
    };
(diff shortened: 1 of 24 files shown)


services/psibase_tests/include/services/test/VerifyWithCall.hpp
```diff
@@ -0,0 +1,17 @@
...
+{
+   struct VerifyWithCall : psibase::Service
+   {
+      static constexpr auto service = psibase::AccountNumber{"verify-call"};
+      static constexpr auto flags   = psibase::CodeRow::isAuthService;
+      void                  verifySys(psibase::Checksum256 transactionHash,
+                                      psibase::Claim       claim,
+                                      std::vector<char>    proof);
+   };
+   PSIO_REFLECT(VerifyWithCall, method(verifySys, claim, proof))
```
James-Mart, line 16: Missing `transactionHash`

PR #1356, opened 2025-08-08
Restrict sudo when calling a node-local service

swatanabe:
This is an essential security measure before we have multiple local services that call each other. Such services need to reliably know the sender without needing to trust the chain.

Requirements:
- If there is a node-local service for a particular account, subjective execution MUST NOT call the on-chain service (confidentiality and integrity)
- If the sender of an action to a node-local service is a node-local account, then the real sender MUST NOT be an on-chain service.
  - Assumption: All node-local accounts are services

### DIFF
--- a/libraries/psibase/common/include/psibase/nativeTables.hpp
+++ b/libraries/psibase/common/include/psibase/nativeTables.hpp
@@ -78,10 +78,12 @@ namespace psibase
       static constexpr uint64_t forceReplay           = uint64_t(1) << 7;
       static constexpr uint64_t allowSocket           = uint64_t(1) << 8;
       static constexpr uint64_t allowNativeSubjective = uint64_t(1) << 9;
+      static constexpr uint64_t isReplacement         = uint64_t(1) << 10;
 
-      static constexpr auto allFlags =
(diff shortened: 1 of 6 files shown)


libraries/psibase/common/include/psibase/nativeTables.hpp
```diff
@@ -78,6 +78,12 @@ namespace psibase
       static constexpr uint64_t forceReplay           = uint64_t(1) << 7;
       static constexpr uint64_t allowSocket           = uint64_t(1) << 8;
       static constexpr uint64_t allowNativeSubjective = uint64_t(1) << 9;
+      static constexpr uint64_t isReplacement         = uint64_t(1) << 10;
```
James-Mart, line 81: Update needed to `doc/src/specifications/data-formats/package.md`

We really need a doc somewhere that explains what each of these flags are for.

swatanabe:
> Should we require (in wasm) that node-local services are deployed to `x-` accounts?

I don't think it's necessary. The rules above need to be enforced even if the chain doesn't respect the `x-` convention. Also, deploying a service with `isReplacement` will be to a non-`x-` account.

James-Mart (approved):

PR #1359, opened 2025-08-11
integrate permissions plugin with package-template

James-Mart:
Adds example usage of the `permissions` plugin to manage prompting the user to authorize third-party usage of a plugin on the user's behalf


cool-ant (approved):
Man, that's clean!! Nice work!

PR #1370, opened 2025-08-13
Refactor host:common/types out to host:types plugin

cool-ant:


package-templates/basic-01/{{project-name | upper_camel_case}}/plugin/Cargo.toml.liquid
```diff
@@ -10,7 +10,7 @@ publish.workspace = true
 psibase = { path = "../../../../rust/psibase/", version = "{{version}}" }
 serde = { version = "1.0.208", features = ["derive"] }
 serde_json = { version = "1.0.125" }
-wit-bindgen-rt = { version = "0.30.0", features = ["bitflags"] }
+wit-bindgen-rt = { version = "0.44.0", features = ["bitflags"] }
```
James-Mart, line 13: This is not correct, actually. the version of `wit-bindgen-rt` is related to the version of cargo-component used, which was not updated in this pr.

James-Mart:
The `host:types` plugin is missing from the host package

package-templates/basic-01/{{project-name | upper_camel_case}}/plugin/Cargo.toml.liquid
cool-ant: I'm still running cargo-component 0.15.0, and this works fine. We've got versions ranging all the way up to 0.39.0 in various services without trouble.

PR #1373, opened 2025-08-14
Workaround to avoid always rebuilding evaluations

swatanabe:
- ecies indirectly sets features of digest and possibly others
- average sets "libm" feature of num-traits


James-Mart:
Oh thank goodness 🙌

Velua (approved):

PR #1374, opened 2025-08-14
Add x-http service

swatanabe:
x-http is now the entry point for HTTP requests.

### DIFF
--- a/packages/local/XHttp/include/services/local/XHttp.hpp
+++ b/packages/local/XHttp/include/services/local/XHttp.hpp
@@ -1,0 +1,36 @@
+#pragma once
+
+#include <psibase/Service.hpp>
+#include <psibase/Table.hpp>
+#include <services/system/HttpServer.hpp>
+
+namespace LocalService
+{
+   using SystemService::PendingRequestRow;
+   using SystemService::PendingRequestTable;
+
+   struct XHttp : psibase::Service
+   {
+      static constexpr auto service = psibase::AccountNumber{"x-http"};
+
+      using Subjective = psibase::SubjectiveTables<PendingRequestTable>;
+
+      /// Sends a message to a socket. HTTP sockets should use sendReply, instead.
+      void send(std::int32_t socket, psio::view<const std::vector<char>> data);
+
+      /// Enables or disables automatic closing of the socket
+      /// when the transaction context exits.
+      ///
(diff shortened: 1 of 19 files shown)


packages/local/XHttp/src/XHttp.cpp
```diff
@@ -0,0 +1,126 @@
...
+      auto to   = value ? &owned : &requests;
+      auto row  = from->get(socket);
+      if (row && row->owner == sender)
+      {
+         from->remove(*row);
+         to->put(*row);
+      }
+      else
+      {
+         abortMessage(sender.str() + " cannot send a response on socket " + std::to_string(socket));
```
James-Mart, line 46: If `row` is nullptr this is probably the wrong error message

packages/local/XHttp/src/XHttp.cpp
```diff
@@ -0,0 +1,126 @@
...
+{
+   auto sender = getSender();
+   PSIBASE_SUBJECTIVE_TX
+   {
+      auto requests = Subjective{}.open<PendingRequestTable>();
+      auto owned    = Temporary{}.open<PendingRequestTable>();
+
+      auto from = value ? &requests : &owned;
+      auto to   = value ? &owned : &requests;
+      auto row  = from->get(socket);
```
James-Mart, line 38: I would probably expect this function to be idempotent

packages/local/XHttp/include/services/local/XHttp.hpp
```diff
@@ -0,0 +1,36 @@
...
+      /// Sends a message to a socket. HTTP sockets should use sendReply, instead.
+      void send(std::int32_t socket, psio::view<const std::vector<char>> data);
+
+      /// Enables or disables automatic closing of the socket
+      /// when the transaction context exits.
+      ///
+      /// Can be called inside `PSIBASE_SUBJECTIVE_TX`
+      void autoClose(std::int32_t socket, bool value);
+
+      /// Sends an HTTP response. The socket have autoClose enabled.
```
James-Mart, line 27: Missing "must"

packages/local/XHttp/src/XHttp.cpp
swatanabe: In practice it doesn't matter much, because the caller has to keep autoClose in sync with its own tables.

James-Mart (approved):

PR #1388, opened 2025-08-20
Add dynamic service hooks

James-Mart:
# Background

Sometimes a service (aka "parent service") defines a contract fulfilled by another service (aka "child service") that registers with the parent. It can then call actions defined by that interface on arbitrary child services. This could be called the "service hooks" pattern.

The issue is that once such hooks are defined, adding a new hook to the contractual interface and incorporating it into the parent service will break all child services, because the new hook has not yet been added to the child services.

The goal here is to allow for service hooks to be dynamic, so that a contractual hook interface can be updated over time without breaking children. Attempting to call a hook on a child service that is not defined should be a no-op instead of a tx failure.

# Implementation

If there was an intrinsic for checking whether an action exists on a service then I could make a special `ServiceCaller` that no-opped when the specified action dne, which I like mainly because then when the hook is called, the `getSender` would still be the parent service.

Without such an intrinsic, I think I have no choice but to call a well-known, `handle_hook` action, and pass a packed payload, and essentially implement an internal dispatcher within the `handle_hook` action that dispatches (or no-ops) to the actions defined by the hook interface. 

The main downside is that when the actual hook action is called, the caller is the child service:
```
user@parent_service->some_action
parent_service@child_service->handle_hook
child_service@child_service->hook // caller of the hook is the child_service
```

Another way to workaround this issue is if we enabled services to `run_as` their current caller as long as they are calling their own actions, but we are waiting for enabling that until after billing is implemented to ensure that doesn't introduce a vulnerability.

### DIFF
--- a/rust/psibase/src/services/hook_handler.rs
+++ b/rust/psibase/src/services/hook_handler.rs
@@ -1,3 +1,66 @@
+#[macro_export]
+macro_rules! define_service_hooks {
+    ($(($parent_service:expr, $($hook_name:ident),*)),*) => {
+        #[allow(non_camel_case_types)]
+        enum ServiceHook {
+            $(
+                $(
+                    $hook_name {
+                        hook: $crate::MethodNumber,
+                        data: action_structs::$hook_name,
+                    },
+                )*
+            )*
+            Unknown,
+        }
+
+        impl ServiceHook {
+            fn new(hook_name: &str, data: &[u8]) -> Self {
+                use $crate::fracpack::Unpack;
+
+                match hook_name {
+                    $(
+                        $(
+                            action_structs::$hook_name::ACTION_NAME => {
+                                ServiceHook::$hook_name {
+                                    hook: $crate::MethodNumber::from(action_structs::$hook_name::ACTION_NAME),
+                                    data: action_structs::$hook_name::unpacked(data).unwrap(),
+                                }
+                            },
+                        )*
+                    )*
+                    _ => {
+                        ServiceHook::Unknown
+                    }
+                }
+            }
+
+            fn execute(self) {
+                let my_service = get_service();
+                let s = ServiceCaller {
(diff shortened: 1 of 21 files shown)


swatanabe:
> the trace shows the hook that was called (better), but the sender is the child service (worse)

I think the trace should be considered a debugging tool only. We shouldn't let it drive design decisions.

> The issue is that once such hooks are defined, adding a new hook to the contractual interface and incorporating it into the parent service will break all child services, because the new hook has not yet been added to the child services.

Another way to solve this is to version the hook interface and specify the version when registering.

rust/psibase/src/services/hook_handler.rs
```diff
@@ -0,0 +1,88 @@
...
+#[allow(non_snake_case, unused_variables)]
+pub mod HookHandler {
+    #[action]
+    fn handle_hook(hook_name: String, hook_data: Vec<u8>) {
+        unimplemented!()
+    }
+}
+
+/// Syntactic sugar for calling dynamic hooks
+pub fn call_hook<T>(service: crate::AccountNumber, data: T)
```
swatanabe, line 82: What if you define a type that implements `Caller` and let the macro define
```
pub fn call_hook(service: AccountNumber) -> Actions<HookCaller>;
```

rust/psibase/src/services/hook_handler.rs
```diff
@@ -0,0 +1,88 @@
...
+        }
+
+        impl ServiceHook {
+            fn new(hook_name: &str, data: &[u8]) -> Self {
+                use $crate::fracpack::Unpack;
+
+                match hook_name {
+                    $(
+                        $(
+                            action_structs::$hook_name::ACTION_NAME => {
```
swatanabe, line 24: `ActionMeta` might not be in scope

PR #1400, opened 2025-08-27
Remove platform specific references from plugins

James-Mart:
This removes platform specific references from the plugins.

The host plugins were referring to the "supervisor" as the native host, but now they refer to "Platform" which could be supervisor on web, or something else on a different platform.

Some other plugins were gating functionality unless the caller was "supervisor", but they shouldn't have to hardcode each platform, so there is a new `is_caller_host` function in `host:common` which checks if the caller is either `host` or the account name reported by the platform (e.g. "supervisor").


swatanabe:
I'm not sure I agree with making "supervisor" web-specific. For other platforms, and the CLI specifically, it will probably not be served from the chain at all, and needs a reserved account name that won't conflict with anything else.

James-Mart:
The native layer may have plugins that it wants to load from the chain, and those plugins should run with host privileges. Platforms can't *actually* share a namespace in which to serve assets, so they will need to have their own account. 

So how about a function like:
`get-priveleged-accounts` - and the host can specify a handful of accounts that should be considered of equal privilege to the host (return `true` if a plugin asks `is_sender_host()`).

Does that work for you?

James-Mart:
> I wouldn't object to this is it came for free, but I don't see the point of adding intrinsics to support something that speculative.

I guess we can wait if/until there's a usecase

PR #1401, opened 2025-08-27
Automatically adjust CFT timeout

swatanabe:
The network can be very unstable if the timeout is too low, but it's slow to recover when it's too high. Try to figure out a good value dynamically.

### DIFF
--- a/libraries/net/include/psibase/auto_timeout.hpp
+++ b/libraries/net/include/psibase/auto_timeout.hpp
@@ -1,0 +1,47 @@
+#pragma once
+
+#include <algorithm>
+
+namespace psibase::net
+{
+   // Automatically adjusts a timeout based on the observed network latency.
+   // The timeout increases by a factor of 1.5 every time it expires,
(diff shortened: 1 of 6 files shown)


libraries/net/include/psibase/auto_timeout.hpp
```diff
@@ -0,0 +1,50 @@
...
+   // is significantly lower.
+   template <typename Clock>
+   class auto_timeout
+   {
+     public:
+      explicit auto_timeout(typename Clock::duration initial_value) : _current(initial_value)
+      {
+         start();
+      }
+      void stop() {}
```
James-Mart, line 19: `stop` is a no-op?

James-Mart (approved):

libraries/net/include/psibase/auto_timeout.hpp
swatanabe: It doesn't actually need to do anything. I mainly used it as a hook to put debugging code or breakpoints.

PR #1408, opened 2025-09-03
Add token stream

Velua:
Adds new service, plugin and UI for exponential decay based token streams. 
Users are able to create streams of any token which rewards them with a NFT which is used to redeem the tokens vested in a stream.
Any user can deposit into stream with it's relevant token. 
Streams are deletable once their balances are fully vested + claimed. 

Adds `start_block_after` to tester module.


packages/user/TokenStream/ui/src/components/account-switcher.tsx
James-Mart, line 1: Not a shared component?

packages/user/TokenStream/service/src/lib.rs
```diff
@@ -0,0 +1,306 @@
...
+                .into()
+        }
+
+        /// Deposit `amount` into the bucket.
+        pub fn deposit(&mut self, amount: Quantity) {
+            self.claimable_at_last_deposit = self.total_vested();
+            self.total_deposited = self
+                .total_deposited
+                .value
+                .saturating_add(amount.value)
```
swatanabe, line 117: If this actually saturates, only the amount that is actually added should be debited from the shared balance. Another option is to decrement `total_deposited` by `total_claimed`, and reset `total_claimed` to zero, which makes overflow impossible. Note that overflow is only possible by cycling the same tokens through the stream repeatedly.

packages/user/TokenStream/service/src/lib.rs
```diff
@@ -0,0 +1,306 @@
...
+                .into()
+        }
+
+        pub fn total_vested(&self) -> Quantity {
+            if self.total_deposited.value == 0 {
+                return self.claimable_at_last_deposit;
+            }
+            let now = TransactSvc::call().currentBlock().time.seconds().seconds;
+            let delta_seconds = (now - self.last_deposit_timestamp.seconds) as f64;
+            let factor = (-self.decay_rate() * delta_seconds).exp();
```
swatanabe, line 148: `exp` has unspecified precision. It's almost certainly acceptable, but we should decide how much precision we actually require and add a few tests with hard-coded constants to make sure. Integer multiples of the half-life are easy to verify.

packages/user/TokenStream/service/src/tests.rs
```diff
@@ -0,0 +1,396 @@
...
+        TokenStream::push_from(&chain, BOB)
+            .claim(stream_nft_id)
+            .get()
+            .unwrap();
+
+        check_balance(&chain, token_id, BOB, Some(1));
+
+        chain.start_block_at(
+            TimePointSec {
+                seconds: 3600 * 7 * 52 * 20,
```
swatanabe, line 151: What is this supposed to represent? We should be able to calculate the time at which the remaining unvested balance flushes to zero.

packages/user/TokenStream/service/src/tests.rs
```diff
@@ -0,0 +1,396 @@
...
+
+        Nfts::push_from(&chain, ALICE).credit(stream_nft_id, BOB, "memo".to_string());
+        Nfts::push_from(&chain, BOB).debit(stream_nft_id, "memo".to_string());
+
+        TokenStream::push_from(&chain, BOB)
+            .claim(stream_nft_id)
+            .get()
+            .unwrap();
+
+        check_balance(&chain, token_id, BOB, Some(1));
```
swatanabe, line 147: It's one because the transactions automatically create blocks, which causes some time to pass, right? Consider turning off automatic block start, to check that the amount claimable in the same block as the deposit is zero.

packages/user/TokenStream/service/src/tests.rs
Velua: 02ea0d01faef7cdef9d0c37a1bacd2525ff7044c

Velua: 02ea0d01faef7cdef9d0c37a1bacd2525ff7044c

packages/user/TokenStream/service/src/tests.rs
```diff
@@ -90,6 +90,14 @@ mod tests {
         Ok(token_id)
     }
 
+    pub fn full_vesting_time(half_life_seconds: u32, balance: u64) -> i64 {
+        if balance == 0 {
+            return 0;
+        }
+        let t = half_life_seconds as f64 * ((2.0 * balance as f64).ln() / std::f64::consts::LN_2);
```
swatanabe, line 97: Why is multiplying by 2.0 needed? The unvested balance is rounded down, so it should flush to zero when it drops below 1, not when it drops below 0.5. Also, there's `log2`.

packages/user/TokenStream/plugin/wit/world.wit
```diff
@@ -3,9 +3,37 @@ package token-stream:plugin;
 interface api {
     use host:types/types.{error};
 
+
+    /// Creates a new token stream
+    ///
+    /// Parameters:
+    /// * `half-life-seconds`: Half life of the vesting rate in seconds
```
swatanabe, line 10: I wouldn't say that it's half life of the vesting _rate_. Even though derivatives of all orders have the same half life, it's not the normal way to think about it.

packages/user/TokenStream/service/src/tests.rs
```diff
@@ -0,0 +1,413 @@
...
+                .to_string(),
+            "service 'tokens' aborted with message: credit quantity must be greater than 0"
+                .to_string()
+        );
+
+        check_balance(&chain, token_id, BOB, Some(0));
+
+        let expected_full_vesting_wait_time = full_vesting_time(10000, 500);
+
+        chain.start_block_at(
```
swatanabe, line 166: Should this be `start_block_after`? The initial time set by `reset_clock` doesn't seem to be taken into account.

PR #1412, opened 2025-09-05
Transact snapshot improvements

James-Mart:
* Uses u32 for `setSnapTime` which ensures that it can be returned from queries as a json number instead of getting stringified.
* Adds a query for snapshot info
* Fixes transact plugin, which wasn't easily able to call tx on itself previously


James-Mart:
CC @Velua

swatanabe (approved):

PR #1416, opened 2025-09-09
add debit_up_to

Velua:
Adds `debit_up_to` action to `tokens` service.  
Useful for cases where a service wants to debit the entire balance or potential amount of from another, action returns actual amount being debited.

### DIFF
--- a/packages/user/Tokens/service/src/lib.rs
+++ b/packages/user/Tokens/service/src/lib.rs
@@ -348,7 +348,7 @@ pub mod service {
 
     /// Debit
     ///
-    /// Debits tokens from a shared balance between the creditor and the debitor (sender)
+    /// Debits exact amount of tokens from a shared balance between the creditor and the debitor (sender)
     ///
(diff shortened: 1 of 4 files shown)


packages/user/Tokens/service/src/lib.rs
```diff
@@ -363,6 +363,27 @@ pub mod service {
         SharedBalance::get_assert(creditor, get_sender(), token_id).debit(amount, memo);
     }
 
+    /// Debit up to
+    ///
+    /// Debits up to X amount of tokens from a shared balance between the creditor and the debitor (sender)
+    ///
+    /// Action will return the amount of tokens actually debited. Set amount to 0 for entire shared balance.
```
swatanabe, line 370: I don't think using the existing balance to determine the amount to debit is a good idea. The caller of debit should be the primary source of truth of the amount to debit. Making credit primary risks confusion if there is an existing shared balance for any reason.

PR #1425, opened 2025-09-15
Mm/extract gen keypair to psibase

cool-ant:


James-Mart:
The idea of this pr is to move key generation into psibase so it can be more easily reused both from plugins and other tooling (like the xadmin "plugin" that must work in a state without a booted network).

In the future the generate-keypair can be modified (or a new function can be added) to return the randomly generated seed phrase used to derive a key, rather than directly returning the private key.

James-Mart (approved):

PR #1428, opened 2025-09-16
Move memo to psibase

Velua:
Rather than have `Memo` in the `Tokens` module it's moved to `psibase` so it can be used in other contexts.


James-Mart (approved):

James-Mart (changes requested):
steven's feedback

PR #1430, opened 2025-09-17
add difficulty

Velua:


packages/user/DiffAdjust/service/src/lib.rs
```diff
@@ -0,0 +1,349 @@
...
+        pub fn increment(&mut self, increment_amount: u32) -> u64 {
+            self.check_sender_is_consumer();
+            let difficulty = self.check_difficulty_decrease();
+            self.counter += increment_amount;
+            self.check_difficulty_increase();
+            self.save();
+            difficulty
+        }
+
+        pub fn update_window(&mut self, seconds: u32) {
```
swatanabe, line 172: None of the parameter update functions call `check_difficulty_[increase|decrease]`, which means that they can retroactively alter increases and decreases that were previously reported by the query.

packages/user/DiffAdjust/service/src/lib.rs
```diff
@@ -0,0 +1,349 @@
...
+    /// Update targets
+    ///
+    /// * Requires holding administration NFT.
+    ///
+    /// # Arguments
+    /// * `nft_id` - RateLimit / NFT ID
+    /// * `target_min` - Minimum target difficulty
+    /// * `target_max` - Maximum target difficulty
+    #[action]
+    fn up_targets(nft_id: u32, target_min: u32, target_max: u32) {
```
swatanabe, line 296: `up` might not be the best prefix. My first thought when I saw it was that it adds to the current values rather than setting them.

rust/psibase/src/services/diff_adjust.rs
```diff
@@ -0,0 +1,114 @@
...
+        unimplemented!()
+    }
+
+    /// Update percent change
+    ///
+    /// * Requires holding administration NFT.
+    ///
+    /// # Arguments
+    /// * `nft_id` - RateLimit / NFT ID
+    /// * `percent_ppm` - Percent ppm 50000 = 5%
```
James-Mart, line 93: Just called `ppm`

packages/user/DiffAdjust/service/src/lib.rs
Velua: We're now checking for price decreases on the updates, price increase isn't necessary given it runs during a price increase

swatanabe: Suppose that `target_max = 1000`. If I change it to `target_max = 10`, it could trigger up to 100 difficulty increases. Is that the desired behavior? Considering that `target_max` was `1000` at the time all those increments happened, it seems a bit counterintuitive. Also, exponential increase means that large changes like this are liable to overflow. I think it would be better if decreasing the `target_max` results in at most one immediate difficulty increase.

Velua: I've added clamping `fn check_difficulty_increase(&mut self, clamp_increase: bool) -> u64 {`

swatanabe (approved):

Issue #1447, opened 2025-09-24
xadmin - only query for producers after boot


James-Mart:
Querying before boot generates a server error (500), and 500 errors are not CORS enabled (so x-admin subdomain will see CORS errors in the console).

swatanabe:
Well, the difference is that since it's a specific error generated by wasm, we can add CORS headers if needed.

PR #1448, opened 2025-09-24
XAdmin: Avoid querying producers unless we've successfully booted

cool-ant:
Fixes https://github.com/gofractally/psibase/issues/1447

### DIFF
--- a/packages/local/XAdmin/ui/src/hooks/useProducers.ts
+++ b/packages/local/XAdmin/ui/src/hooks/useProducers.ts
@@ -12,7 +12,7 @@ export const useProducers = () => {
         queryKey: queryKeys.producers,
         queryFn: async () => {
             if (
-                ["startup", "needgenesis"].includes(status?.[0] ?? "") ||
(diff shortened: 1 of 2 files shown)


packages/local/XAdmin/ui/src/hooks/useProducers.ts
```diff
@@ -4,11 +4,21 @@ import { type Producer, getProducers } from "@/lib/graphql/getProducers";
...
 import { useConfig } from "./useConfig";
+import { useStatuses } from "./useStatuses";
 
 export const useProducers = () => {
+    const { data: status, error: statusError } = useStatuses();
     return useQuery<Producer[]>({
         queryKey: queryKeys.producers,
         queryFn: async () => {
+            if (
+                ["startup", "slow", "needgenesis"].includes(
```
James-Mart, line 15: `Slow` doesn't seem right here, does it? And I assume startup is a status while starting up?

James-Mart:
It's weird, when I first open the tab, I get the same error, but if I refresh the X-Admin page with a tab already open, I don't see it

PR #1450, opened 2025-09-25
add sites table rust bindings

James-Mart:
Allows rust services to get direct kv handles to sites tables

### DIFF
--- a/packages/user/Sites/include/services/user/Sites.hpp
+++ b/packages/user/Sites/include/services/user/Sites.hpp
@@ -8,12 +8,12 @@ namespace SystemService
    using SitesContentKey = std::tuple<psibase::AccountNumber, std::string>;
    struct SitesContentRow
    {
-      psibase::AccountNumber     account         = {};
-      std::string                path            = {};
-      std::string                contentType     = {};
-      psibase::Checksum256       contentHash     = {};
-      std::optional<std::string> contentEncoding = std::nullopt;
-      std::optional<std::string> csp             = std::nullopt;
+      psibase::AccountNumber     account;
+      std::string                path;
+      std::string                contentType;
+      psibase::Checksum256       contentHash;
+      std::optional<std::string> contentEncoding;
+      std::optional<std::string> csp;
 
       using Key = psibase::CompositeKey<&SitesContentRow::account, &SitesContentRow::path>;
    };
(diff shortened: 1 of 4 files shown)


packages/user/Sites/include/services/user/Sites.hpp
```diff
@@ -44,11 +44,11 @@ namespace SystemService
    struct SiteConfigRow
    {
       psibase::AccountNumber     account;
-      bool                       spa       = false;
-      bool                       cache     = true;
-      std::optional<std::string> globalCsp = std::nullopt;
+      bool                       spa;
+      bool                       doNotCache;
```
swatanabe, line 48: > invert site config cache logic so default value can be false

The tail is wagging the dog here. Negative boolean variables tend to be more confusing to read because they result in double negatives.

packages/user/Sites/include/services/user/Sites.hpp
```diff
@@ -44,11 +44,11 @@ namespace SystemService
    struct SiteConfigRow
    {
       psibase::AccountNumber     account;
-      bool                       spa       = false;
```
swatanabe, line 47: For types with a trivial default constructor, default initialization is not the same as zero initialization.

rust/psibase/src/schema.rs
```diff
@@ -156,7 +156,20 @@ pub(crate) fn assert_schemas_equivalent(lhs: &Schema, rhs: &Schema) {
...
+
+            // Compare tables by their declared numeric `table` index
+            let l_by_idx: std::collections::HashMap<u16, &TableInfo> =
+                ltables.iter().map(|t| (t.table, t)).collect();
+
+            for rtab in rtables {
+                let Some(ltab) = l_by_idx.get(&rtab.table) else {
+                    let table_name = rtab.name.as_ref().map_or("<unnamed>", |n| n.as_str());
+                    panic!(
+                        "Missing table {}::{} (index {})",
```
swatanabe, line 168: The error should say that it's an extra table, not a missing one.

rust/psibase/src/schema.rs
```diff
@@ -156,7 +156,20 @@ pub(crate) fn assert_schemas_equivalent(lhs: &Schema, rhs: &Schema) {
             if ltables.len() < rtables.len() {
                 panic!("Extra tables in {}", db);
             }
```
swatanabe, line 158: This check is redundant, and the error later on identifies the specific table.

packages/user/Sites/include/services/user/Sites.hpp
James-Mart: Reverted

packages/user/Sites/include/services/user/Sites.hpp
James-Mart: Reverted

PR #1455, opened 2025-09-26
Misc cmakelists fixes

James-Mart:
1. Evaluations package was missing from proddefault
2. ~~xadmin_js dependencies should have been using `${XAdmin_js_DEP}` instead of `XAdmin_js` (causing issues with detecting changes in xadmin)~~
3. authsig dep re-added to producers package. - Until a better solution is found, this is necessary to ensure verify-sig is published before the setproducers call (which uses verify-sig as the producer claim)


James-Mart:
Just realized the xadmin_js deps item was already proposed in https://github.com/gofractally/psibase/pull/1448/files

PR #1457, opened 2025-09-29
add event queries to tokens service

Velua:
Adds missing event queries to the tokens service, `from` was renamed to `recallee` as `from` doesn't appear to work in the sqlite queries, assuming it's a reserved keyword.


swatanabe:
> as from doesn't appear to work in the sqlite queries, assuming it's a reserved keyword

Any string can be used as an identifier by putting it in "double quotes"

PR #1464, opened 2025-10-01
Add candidate producer registration to config app

James-Mart:
config app now allows registration of producer candidates. 
Setting the producers in the config app now only requires specifying the candidates by account name.


packages/system/Producers/include/services/system/Producers.hpp
```diff
@@ -34,6 +48,7 @@ namespace SystemService
 
       void setConsensus(psibase::ConsensusData consensus);
       void setProducers(std::vector<psibase::Producer> prods);
+      void regCandidate(const std::string& endpoint, psibase::Claim claim);
```
swatanabe, line 51: Should a candidate be able to unregister?

packages/user/Config/ui/src/pages/block-production.tsx
```diff
@@ -20,31 +22,95 @@ import {
...
+    TableCell,
+    TableHead,
+    TableHeader,
+    TableRow,
+} from "@shared/shadcn/ui/table";
 import { Textarea } from "@shared/shadcn/ui/textarea";
 
-const useProducersPlugin = (method: string) =>
+const RegisterCandidateParams = z.object({
+    endpoint: z.string().min(1, "Endpoint is required"),
```
Velua, line 36: Might be helpful
```suggestion
    endpoint: z.string().url().min(1, "Endpoint is required"),
```

Velua (approved):
I think we could later delete the first Producer name form and just allow the user to select rows in the Registered candidates table.

PR #1472, opened 2025-10-02
Tune cicd cache sizes

James-Mart:


James-Mart:
I'm not entirely sure what the disk savings are from this, but I propose we merge this, then on a subsequent cicd build I will check the cache size (on the runner's disk vs the uploaded cache) and then either keep this or revert it.

PR #1474, opened 2025-10-03
Mm/shared cache rust

cool-ant:


.github/workflows/ubuntu-builder.yml
```diff
@@ -33,9 +33,16 @@ jobs:
...
           path: .caches
-          key: $ubuntu-${{ matrix.ubuntu-version }}-caches-${{ steps.ccache_cache_timestamp.outputs.timestamp }}
+          key: ubuntu-${{ matrix.ubuntu-version }}-caches-${{ hashFiles('**/Cargo.lock') }}
           restore-keys: |
-            $ubuntu-${{ matrix.ubuntu-version }}-caches-
+            ubuntu-${{ matrix.ubuntu-version }}-caches-
+      - name: Preserve Rust artifacts
+        uses: actions/cache@v3
+        with:
+          path: build/rust
```
swatanabe, line 42: Preserving the build directory is probably a bad idea. It will grow indefinitely, since cargo doesn't clean up old build artifacts.

PR #1475, opened 2025-10-03
Invite credentials

James-Mart:


packages/user/Invite/test/src/Invite-test.cpp
```diff
@@ -62,16 +63,16 @@ namespace
...
        "cGky3DiVxoHkPFfGo81ZkIDBk3NOMSFiw5hZiscqwtY464TFBKdc8evq\n"
        "-----END PRIVATE KEY-----\n");
+
+   uint32_t inviteId = 1;
 }  // namespace
 
 // Helper functions for tests
 auto createInvite = [](auto& user, const auto& pubKey)
-{ return user.createInvite(pubKey, std::nullopt, std::nullopt, std::nullopt, std::nullopt); };
+{ return user.createInvite(++inviteId, pubKey, 1, false, ""); };
```
swatanabe: This will cause slightly different behavior depending on which tests are run in what order, because `inviteId` is global.

packages/user/Invite/include/services/user/Invite.hpp
```diff
@@ -13,102 +13,79 @@ namespace UserService
...
+                               std::string secret);
+
+         /// Called by an invite credential (not a user account) to create the specified
+         /// account. The new account is authorized by the specified public key.
+         void createAccount(psibase::AccountNumber account, Spki accountKey);
+
+         /// The sender accepts an invite.
+         /// Calling this action also requires that the sender authorizes the transaction with the
+         /// proof for the credential associated with an invite.
+         void accept();
```
swatanabe: I don't think deducing the invite id from the transaction is a good idea.
- It isn't obvious what is being accepted
- It makes some dodgy actions that appear safe on their own exploitable
  - Removing a signature is harmless because it'll just make the transaction fail, right?
  - Adding a signature with an explicit private key is safe, right?
  - Alas, the combination of the above allows someone to force this version of accept to accept the wrong invite.

packages/user/Invite/src/Invite.cpp
```diff
@@ -1,40 +1,81 @@
...
       check(initRecord.has_value(), UserService::Errors::uninitialized);
    }
+
+   if (getSender() == Credentials::CREDENTIAL_SENDER)
+   {
+      if (m != "createAccount"_m)
+      {
+         string error = Credentials::CREDENTIAL_SENDER.str() + canOnlyCallCreateAccount.data();
+         abortMessage(error);
+      }
```
swatanabe: This looks error prone. I think it would be better for Credentials::create to take a whitelist of actions, so it's impossible to accidentally forget.

packages/user/Invite/plugin/src/db.rs
```diff
@@ -1,20 +1,277 @@
...
+        let details = Self::get_current_details(invite_id);
+        if details.is_none() {
+            return false;
+        }
+        details.unwrap().numAccounts > 0
+    }
+
+    pub fn account_created() {
+        let invite_id = Self::active_invite_id().expect("No active invite token");
+        if let Some(device) = Self::device_memory().get(&invite_id.to_string()) {
```
swatanabe: This pattern can be simplified by using `map_or` with `Default`.

packages/user/Invite/plugin/src/db.rs
```diff
@@ -1,20 +1,277 @@
...
+            Database {
+                mode: DbMode::NonTransactional,
+                duration: StorageDuration::Session,
+            },
+            "invites",
+        )
+    }
+
+    // Details about this device's interactions with invites.
+    fn device_memory() -> Bucket {
```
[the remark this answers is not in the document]
James-Mart: renamed to device_history

PR #1500, opened 2025-10-14
Configure hakari for optimized builds

cool-ant:
Hakari creates common dependencies between projects that otherwise call for feature flag differences in compiled dependencies. Homogenizing versions allows for *much* greater reuse, as these feature flag difference in versions trickle through and impact everything downstream of them in the build.

Benchmark seems to confirm ~5.5GB saved and
1) on a fresh build: Rust build runs in ~56% the time it does on `main`
2) same time for a no-op rebuild, i.e., nothing gets inappropriately rebuilt once a build completes.

NOTE: hakari will need to be added to the img-builder before this gets merged.


packages/user/.config/hakari.toml
```diff
@@ -0,0 +1,27 @@
...
+# For more about the new feature resolver, see:
+# https://blog.rust-lang.org/2021/03/25/Rust-1.51.0.html#cargos-new-feature-resolver
+resolver = "2"
+
+# Add triples corresponding to platforms commonly used by developers here.
+# https://doc.rust-lang.org/rustc/platform-support.html
+platforms = [
+    "x86_64-unknown-linux-gnu",  # Linux x86_64 (primary dev platform)
+    "aarch64-unknown-linux-gnu", # Linux ARM64 (container platform, ARM servers)
+    "wasm32-wasip1",             # CRITICAL: psibase services compile to WebAssembly
```
James-Mart: Is `wasm32-wasip1` needed here?
It's a build target, but this looks like it's supposed to be a list of host platforms.

cool-ant:
Ok, given this is approved, I've create a new PR off of this branch with the workspace hack added to the package-templates basic-01 template. https://github.com/gofractally/psibase/pull/1510

PR #1508, opened 2025-10-20
Add Fractal exile table

Velua:
Rather than maintain exiled members in the `FractalMember` with a status of `Exiled` they are instead removed from the table entirely and added to the `FractalExile` table. 

This also bulk removes other table instances a Fractal member might leave behind like `GuildApplication`, `GuildAttest` and `GuildMember`

### DIFF
--- a/packages/user/FractalCore/plugin/wit/world.wit
+++ b/packages/user/FractalCore/plugin/wit/world.wit
@@ -59,6 +59,7 @@ interface user-guild {
 world imports {
     import user-guild;
     import admin-guild;
+    import admin-fractal;
     import user-fractal;
     import user-eval;
 }


packages/user/Fractals/service/src/tables/fractal_member.rs
```diff
@@ -70,11 +77,17 @@ impl FractalMember {
         MemberStatus::Visa == self.member_status.into()
     }
 
-    pub fn check_has_visa_or_citizenship(&self) {
-        check(
-            self.has_visa() || self.is_citizen(),
-            "must have visa or citizenship",
-        );
+    pub fn exhile(&self) {
```
James-Mart: exhile => exile

packages/user/Fractals/service/src/lib.rs
```diff
@@ -358,6 +350,15 @@ pub mod service {
         Guild::add(fractal, guild_account, get_sender(), display_name);
     }
 
+    /// Exile a fractal member.
+    ///
+    /// # Arguments
+    /// * `member` - The fractal member to be exiled.
+    #[action]
+    fn exile_member(member: AccountNumber) {
```
James-Mart: As mentioned in slack, 

>  I think the sender should be the legislative account. And the action could also just take the fractal as a param, so it can just have (something like) `assert(getSender() == FractalTable::get(fractal).legislature)`. I don't think you want to look up the fractal based on the sender, otherwise you will need a another index on the fractal table.

PR #1529, opened 2025-10-28
Exclude solo plugins from package workspaces

James-Mart:
Because of the nested workspaces and the fact that cargo looks for the nearest container workspace when building a leaf crate, "solo plugins" (those not otherwise included in a cargo_psibase_package) fail to build directly with `cargo component build -r` unless they are excluded from the user/system package workspaces.


cool-ant (approved):
just the one question. otherwise
LGTM

PR #1544, opened 2025-11-04
Implement HTTP requests from local services

swatanabe:


packages/psibase_tests/src/XProxy.cpp
```diff
@@ -0,0 +1,129 @@
...
+   HttpReply        reply{.status      = HttpStatus::badGateway,
+                          .contentType = "text/html",
+                          .body        = std::vector(msg.begin(), msg.end())};
+   PSIBASE_SUBJECTIVE_TX
+   {
+      if (auto row = table.get(socket))
+      {
+         to<XHttp>().autoClose(row->socket2, true);
+         to<XHttp>().sendReply(row->socket2, std::move(reply));
+         table.remove(*row);
```
James-Mart: Is it possible for neither `onReply` nor `onError` to ever be called?

packages/psibase_tests/src/XProxy.cpp
```diff
@@ -0,0 +1,129 @@
...
+            requests.put({upstream, *socket});
+         }
+         return {};
+      }
+   }
+   return {};
+}
+
+void XProxy::onReply(std::int32_t socket, psibase::HttpReply reply)
+{
```
James-Mart: Should this (and `onError`) assert that the sender is `XHttp`?

packages/psibase_tests/src/XProxy.cpp
```diff
@@ -0,0 +1,129 @@
...
+             method(onReply, socket, reply),
+             method(onError, socket))
+PSIBASE_REFLECT_TABLES(XProxy, XProxy::Subjective, XProxy::Session)
+
+std::optional<HttpReply> XProxy::serveSys(HttpRequest req, std::optional<std::int32_t> socket)
+{
+   check(getSender() == XHttp::service, "Wrong sender");
+   auto target = req.path();
+
+   if (target == "/set_origin_server")
```
James-Mart: I guess anyone can set the origin server if this package is installed?

PR #1549, opened 2025-11-04
increase pr limit in changelog template

James-Mart:


.github/workflows/changelog-template.json
```diff
@@ -86,5 +86,6 @@
                 }
             ]
         }
-    ]
+    ],
+    "max_pull_requests": 300,
```
swatanabe: This is invalid JSON. I don't know whether the tool cares.

PR #1553, opened 2025-11-05
Transact -- Add permissions config

cool-ant:


packages/system/Transact/plugin/src/lib.rs
```diff
@@ -30,11 +30,32 @@ use exports::transact::plugin::{
...
+psibase::define_trust! {
+    descriptions {
+        Low => "",
+        Medium => "",
+        High => "
+        🚨 WARNING 🚨 
+        This approval will grant the caller the ability to control which of your accounts takes requested actions, including the capability to make unintended use of any of your accounts! Make sure you completely trust the caller's legitimacy.
+
+        High trust grants these abilities:
+            - Hook actions sender
```
James-Mart: The warning might be good enough on this one, I'm not sure the user will really learn anything from this list because they won't understand what "hook actions sender" means

PR #1565, opened 2025-11-06
Registry -- Add permissions config

cool-ant:


packages/user/Registry/plugin/src/lib.rs
James-Mart: Ah, we may have forgotten during the audit - all of the functions in here should whitelist workshop

James-Mart (approved):

PR #1567, opened 2025-11-06
Tokens -- Add permissions config

cool-ant:


packages/user/Tokens/plugin/src/lib.rs
```diff
@@ -23,6 +23,31 @@ pub mod query {
...
+            - Issue and manage token supply
+            - Manage token transfer settings
+            - Transfer tokens on your behalf
+        ",
+    }
+    functions {
+        None => [decimal_to_u64, u64_to_decimal],
+        Low => [],
+        Medium => [create, enable_user_manual_debit, enable_user_keep_zero_balances, enable_balance_manual_debit, enable_balance_keep_zero_balances, del_balance_config],
+        High => [recall, mint, map_symbol, enable_token_untransferable, enable_token_unrecallable, credit, uncredit, debit, reject, burn],
```
swatanabe: `enable_balance_manual_debit` and `debit` should probably have the same level.

cool-ant: Does `enable_balance_manual_debit` act automatically on standing credits when it's set to false? if not, I'd argue it it fits the Medium desc. If it action results when it's flipped, I agree.
Then it's a High trust thing to do the debit (currently and after flipping this flag), but it's Medium to just flip the setting.
@James-Mart ?

James-Mart: It only affects future debits. So technically it does not immediately affect others, while `debit` immediately does (which is part of the criteria that distinguishes Medium from High).

But I can honestly see it either way, since changing manual debit affects others' *future* interactions with you.

cool-ant: I'd argue that the relevant action for the effect we're discussing is the other person executing a "send". *That's* where the permissions matters. You tweaking your own settings is your business.

swatanabe: Those aren't really hard and fast rules. The main point is how contained the damage is if you get it wrong. Tokens are used to trade for on-chain resources, use of other services, and possibly even real-world goods and services. Thus, the effects of token transfers are not contained at all. The level associated with changing a setting should include the long-term effects of that setting. You can make a reasonable case that enabling auto-debit can be medium, but in that case, debit itself should also be medium.

PR #1576, opened 2025-11-08
Authenticated token queries

James-Mart:
This PR:
* Fixes a lot of bugs related to authenticated queries
* Adds the first actual example of queries being restricted by user (tokens, e.g. `balance`, `balChanges`, etc.).
* To allow the wallet on the homepage to continue to function, these "private" queries are sent through the tokens plugin rather than direct to the tokens host using fetch, which allows for the logged-in-user's query token to be attached to the request.
* C++ tester now has a `login` function that returns the token, and `get` and `post` now take optional token params

### DIFF
--- a/programs/psinode/tests/psinode.py
+++ b/programs/psinode/tests/psinode.py
@@ -150,16 +150,10 @@ class TransactionError(Exception):
         self.trace = trace
 
 class GraphQLError(Exception):
-    def __init__(self, json):
-        errors = json.get('errors', [])
-        if errors and isinstance(errors, list):
-            message = errors[0].get('message', str(errors[0]))
-        elif isinstance(errors, dict):
-            message = errors.get('message', str(errors))
-        else:
-            message = str(errors)
-        super().__init__(message)
-        self.json = json
+    def __init__(self, payload):
+        msg = payload["errors"][0].get("message", "GraphQL error")
+        super().__init__(msg)
+        self.json = payload
 
 class PrivateKey:
     def __init__(self, data=None):


libraries/psibase/common/include/psibase/jwt.hpp
```diff
@@ -9,12 +10,21 @@ namespace psibase
...
 
    template <typename T>
-   T decodeJWT(const JWTKey& key, std::string_view token)
+   std::optional<T> decodeJWT(const JWTKey&                 key,
+                              std::string_view              token,
+                              std::function<bool(const T&)> validatePayload)
    {
-      return psio::convert_from_json<T>(decodeJWT(key, token));
+      auto payload = psio::convert_from_json<T>(decodeJWTPayloadUnsafe(token));
+      if (validatePayload(payload))
```
swatanabe: This is kind of backwards. I think it would make more sense to verify the MAC without aborting. If we ever decide to use encrypted tokens, decoding a token without the correct key isn't even possible.

libraries/psibase/tester/include/psibase/tester.hpp
```diff
@@ -429,17 +429,39 @@ namespace psibase
...
+             const T&                        data,
+             std::optional<std::string_view> token = std::nullopt)
       {
-         return http<R>(makePost(account, target, data));
+         auto req = makePost(account, target, data);
+         if (token)
+         {
+            req.headers.push_back(
+                {.name = "Authorization", .value = std::string("Bearer ") + std::string(*token)});
+         }
```
swatanabe: This logic should be in `makePost` (and `makeGet`).

packages/user/Tokens/query-service/src/lib.rs
```diff
@@ -87,6 +96,13 @@ mod service {
             before: Option<String>,
             after: Option<String>,
         ) -> async_graphql::Result<Connection<RawKey, SharedBalance>> {
+            if self.user != Some(user) {
+                return Err(async_graphql::Error::new(format!(
+                    "permission denied: '{}' must authorize your app to make this query. Send it through `tokens:plugin/authorized::graphql`.",
```
swatanabe: DRY

programs/psinode/tests/psinode.py
```diff
@@ -298,13 +305,16 @@ def set_producers(self, prods, algorithm=None):
...
+    def graphql(self, service, query, token=None):
         '''
         Sends a GraphQL query to a service and returns the result as json
 
         Raise GraphQLError if the query fails
         '''
-        with self.post('/graphql', service=service, json={'query': query}) as result:
+        kwargs = {'json': {'query': query}}
+        if token is not None:
+            kwargs['headers'] = {'Authorization': 'Bearer ' + token}
```
swatanabe: I think it would be better always to pass `headers`, but initialize it to an empty dict, instead of using `kwargs` to include or not include `headers`.

packages/user/Tokens/query-service/src/lib.rs
```diff
@@ -298,9 +335,24 @@ mod service {
...
     #[action]
     #[allow(non_snake_case)]
-    fn serveSys(request: HttpRequest) -> Option<HttpReply> {
+    fn serveSys(
+        request: HttpRequest,
+        _socket: Option<i32>,
+        user: Option<AccountNumber>,
+    ) -> Option<HttpReply> {
+        if get_sender() != AccountNumber::from("http-server") {
+            return Some(HttpReply {
```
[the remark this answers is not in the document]
James-Mart: Good point

programs/psinode/tests/psinode.py
```diff
@@ -151,7 +151,14 @@ def __init__(self, trace):
 
 class GraphQLError(Exception):
     def __init__(self, json):
-        super().__init__(json['errors']['message'])
+        errors = json.get('errors', [])
+        if errors and isinstance(errors, list):
+            message = errors[0].get('message', str(errors[0]))
+        elif isinstance(errors, dict):
+            message = errors.get('message', str(errors))
```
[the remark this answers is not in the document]
James-Mart: Oh, maybe? This is what worked and I didn't check the spec

libraries/psibase/common/include/psibase/jwt.hpp
James-Mart: > If we ever decide to use encrypted tokens, decoding a token without the correct key isn't even possible.

Good point.

So wait, your proposal here is that if validation fails, then don't abort, and don't return an http error, simply treat user = None. Right?

But if the validation succeeds, and then I decode the token and find that it's expired, then should that be an http error?

packages/user/Host/plugin/auth/src/lib.rs
```diff
@@ -9,64 +9,55 @@ use exports::host::auth::api::Guest as Api;
...
-    duration: StorageDuration::Persistent,
-};
-
 impl Api for HostAuth {
     fn set_logged_in_user(user: String, app: String) -> Result<(), Error> {
         check_caller(&["accounts"], "set-logged-in-user@host:auth/api");
 
-        let bucket = KvStore::Bucket::new(DB, &get_query_token_bucket_name(&user));
-        let query_token = bucket.get(&&app);
+        let query_token = Bucket::new(DB, &user)
```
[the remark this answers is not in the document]
James-Mart: I did consider this, but was thinking this plugin is really only meant to perform this one job. I suppose it's worth keeping just in case.

packages/user/Tokens/query-service/src/lib.rs
James-Mart: Fixed

libraries/psibase/common/include/psibase/jwt.hpp
swatanabe: > and don't return an http error

I could go either way on that.

PR #1579, opened 2025-11-09
add generate random account plugin function

Velua:
Quickly generate a random account number + check it doesn't exist on the network. 

This will be a dependency for Fractals where we want to create several role accounts with no concern of their actual names.

### DIFF
--- a/packages/system/Accounts/plugin/accounts/wit/world.wit
+++ b/packages/system/Accounts/plugin/accounts/wit/world.wit
@@ -58,6 +58,7 @@ interface api {
     get-current-user: func() -> option<string>;
 
     /// Generates a random account number not in use on the network
+    /// Supports optional prefix of up to 9 chars, must not start with "x-"
     gen-rand-account: func(prefix: option<string>) -> result<string, error>;
 }
 


packages/system/Accounts/plugin/accounts/src/helpers.rs
```diff
@@ -36,3 +36,44 @@ pub fn assert_valid_account(account: &str) {
...
+    loop {
+        if tries >= max_tries {
+            return Err(MaxGenerationAttemptsExceeded().into());
+        }
+        tries += 1;
+
+        let mut account = String::new();
+
+        // First character: lowercase a-z
+        let first_chars: &[char] = &(b'a'..=b'z').map(|b| b as char).collect::<Vec<_>>();
```
James-Mart: You shouldn't be building `first_chars` and `allowed_chars` repeatedly on each loop (in fact these could even be created statically)

packages/system/Accounts/plugin/accounts/src/helpers.rs
```diff
@@ -36,3 +36,44 @@ pub fn assert_valid_account(account: &str) {
...
+
+        for _ in 0..9 {
+            account.push(*allowed_chars.choose(&mut rng).unwrap());
+        }
+
+        let account_number = AccountNumber::from(account.as_str());
+        if account_number.to_string() == account {
+            let is_available = AccountsPlugin::get_account(account_number.to_string())?.is_none();
+            if is_available {
+                return Ok(account_number);
```
James-Mart: This function should just return a string because strings can cross plugin boundaries. Callers can convert to `AccountNumber` if needed.

packages/system/Accounts/plugin/accounts/src/helpers.rs
```diff
@@ -36,3 +36,44 @@ pub fn assert_valid_account(account: &str) {
...
+            .map(|b| b as char)
+            .collect::<Vec<_>>();
+
+        for _ in 0..9 {
+            account.push(*allowed_chars.choose(&mut rng).unwrap());
+        }
+
+        let account_number = AccountNumber::from(account.as_str());
+        if account_number.to_string() == account {
+            let is_available = AccountsPlugin::get_account(account_number.to_string())?.is_none();
```
James-Mart: `get_account` already does a roundtrip check (using `from_exact`), no need to roll your own.

packages/system/Accounts/plugin/accounts/src/helpers.rs
```diff
@@ -36,3 +36,44 @@ pub fn assert_valid_account(account: &str) {
...
+            if is_available {
+                return Ok(account_number);
+            }
+        }
+    }
```
James-Mart: The entire loop can be reduced to 
```
    for _ in 0..MAX_TRIES {
        let mut account = String::new();
        account.push(*first_chars.choose(&mut rng).unwrap());
        for _ in 0..9 {
            account.push(*allowed_chars.choose(&mut rng).unwrap());
        }

        if let Ok(None) = AccountsPlugin::get_account(account.clone()) {
            return Ok(account);
        }
    }
    Err(MaxGenerationAttemptsExceeded().into())
```

packages/system/Accounts/plugin/accounts/src/interfaces/api.rs
```diff
@@ -71,4 +72,9 @@ impl API for AccountsPlugin {
     fn get_current_user() -> Option<String> {
         AppsTable::new(&Client::get_active_app()).get_logged_in_user()
     }
+
+    fn gen_rand_account() -> Result<String, Error> {
```
James-Mart: One minor improvement I just thought of is that this could take a `prefix` parameter, and the generated account will start with the prefix. So for example all "representative roles" could be prefixed with "r-" and all council roles with "c-", etc. Just to make the accounts very slightly more consistent/understandable without using a dedicated front-end

PR #1582, opened 2025-11-12
Add Mapping table to Symbols

Velua:


packages/user/Symbol/service/src/lib.rs
```diff
@@ -114,21 +112,11 @@ pub mod tables {
         pub symbolId: AccountNumber,
         #[graphql(skip)]
         pub ownerNft: NID,
-        pub tokenId: Option<TID>,
```
James-Mart: Why split this out into a `Mapping` table?

James-Mart (approved):

PR #1588, opened 2025-11-14
Remake Token pending transfers query

cool-ant:
Enabling new userPending query to get all sharedBalances related to a user (regardless of whether they're the debitor or creditor). UI updated to use the new query.
During this work, the following issues were unearthed:
https://github.com/gofractally/psibase/issues/1606

### DIFF
--- a/packages/user/Tokens/service/src/tables.rs
+++ b/packages/user/Tokens/service/src/tables.rs
@@ -378,7 +378,7 @@ pub mod tables {
         }
 
         fn get(creditor: AccountNumber, debitor: AccountNumber, token_id: TID) -> Option<Self> {
-            let shared_creditor_bal_ids = UserPendingTable::new()
+            let shared_creditor_bal_ids = UserPendingTable::read()
                 .get_index_pk()
                 .range((creditor, token_id, 0_u64)..(creditor, token_id, u64::MAX))
                 .map(|bal| bal.shared_bal_id)
--- a/packages/user/Homepage/ui/src/apps/tokens/components/pending/button-accept.tsx
+++ b/packages/user/Homepage/ui/src/apps/tokens/components/pending/button-accept.tsx
@@ -1,4 +1,4 @@
-import type { PendingBalance } from "../../hooks/tokensPlugin/use-user-lines-of-credit";
+import type { PendingBalance } from "../../hooks/tokensPlugin/use-pending-balances";
 import type { PendingActionProps } from "../../pending";
 
 import { ArrowDown, Loader2 } from "lucide-react";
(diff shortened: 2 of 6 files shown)


packages/user/Tokens/query-service/src/lib.rs
```diff
@@ -75,46 +76,22 @@ mod service {
...
             before: Option<String>,
             after: Option<String>,
-        ) -> async_graphql::Result<Connection<RawKey, SharedBalance>> {
+        ) -> async_graphql::Result<Connection<RawKey, UserPendingRecord>> {
             self.check_user_auth(user)?;
 
-            TableQuery::subindex::<(AccountNumber, u32)>(
-                SharedBalanceTable::with_service(tokens::SERVICE).get_index_by_debitor(),
-                &(user),
+            // wrong table
```
James-Mart: It's the correct table

packages/user/Tokens/service/src/tables.rs
```diff
@@ -428,6 +473,12 @@ pub mod tables {
             let quantity = quantity.min(self.balance);
             self.sub_balance(quantity);
             Balance::get_or_new(self.creditor, self.token_id).add_balance(quantity);
+
+            if self.balance == 0.into() {
+                let upt = UserPendingTable::new();
+                upt.erase(&(self.creditor, self.token_id, self.shared_bal_id));
+                upt.erase(&(self.debitor, self.token_id, self.shared_bal_id));
+            }
```
James-Mart: This should be in `sub_balance`, and gated by the `keep_zero_balance` flag

packages/user/Tokens/service/src/tables.rs
```diff
@@ -459,10 +504,30 @@ pub mod tables {
...
 
         fn add_balance(&mut self, quantity: Quantity) {
+            if self.balance == 0.into() {
+                UserPendingRecord::add(
+                    self.creditor,
+                    self.debitor,
+                    self.token_id,
+                    self.shared_bal_id,
+                );
+                UserPendingRecord::add(
```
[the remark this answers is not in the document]
cool-ant: ok, clearly redundant, but wrong? how wrong? I don't see how/what?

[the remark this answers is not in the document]
cool-ant: ohhhh, ok, when I eliminated what I thought was identical (the wrong one), I couldn't find the wrong part. k, done. Thanks.

Velua (changes requested):
I don't see the point in this change, also at the expense of adding the additional state and complexity. 

The new `user_pending` query forces specifying the `token_id` whereas `user_credits` and `user_debits` would work across all tokens. This wouldn't be helpful if a UI wants to show the user all his credits and debits. 

If we want filtering by `token_id` we could just add an Option to their params and pass it to `TableQuery::subindex(.., &(token_id, user))`  to the pre-existing `user_credits` and `user_debits` with an additional secondary key.

```
#[secondary_key(2)]
fn by_token(&self) -> (TID, AccountNumber, AccountNumber) {
    (self.token_id, self.creditor, self.debitor)
}
```

Current API will mean a breaking change to the UI.

packages/user/Homepage/ui/src/apps/tokens/hooks/tokensPlugin/use-user-lines-of-credit.ts
brandonfancher: Do you want to rename this file too?

PR #1591, opened 2025-11-14
add helper to workshop plugin for finding compressible service method names

James-Mart:


packages/user/Workshop/plugin/src/lib.rs
```diff
@@ -138,4 +141,10 @@ impl Registry for WorkshopPlugin {
     }
 }
 
+impl Development for WorkshopPlugin {
+    fn is_valid_method_name(name: String) -> bool {
+        MethodNumber::from_str(&name).unwrap().to_string() == name
```
Velua: You don't wanna `unwrap_or(false)` here?

PR #1592, opened 2025-11-17
Getting system token

James-Mart:
I forgot to add an inline action for getting the system token


Velua:
https://github.com/gofractally/psibase/pull/1522 now depends on this

Velua (approved):

PR #1596, opened 2025-11-18
Credentials auth service fix

James-Mart:
If an auth service doesn't define `isAuthSys` and `isRejectSys`, then calling them will abort.

If the top-level auth service doesn't search authorizers in a smart way (partition and only make indirect calls if there are insufficient existing authorizers), then this could potentially be an exploit wherein a single potentially deeply-nested account could use an auth service that aborts in `isAuthSys`, effectively blocking staged transactions.

-- EDIT --

We are just going to abort anyways. The above scenario is still possible, but this is the wrong fix because the system needs to be able to handle an abort in cases like this regardless, since users have full control over their own auth services.

Two better ways to address the above scenario that can be done in some future fix:
1. Allow the front-ends to specify whether they want auto-executing staged-tx or not
2. Build better tooling for construction and submission of offline multi-sig

### DIFF
--- a/packages/system/Credentials/service/src/lib.rs
+++ b/packages/system/Credentials/service/src/lib.rs
@@ -173,7 +173,7 @@ pub mod service {
         _authorizers: Vec<AccountNumber>,
         _auth_set: Option<Vec<AccountNumber>>,
     ) -> bool {
-        false
+        abort_message("isAuthSys not supported");
     }
 
     #[action]
@@ -183,8 +183,9 @@ pub mod service {
         _authorizers: Vec<AccountNumber>,
         _auth_set: Option<Vec<AccountNumber>>,
     ) -> bool {
-        true
+        abort_message("isRejectSys not supported");
     }
+
 
     /// Creates a credential
     ///


swatanabe:
I don't understand what this is trying to solve. Anyone can create and use an auth service that aborts. It would take far more than just this to prevent it. How would this even come up? Is there any legitimate use for using the credential sender in a multisig?

swatanabe:
> > I don't understand what this is trying to solve.
> 
> Isn't it better for this defined behavior, rather than an abort?
> 

abort is defined behavior. Unless it creates a situation that can't be escaped from, an abort makes debugging easier.

> This also might prevent some esoteric ways of messing with the network

It doesn't prevent it at all.

James-Mart:
> > > I don't understand what this is trying to solve.
> > 
> > Isn't it better for this defined behavior, rather than an abort?
> 
> abort is defined behavior. Unless it creates a situation that can't be escaped from, an abort makes debugging easier.

Okay, I still think implementing the full interface is more explicit, and therefore better. 
I can change it to explicitly abort if that's preferred behavior for auth services that don't support staged transactions.

> > This also might prevent some esoteric ways of messing with the network
> 
> It doesn't prevent it at all.

1. With abort, "any auto-execute staged-tx proposals would be blocked".
2. Blocking a method of proposing staged transactions is messing with the network.
3. The plugin infra currently proposes auto-execute staged tx. 
=> This change prevents some ways of messing with the network.

With which premise do you disagree?

PR #1608, opened 2025-12-01
add a fixed point continuous fib function

James-Mart:
I wanted a thorough derivation of the algo in a doc comment to make it easier for others to understand how it works

screenshot (1)

The behavior of this function is shown in the following graphs
image

As you can see, the difference between the actual fib value and the value on the continuous curve is minimized over time. Here is a graph of the error up to input=11:
image

### DIFF
--- a/packages/user/Fractals/service/src/helpers/fib.rs
+++ b/packages/user/Fractals/service/src/helpers/fib.rs
@@ -212,7 +212,7 @@
 // Fixed-point scales
 const S: u128 = 1_000_000_000_000; // 12 decimal places internal scale
 const EXTERNAL_S: u128 = 10_000; // 4 decimal places
-const MAX_INPUT_SCALED: u32 = 37 * (EXTERNAL_S as u32); // Determined by manual testing
+const MAX_INPUT_SCALED: u32 = 32 * (EXTERNAL_S as u32); // Power of 2 (scaled) nearest to the max input for which the output is accurate.
 
 /// Precomputed scaled integer for ln(φ)
 /// ln(phi) * S ≈ 0.48121182505960347 * 1e12
@@ -353,7 +353,8 @@ mod tests {
 
     #[test]
     fn check_accuracy() {
-        let max_x_scaled: u32 = 37 * (EXTERNAL_S as u32);
+        let max_x_scaled: u32 = 32 * (EXTERNAL_S as u32);
+        // 1e-4 accuracy experimentally determined to break at input 37.381
 
         for x_scaled in 0..=max_x_scaled {
             let x = x_scaled as f64 / EXTERNAL_S as f64;


packages/user/Fractals/service/src/helpers/fib.rs
```diff
@@ -1,18 +1,335 @@
...
+/// 
+/// - the dominant sources of error are fixed-point rounding of constants and
+///   intermediate products, and  
+/// - the Taylor truncation itself is negligible relative to the \(10^{-4}\)
+///   absolute error tolerance for \(F(x)\) over the supported domain.
+
+// Fixed-point scales
+const S: u128 = 1_000_000_000_000; // 12 decimal places internal scale
+const EXTERNAL_S: u128 = 10_000; // 4 decimal places
+const MAX_INPUT_SCALED: u32 = 37 * (EXTERNAL_S as u32); // Determined by manual testing
```
swatanabe: Larger values fit in a 64-bit output. This limit seems to be a result of choosing S = 10<sup>12</sup>. log<sub>ϕ</sub>(S<sub>external</sub> / S) = 38.3

packages/user/Fractals/service/src/helpers/fib.rs
```diff
@@ -1,18 +1,266 @@
...
+/// approximation using fixed-point integer arithmetic.
+/// 
+/// \[
+/// F(x) \approx \frac{\varphi^x}{\sqrt{5}}
+/// \quad\text{with}\quad
+/// \varphi = \frac{1 + \sqrt{5}}{2}.
+/// \]
+/// 
+/// To make this deterministic and avoid floating point at runtime, we implement
+/// everything in fixed-point integer arithmetic.
```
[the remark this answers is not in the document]
swatanabe: The precise numbers are
- d/dt e<sup>t</sup>𝚫t <= ϕ𝚫t = 0.81/S (only considering the rounding error introduced by x * log ϕ).
- 0 (exact) + 0 (exact) + 0.5 (rounding) + 0.58 (0.5 + previous * t / k) + 0.57, ..., eventually converging to ~0.5 K/S. big-O isn't good notation here, because the constant factor matters. K/S is an upper bound on the error.
- The remainder of the taylor series is ~.33/S

packages/user/Fractals/service/src/helpers/fib.rs
swatanabe: That seems a little bit backwards. S is purely internal and doesn't directly show up in the API.  (how  did you choose S? I suspect that the maximum S that will work is ~10<sup>22</sup> since you're using u128). I'd rather decide the maximum value that we want to support first and then find a suitable value for S.

James-Mart: > I'd rather decide the maximum value that we want to support first and then find a suitable value for S.

This is basically what I did, the maximum value based on our current needs that we must support is `fib(12)`. `S=1e8` allows up to `fib(14)`.  But then if I decided if there was a simple improvement I could make to increase the available range (increasing `1e8` to `1e12`) then might as well go for it.

I stopped bumping at `1e12` because beyond that I was overflowing `u128` during some intermediate steps (`phi_integer_power_scaled` I think) and didn't feel like figuring out how to rework the math to push it further since we only actually needed `fib(12)` in the first place and `1e12` gets us to `fib(37)`.

James-Mart: Would you for some reason prefer that I drop it back to `1e8` so it's the minimum needed for our requirements?

packages/user/Fractals/service/src/helpers/fib.rs
```diff
@@ -1,18 +1,340 @@
...
+//! everything in fixed-point integer arithmetic.
+//!
+//! ---
+//!
+//! #### 1. Fixed-point representation
+//!
+//! Choose a global internal scale
+//!
+//! \\[
+//! S = 10^12.
```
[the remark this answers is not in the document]
James-Mart: Done

packages/user/Fractals/service/src/helpers/fib.rs
swatanabe: Aha, so S is chosen as the largest that won't cause overflow. Maybe make the limit 32, since that's a power of 2 and more than double the actual need.

PR #1612, opened 2025-12-03
Oxidise NFT

Velua:
Replaces C++ service version of `Nft` with Rust. 
One change is that burned NFT IDs cannot be reused as they were in the previous version.

### DIFF
--- a/packages/user/Nft/plugin/src/lib.rs
+++ b/packages/user/Nft/plugin/src/lib.rs
@@ -63,7 +63,7 @@ impl User for NftPlugin {
         trust::assert_authorized(trust::FunctionName::uncredit)?;
 
         let packed_args = Nft::action_structs::uncredit {
-            memo: memo.as_str().into(),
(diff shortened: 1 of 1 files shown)


Velua:
@James-Mart Can you tell the cause of the CI ? I can't sus out what's wrong with events.

James-Mart:
> @James-Mart Can you tell the cause of the CI ? I can't sus out what's wrong with events.

I think you've not fully removed the build configuration that tries to build the nft plugin separately
https://github.com/gofractally/psibase/pull/1612/files#diff-1e7de1ae2d059d21e1dd75d5812d5a34b0222cef273b7c3a2af62eb747f9d20aR293

CI Error:
> Error copying file (if different) from ".../**plugin_workspaces**/.../nft.wasm" to ...

Velua:
@James-Mart @swatanabe I'm stuck, I can't tell why this isn't building on CI, it builds and passes all tests locally.

James-Mart (approved):

PR #1617, opened 2025-12-08
Symbol creation support for administrators

Velua:
## Problem: Chicken & Egg. 
In order to start selling symbol lengths on symbol we need a billing token, however, we can't create a token + symbol* without a symbol length, which requires a billing token. 

*Technically, it's not critical to have a symbol for the billing / network token but we don't really want to use the network without it.*

## Solution: Short circuit for admins.
Introduce a shortcut to creating a Symbol that only those with root privileges can enjoy. 
This change means the Symbol service can create a `Symbol` table instance without any respective `SymbolLength` table instance to go along with it.


James-Mart:
> Thus, the [tokens] plugin function should not attempt to always call SA tokens::credit if there's nothing to actually credit.

Agreed, I think credits of 0 currently fail in the service anyway, plugin should probably also fail.

> ...I suppose one downside of this is that in the event the user expects something to be free during his UI experience on the Symbol purchase interface he might be suddenly thrown off by the [tokens permissions prompt]

Not really, the symbol plugin can just check if the quantity is 0 (in addition to checking whether there's a billing token), if so, don't call `tokens::credit` PF

>I'm guessing it shouldn't be up to the service to assume what the user wants, so tokens::reject from it's end is a bad smell to me

I can imagine a service explicitly using such a policy because they want to minimize having open credit lines with users, which seems fine to me. 

> PF symbol::create throw in a PF tokens::uncredit action to retrieve any dust

... this seems hardly less opinionated than the service calling reject. If we have this in the symbol::create pf, then we can't really change that behavior later without potentially breaking embedder apps.

I'm just saying - But like I said I'm fine with being opinionated here so injecting an `uncredit` OR calling service `reject` both work for me. 

Lastly, the following is unclear to me:

> This design avoids the introduction of special use SA + PF actions for admins to use

How so? Haven't read the code yet, but from this description I'm not seeing how these changes allow for the first symbol to be created outside of the symbol market?

Velua:
> > Thus, the [tokens] plugin function should not attempt to always call SA tokens::credit if there's nothing to actually credit.
> 
> Agreed, I think credits of 0 currently fail in the service anyway, plugin should probably also fail.

Should it fail? Or silently pass? 
Because the caller might not know if he's calling PF `tokens::credit` with a 0, unless there's additional code written on his side, which seems a shame to not have that abstracted, unless he receives the error in the result and conditionally ignores it... 

> 
> > ...I suppose one downside of this is that in the event the user expects something to be free during his UI experience on the Symbol purchase interface he might be suddenly thrown off by the [tokens permissions prompt]
> 
> Not really, the symbol plugin can just check if the quantity is 0 (in addition to checking whether there's a billing token), if so, don't call `tokens::credit` PF

True, forgot we can just conditionally call that. 

> 
> > I'm guessing it shouldn't be up to the service to assume what the user wants, so tokens::reject from it's end is a bad smell to me
> 
> I can imagine a service explicitly using such a policy because they want to minimize having open credit lines with users, which seems fine to me.
> 
> > PF symbol::create throw in a PF tokens::uncredit action to retrieve any dust
> 
> ... this seems hardly less opinionated than the service calling reject. If we have this in the symbol::create pf, then we can't really change that behavior later without potentially breaking embedder apps.
> 
> I'm just saying - But like I said I'm fine with being opinionated here so injecting an `uncredit` OR calling service `reject` both work for me.
> 
> Lastly, the following is unclear to me:
> 
> > This design avoids the introduction of special use SA + PF actions for admins to use
> 
> How so? Haven't read the code yet, but from this description I'm not seeing how these changes allow for the first symbol to be created outside of the symbol market?

There's a new `create` in the `symbol` interface of the `config` plugin, but it calls the same PF `symbol::create` which calls the same SA `symbol::create`. 

In the SA `symbol::create` all billing related items is wrapped inside of `if sender != crate::Wrapper::SERVICE` 

I'll proceed with adding a `tokens::reject` in the SA `symbol::create`.

packages/user/Symbol/service/src/lib.rs
```diff
@@ -325,7 +335,12 @@ pub mod service {
 
     #[action]
     fn create(symbol: AccountNumber) {
-        Symbol::add(symbol);
+        Symbol::add(symbol, get_sender());
+    }
+
+    #[action]
+    fn admin_create(symbol: AccountNumber, recipient: AccountNumber) {
```
James-Mart: I don't understand this... 
1. you don't need an `admin_create` action if the admin is detected as `sender == get_service()`.
2. You don't need a recipient parameter, the owner of tokens created by the admins is the symbol service.

packages/user/Tokens/plugin/src/lib.rs
```diff
@@ -173,17 +173,23 @@ impl Helpers for TokensPlugin {
...
+            .packed();
 
-        let packed_args = Actions::credit {
-            amount: Self::decimal_to_u64(token_id, amount)?.into(),
-            memo: memo.try_into().unwrap(),
-            debitor: debitor.as_str().into(),
-            token_id,
+            add_action_to_transaction(Actions::credit::ACTION_NAME, &packed_args)
+        } else {
+            Ok(())
```
James-Mart: I don't think this is an improvement. The caller *should* be forced to acknowledge that it's not doing anything, a silent no-op.

Proposed alternatives that let the caller decide if a no-op is the right behavior:
* Leave as is, let the caller check if the amount > 0
* Return an error, let the caller handle the error

packages/user/Tokens/service/src/lib.rs
```diff
@@ -461,7 +461,7 @@ pub mod service {
     /// * `memo`     - Memo
     #[action]
     fn reject(token_id: TID, creditor: AccountNumber, memo: Memo) {
-        SharedBalance::get_assert(creditor, get_sender(), token_id).reject(memo);
+        SharedBalance::get_or_new(creditor, get_sender(), token_id).reject(memo);
```
James-Mart: This changes `reject` from a fail if invalid to a no-op.

1) If we are going to use no-op, this should just conditionally call reject if get(...).is_some(), instead of creating an empty SharedBalance and immediately rejecting it.
2) Should it be a no-op? `credit` and `debit` fail if the amount is 0. Reject seems to me like the most likely action where a no-op is justified... but it does make it different than the other actions.

One other design to consider is a separate `reject_any` action that no-ops, and leave the default `reject` as an assert. Then the caller can express their intended usage based on which action is called.

packages/user/Tokens/plugin/src/lib.rs
Velua: To use explicit error code

packages/user/Tokens/service/src/lib.rs
Velua: 1. As disccused, we'll ignore that feedback and leave as is. 
2. Concluded in discussion, we're happy to no-op.

packages/user/Symbol/service/src/lib.rs
Velua: Rename to `create_for` and make it callable by the general public.

packages/user/Symbol/service/src/lib.rs
```diff
@@ -325,7 +335,13 @@ pub mod service {
...
+        check(get_sender() == get_service(), "only service can call this action");
```
James-Mart: "Only service can call this action" 

Did you mean, "symbol service"?

James-Mart:
```
/home/runner/work/psibase/psibase/packages/user/Symbol/test/src/Symbol-test.cpp:326: FAILED:
  CHECK( !err )
with expansion:
  false
with message:
  Missing action admin_create
```

PR #1624, opened 2025-12-09
Add Fractal core exile UI

Velua:
Introduces the Judiciary and Legislative pages which touches on the role of each branch and which guild is occupying the position. 
Beneath it, we conditionally show a control panel component to any administrative users which is either a representative of the guild or a council member. 
The control panel will show forms related to the actions the branch is allowed to execute.  

First control panel item is added for the Judiciary which lets them exile a Fractal member.

### DIFF
--- a/packages/user/FractalCore/ui/src/components/modals/exile-fractal-member-modal.tsx
+++ b/packages/user/FractalCore/ui/src/components/modals/exile-fractal-member-modal.tsx
@@ -6,8 +6,13 @@ import {
     DialogHeader,
     DialogTitle,
 } from "@shared/shadcn/ui/dialog";
-import { useAppForm } from "../form/app-form";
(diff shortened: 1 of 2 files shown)


packages/user/FractalCore/ui/src/components/modals/exile-fractal-member-modal.tsx
```diff
@@ -0,0 +1,71 @@
...
+                        onSubmit={(e) => {
+                            e.preventDefault();
+                            void form.handleSubmit();
+                        }}
+                        className="mt-3 w-full space-y-6"
+                    >
+                        <form.AppField
+                            name="member"
+                            children={(field) => (
+                                <field.TextField label="Member" />
```
brandonfancher: This may be a good candidate for the shared form field in `field-account-existing.tsx`

PR #1633, opened 2025-12-11
Add tests that the C++ declarations of rust packages match

swatanabe:
This also updates the C++ schema comparison to match the rust implementation:
- u8 and i8 are considered equivalent (needed because of C++ char)
- A struct containing a single scalar is equivalent to just the scalar


libraries/psibase/tester/src/checkSchema.cpp
```diff
@@ -0,0 +1,157 @@
...
+         {
+            const auto& rty = rpos->second;
+            if (!matcher.match(lty.params, rty.params))
+               return std::format("Parameter types for action {} do not match", name);
+            if (lty.result && rty.result)
+            {
+               if (!matcher.match(*lty.result, *rty.result))
+                  return std::format("Return type for {} does not match", name);
+            }
+            else if (!lty.result && rty.result)
```
James-Mart: Why is this an empty body?
Also this condition `(!lty.result && rty.result)` is repeated

PR #1640, opened 2025-12-15
Conditional credit on token stream

Velua:
There's awkward behaviour with the `claim` action where it assumes that the `claimed_amount` is above zero which will throw when `Tokens::credit` is called. 

This wraps the credit + event actions in `if claimed_amount.value > 0` 

This way, services can safely call `claim` and conditionally call debit themselves.

### DIFF
--- a/packages/user/TokenStream/service/src/lib.rs
+++ b/packages/user/TokenStream/service/src/lib.rs
@@ -232,7 +232,7 @@ pub mod service {
         );
 
         stream.deposit(amount);
-        
+
         Wrapper::emit().history().updated(
             nft_id,
             sender,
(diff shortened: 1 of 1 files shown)


Velua:
In the claim method of the stream we have...

```
        pub fn claim(&mut self) -> Quantity {
            let claimable = self
                .total_vested()
                .value
                .saturating_sub(self.total_claimed.value);
            self.total_claimed = self.total_claimed.value.saturating_add(claimable).into();
            self.save();
            claimable.into()
        }
```

Is claimable is `0` then the `self.total_claimed` will do nothing and save will be called, is that okay? Or should we spare the redundant save and do an early return?
Like...
```
        pub fn claim(&mut self) -> Quantity {
            let claimable = self
                .total_vested()
                .value
                .saturating_sub(self.total_claimed.value);
            if claimable == 0 {
                return claimable.into();
            }
            self.total_claimed = self.total_claimed.value.saturating_add(claimable).into();
            self.save();
            claimable.into()
        }
```

PR #1641, opened 2025-12-15
Add exile, set, remove and resign guild rep in Fractals UI

Velua:
Depends on https://github.com/gofractally/psibase/pull/1637

### DIFF
--- a/packages/user/FractalCore/ui/src/components/guild-overview-card.tsx
+++ b/packages/user/FractalCore/ui/src/components/guild-overview-card.tsx
@@ -1,28 +1,36 @@
+import { Crown, User, Users } from "lucide-react";
+import z from "zod";
+
 import { useGuild } from "@/hooks/use-guild";
+import { Guild } from "@/lib/graphql/fractals/getGuild";
+
 import { useChainId } from "@shared/hooks/use-chain-id";
+import { createIdenticon } from "@shared/lib/create-identicon";
 import {
     Card,
     CardContent,
     CardHeader,
     CardTitle,
 } from "@shared/shadcn/ui/card";
-import { createIdenticon } from "@shared/lib/create-identicon";
 import { Skeleton } from "@shared/shadcn/ui/skeleton";
-import { User, Users, Crown } from "lucide-react";
-import z from "zod";
-import { Guild } from "@/lib/graphql/fractals/getGuild";
 
-const zLeadership = z.enum(['RepAndCouncil', 'RepOnly', 'CouncilOnly']);
+const zLeadership = z.enum(["RepAndCouncil", "RepOnly", "CouncilOnly"]);
 
-const getLeadershipStatus = (guild?: Guild | null): z.infer<typeof zLeadership> | undefined => {
+const getLeadershipStatus = (
+    guild?: Guild | null,
+): z.infer<typeof zLeadership> | undefined => {
     if (guild == undefined || guild == null) return;
     if (guild.rep && guild.council) return zLeadership.Enum.RepAndCouncil;
     if (guild.rep) return zLeadership.Enum.RepOnly;
     if (guild.council) return zLeadership.Enum.CouncilOnly;
     throw new Error("Cannot determine leadership status");
-}
+};
 
-export const GuildOverviewCard = ({ guildAccount }: { guildAccount?: string }) => {
+export const GuildOverviewCard = ({
+    guildAccount,
+}: {
+    guildAccount?: string;
+}) => {
     const { data: guild } = useGuild(guildAccount);
     const { data: chainId } = useChainId();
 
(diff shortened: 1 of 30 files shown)


packages/user/FractalCore/ui/src/components/modals/set-guild-rep-modal.tsx
```diff
@@ -0,0 +1,82 @@
...
+    const form = useAppForm({
+        defaultValues: {
+            representation: {
+                newRep: ""
+            },
+        },
+        onSubmit: async ({ value: { representation: { newRep } } }) => {
+            await setGuildRep([guildAccount!, newRep]);
+            openChange(false);
+        },
```
brandonfancher: I wonder if we could reset the form in this `onSubmit` instead of the `useEffect` on `L44` below. If so, that could be applied to all of these modals. Would something like this work?

```
onSubmit: async ({ value: { representation: { newRep } }, formApi }) => {
    await setGuildRep([guildAccount!, newRep]);
    openChange(false);
    formApi.reset();
},
```

packages/user/FractalCore/ui/src/components/modals/set-min-scorers-modal.tsx
```diff
@@ -0,0 +1,76 @@
...
+}: {
+    show: boolean;
+    openChange: (show: boolean) => void;
+}) => {
+    const { mutateAsync: setMinScorers } = useSetMinScorers();
+
+    const { data: fractal } = useFractal();
+    const form = useAppForm({
+        defaultValues: {
+            minScorers: fractal?.fractal?.minimumRequiredScorers || 4,
```
brandonfancher: I assume `minimumRequiredScorers` can never be `0`, but if that were a possible value, it would be overridden with `4` here. The `??` operator would be more precise here.

packages/user/FractalCore/ui/src/components/guild-overview-card.tsx
```diff
@@ -0,0 +1,74 @@
+import { useGuild } from "@/hooks/use-guild";
+import { useChainId } from "@shared/hooks/use-chain-id";
+import {
+    Card,
+    CardContent,
+    CardHeader,
+    CardTitle,
+} from "@shared/shadcn/ui/card";
+import { createIdenticon } from "@/lib/createIdenticon";
```
brandonfancher: We already have `createIdenticon` in `shared-ui`

packages/user/FractalCore/ui/src/components/guild-overview-card.tsx
```diff
@@ -0,0 +1,74 @@
...
+                        <div>
+                            <h3 className="text-muted-foreground mb-1 text-sm font-medium">
+                                Mission
+                            </h3>
+                            <p className="text-sm leading-relaxed">
+                                {guild?.description || "No mission available."}
+                            </p>
+                        </div>
+                        <div className="flex items-center text-muted-foreground">
+                            Lead by {guild?.rep ? `representative ${guild.rep.member}` : 'council'}
```
brandonfancher: "Led" ?

packages/user/FractalCore/ui/src/pages/fractal/branches/judicial/index.tsx
```diff
@@ -0,0 +1,77 @@
...
+    Card,
+    CardContent,
+    CardDescription,
+    CardHeader,
+    CardTitle,
+} from "@shared/shadcn/ui/card";
+import { Gavel } from "lucide-react";
+
+import { GuildOverviewCard } from "@/components/guild-overview-card";
+import { ErrorCard } from "@/components/error-card";
```
brandonfancher: We have a shared error-card now

brandonfancher (changes requested):

packages/user/FractalCore/ui/src/components/modals/set-min-scorers-modal.tsx
```diff
@@ -0,0 +1,76 @@
...
+    Dialog,
+    DialogContent,
+    DialogHeader,
+    DialogTitle,
+} from "@shared/shadcn/ui/dialog";
+import { useAppForm } from "@shared/components/form/app-form";
+import { useSetMinScorers } from "@/hooks/fractals/use-set-min-scorers";
+import { zU8 } from '@shared/lib/schemas/u8'
+import { MIN_MINIMUM_REQUIRED_SCORERS } from "@/lib/constants";
+import { useFractal } from "@/hooks/fractals/use-fractal";
```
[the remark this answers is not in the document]
Velua: True, lets ignore prettier for now and after approval but before merge we'll hit it with a prettier commit of something like `yarn exec prettier ./{user|system}/{uiDir}/ui --write` as it appears to change more than what's being updated in this PR so will get confusing.

packages/user/FractalCore/ui/src/hooks/fractals/use-dist-token.ts
```diff
@@ -0,0 +1,24 @@
...
+import { useFractalAccount } from "./use-fractal-account";
+
+export const useDistToken = () => {
+    const fractalAccount = useFractalAccount();
+    return usePluginMutation(fractalCorePlugin.userFractal.distToken, {
+        error: "Failed token distribution",
+        loading: "Distributing",
+        success: "Distributed tokens",
+        isStagable: true,
+        onSuccess: (_, status) => {
```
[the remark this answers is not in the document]
Velua: To replace with `usePluginFunctionCallMutation`

Issue #186, opened 2022-09-27
Integrate sub-account system with token-sys


James-Mart:
## Background 

Anyone building an escrow system, DEX, or custodial service will benefit from the ability to maintain sub-accounts for each of their users, rather than having all custodied tokens live in the same balance. We know that the services need to track how much each of their depositors have available at any given time, so that accounting logic could get duplicated in each instantiation of a custodial service.

## Desired solution properties

A solution is desired that:
1. Prevents the accounting logic from being duplicated in each separate custodial service
2. Keeps various user balances separated into different accounts, in order to help maintain the invariant that the sum of all of a user's withdraws must be <= the sum of all deposits (A bug that allows a user to withdraw more than expected will still only be able to withdraw, at max, the user's own balance).

## Solution 1: Bank service

One way to address this concept would be to create something like a "Bank" service that gives users the ability to create many sub-accounts, deposit/withdraw from each of them, and automatically handles the accounting in one place. 

The issue with such a solution is that it's expensive. Imagine a DEX user who deposits funds to the DEX. The flow of actions & events in a single transaction would be:
* User credits DEX (user@token-sys:credit)
* User notifies DEX of credit (user@dex:deposit)
* Dex debits the deposit (dex@token-sys:debit)
    * History Event [Transferred]
* Dex credits the bank (dex@token-sys:credit)
* Dex notifies the bank (dex@bank:deposit)
* Bank debits the deposit (bank@token-sys:debit)
    * History event [Transferred]

Therefore, with this design, all transfers into or out of custodial services would each result in at least two transferred events, doubling the resource cost.

## Solution 2: Sub-account integration with token-sys

Alternatively, we could integrate a sub-account system with the token-sys contract itself. This would simplify the construction of custodial services, and minimize the cost of maintaining a queryable event history for this type of interaction. Under this design, a user performing the same DEX deposit would have an action/event flow that would look like the following:

* User credits the DEX (user@token-sys:credit)
* User notifies DEX of credit (user@dex:deposit)
* Dex debits the deposit into subaccount (dex@token-sys:debitsub)
    * Transferred event

PR #1650, opened 2025-12-17
Tokens subaccounts

James-Mart:
Closes https://github.com/gofractally/psibase/issues/186

This adds support for subaccounts in the tokens service. This is a feature that has been discussed for a long time, and it's finally being added as it's used in objective billing (and in several other features in active development).

An account "alice" can now create subaccounts within the tokens app, e.g. "checking", "savings". Tokens can be moved from alice's primary balance to the balance of a subaccount. Tokens cannot be sent directly from a subaccount and must instead first be transferred back to the main account balance. 

A subaccount can be deleted, which will return all tokens it contains back to the primary account balance.

Another use-case is for services that custody tokens on behalf of users, a sub-account could be created per user. 
Partitioning balances can also be helpful as a hedge against service bugs that would allow for the primary balance to get unexpectedly spent, it is not necessarily the case that subaccounts would be affected.

--
Note that this PR only includes the service implementation, there is no client-side support yet in the tokens plugin for the creation/management of subaccounts. As is, it can only easily be used by other services.

### DIFF
--- a/packages/user/Tokens/service/src/tables.rs
+++ b/packages/user/Tokens/service/src/tables.rs
@@ -844,12 +844,13 @@ pub mod tables {
                 .range((self.id, 0)..(self.id, u32::MAX))
                 .collect();
 
-            for mut sub_balance in sub_balances {
+            let sab_table = SubAccountBalanceTable::read_write();
+            for sub_balance in sub_balances {
                 if sub_balance.balance.value > 0 {
                     Balance::get_or_new(self.account, sub_balance.token_id)
                         .add_balance(sub_balance.balance);
                 }
-                sub_balance.delete();
+                sab_table.erase(&sub_balance.pk());
             }
 
             SubAccountTable::read_write().erase(&(&self.pk()));


packages/user/Tokens/query-service/src/lib.rs
```diff
@@ -44,6 +45,22 @@ mod service {
...
+            sub_account: &str,
+        ) -> async_graphql::Result<SubAccount> {
+            SubAccountTable::with_service(tokens::SERVICE)
+                .get_index_pk()
+                .get(&(user, sub_account.to_string()))
+                .ok_or_else(|| {
+                    async_graphql::Error::new(format!(
+                        "Sub-account '{}' not found for user '{}'",
+                        sub_account, user
+                    ))
```
Velua: I kinda prefer just a straight `Option`. 
Otherwise this can tank the rest of the Graphql query which is annoying.

packages/user/Tokens/service/src/tables.rs
```diff
@@ -727,4 +745,185 @@ pub mod tables {
...
+                .get_index_pk()
+                .range((self.id, 0)..(self.id, u32::MAX))
+                .collect();
+
+            for mut sub_balance in sub_balances {
+                if sub_balance.balance.value > 0 {
+                    Balance::get_or_new(self.account, sub_balance.token_id)
+                        .add_balance(sub_balance.balance);
+                }
+                sub_balance.delete();
```
Velua: This calls `table::read_write` for every sub_balance rather than just the once if we initialised it outside of the loop. Are we concerned about this or not a big deal?

James-Mart: It should be moved outside the loop

James-Mart: I fixed this particular instance but there's actually a ton of these kinds of inefficiencies. These interfaces would be better in general if we were passing handles around instead of fetching new handles all the time. We can clean these up later when we are optimizing for more efficient cpu usage

PR #1661, opened 2026-01-04
Add token swap

Velua:
Add token swap service, plugin and UI on the home page app to interface with it. 
Plugin functions support adding and removing liquidity, finding paths between pools via the Bellman ford algo. 
Pools have administration NFTs which can be used to set fees on either side of sale of a pool reserve token.


packages/user/TokenSwap/plugin/src/errors.rs
```diff
@@ -0,0 +1,14 @@
+use psibase::plugin_error;
+
+plugin_error! {
+    pub ErrorType
+    TokenDoesNotExist => "Token does not exist in pool",
+    QueryResponseParseError(msg: String) => "Query response parsing error: {msg}",
+    QueryError(msg: String) => "Query error: {msg}",
+    InvalidFormat(msg: String) => "Invalid format: {msg}",
+    SlippageTooHigh(slippage: u32) => "Slippage must be in ppm format: received {slippage}, must be between 0 - 1,000,000",
+    NoPathFound => "Failed to find pools to fulfil trade of these tokens",
```
James-Mart: `NoPathFound` is unused

packages/user/TokenSwap/service/src/lib.rs
```diff
@@ -0,0 +1,583 @@
...
+                Self::get(liquidity_token),
+                format!("pool does not exist, liquidity id: {}", liquidity_token).as_str(),
+            )
+        }
+
+        fn pool_token(&self) -> TokenRecord {
+            Tokens::call().getToken(self.liquidity_token)
+        }
+
+        pub fn pool_token_supply(&self) -> Quantity {
```
James-Mart: Can be private

packages/user/TokenSwap/service/src/lib.rs
```diff
@@ -0,0 +1,583 @@
...
+                .getByToken(self.token_id)
+                .map(|s| s.symbolId)
+        }
+
+        pub async fn pool(&self) -> Pool {
+            Pool::get_assert(self.pool_id)
+        }
+    }
+
+    impl Pool {
```
James-Mart: The following functions can all be made private:
* `pool_token_supply`
* `debit_lp_tokens_from_sender`
* `mint_pool_tokens`
* `credit_sender_pool_tokens`
* `burn_lp_tokens`

packages/user/TokenSwap/service/src/lib.rs
```diff
@@ -0,0 +1,583 @@
...
+    }
+
+    #[table(name = "ReserveTable", index = 1)]
+    #[derive(Serialize, Deserialize, SimpleObject, ToSchema, Fracpack, Debug)]
+    #[graphql(complex)]
+    pub struct Reserve {
+        pub pool_id: TID,
+        pub token_id: TID,
+        pub fee_ppm: u32,
+        pub admin_nft: NID,
```
James-Mart: What if:

The `admin_nft` lived in the Pool instead of the reserve. Creator of the pool gets the NFT. The wrapper service that wants to forward authority over each reserve fee to independent parties can itself issue and distribute NFTs, and those admins will set the fees by calling actions through the wrapper service.

This also simplifies the implementation we've previously discussed where:
* `new_pool` uses the specified NFT id (if provided) or issues one if not. Now the wrapper service can have a single NFT that ensures they can manage any of their created pools.

packages/user/TokenSwap/service/src/lib.rs
```diff
@@ -0,0 +1,583 @@
...
+            a_reserve.admin_nft,
+            b_reserve.admin_nft,
+        )
+    }
+
+    /// Updates the swap fee for one of the tokens in an existing pool.
+    ///
+    /// Only the current owner of the administration NFT for the specified token's reserve
+    /// is allowed to change the fee. The fee is set in parts per million (ppm).
+    /// Maximum allowed value is typically 999_999 (≈99.9999%).
```
James-Mart: ~~typically~~

packages/user/TokenSwap/query-service/src/lib.rs
James-Mart: A couple extra queries would be nice:
* **[authenticated]** - In what pools is the specified user providing liquidity?
  * Should return the list of LP token IDs and the number of those tokens that user holds
* **[public]** - How many of each reserve token would be received given X if selling X LP tokens?

packages/user/TokenSwap/plugin/wit/world.wit
```diff
@@ -0,0 +1,156 @@
...
+
+}
+
+interface queries {
+
+    use host:types/types.{error};
+    use types.{path, tid, pool, token-amount};
+
+    /// Fetches all pools in the network
+    fetch-pools: func() -> result<list<pool>, error>;
```
James-Mart: Since this isn't an authenticated query, couldn't UIs just query for this directly on the query service?

Velua: Yeah was just a throw in.

packages/user/TokenSwap/plugin/wit/world.wit
```diff
@@ -0,0 +1,156 @@
...
+    /// after fees/tariffs, within the given slippage tolerance and maximum hop count.
+    ///
+    /// Arguments:
+    ///   - `pools`: `some(list<pool>)` → use these pools only (offline / cached mode)
+    ///             `none` → automatically fetch current pools via GraphQL / chain query
+    ///   - `from-amount`: token and decimal amount the user wants to sell/spend
+    ///   - `to-token`: desired output token ID
+    ///   - `slippage`: maximum acceptable slippage in parts per million (1_000_000 = 100%)
+    ///                 Example: 5000 = 0.5% slippage tolerance
+    ///   - `max-hops`: maximum number of pools/swaps in the path (usually 2–6)
```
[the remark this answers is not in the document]
Velua: User doesn't, dev might. It's hardcoded to 3 on the UI use.
Some use cases might involve prioritising trade routes that require few hops.

packages/user/TokenSwap/plugin/wit/world.wit
```diff
@@ -0,0 +1,156 @@
...
+    ///
+    /// Returns:
+    /// * Best found `path` record (including recommended `minimum-return`)
+    quote: func(pools: option<list<pool>>, from-amount: token-amount, to-token: tid, slippage: u32, max-hops: u8) -> result<path, error>;
```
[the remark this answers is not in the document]
James-Mart: I was thinking all of the specified pools would need to be fetched anyway, or do you just want to trust the user?

In fact, you're right - no point in even providing the pool IDs. I think the real consideration is whether it makes sense to leave `pools` out entirely and the plugin queries for pool details to derive the path.

packages/user/TokenSwap/plugin/wit/world.wit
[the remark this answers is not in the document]
James-Mart: But devs who want a specific trade route can just forcibly use a particular trade route, slippage be damned, right?

packages/user/TokenSwap/plugin/src/find_path.rs
```diff
@@ -0,0 +1,235 @@
...
+        // Try every pool connected to current token
+        if let Some(connected_pools) = graph.get(&search_state.token) {
+            for pool in connected_pools {
+                // Determine next token + correct fee/reserves direction
+                let (input_token, output_token) = if pool.token_a == search_state.token {
+                    (pool.token_a, pool.token_b)
+                } else if pool.token_b == search_state.token {
+                    (pool.token_b, pool.token_a)
+                } else {
+                    continue;
```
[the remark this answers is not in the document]
Velua: No it shouldn't, I'll abort here instead.

packages/user/TokenSwap/plugin/wit/world.wit
```diff
@@ -0,0 +1,148 @@
...
+    /// This is the function that actually performs the swap on-chain.
+    ///
+    /// Arguments:
+    ///   - `pools`: Ordered list of pool IDs (as strings) – must match `path.pools`
+    ///   - `amount-in`: The exact input amount and token the user is sending
+    ///   - `min-return`: The minimum acceptable output amount (usually from `path.minimum-return`)
+    swap: func(pools: list<string>, amount-in: token-amount, min-return: string) -> result<_, error>;
+
+
+    /// Get the best swap path from one token to another, respecting slippage and hop limit.
```
swatanabe: What if the optimal strategy is not a linear path? If there are multiple independent paths with a similar result, splitting the tokens between those paths will give better results overall.

packages/user/TokenSwap/plugin/wit/world.wit
```diff
@@ -0,0 +1,148 @@
...
+    /// It performs a graph search to find the path that gives the highest output amount
+    /// after fees/tariffs, within the given slippage tolerance and maximum hop count.
+    ///
+    /// Arguments:
+    ///   - `pools`: `some(list<pool>)` → use these pools only (offline / cached mode)
+    ///             `none` → automatically fetch current pools via GraphQL / chain query
+    ///   - `from-amount`: token and decimal amount the user wants to sell/spend
+    ///   - `to-token`: desired output token ID
+    ///   - `slippage`: maximum acceptable slippage in parts per million (1_000_000 = 100%)
+    ///                 Example: 5000 = 0.5% slippage tolerance
```
swatanabe: Is the slippage tolerance meaningful? I suppose it could make sense to help the user decide whether he wants to do the trade at all, but using to filter the viable paths can only result in the user getting a worse deal.

packages/shared-ui/lib/graphql.ts
```diff
@@ -8,23 +10,24 @@ interface Error {
...
 interface GraphqlResponse<T> {
     data: T;
     errors?: Error[];
 }
 
 export const graphql = async <T>(
     query: string,
-    service?: string,
+    service: Account,
+    includesSibling = true,
```
brandonfancher: This will conflict with the changes I made in #1691, aimed at solving the same problem.

brandonfancher (changes requested):

brandonfancher (changes requested):
I'm getting this error when trying to swap:
Image

brandonfancher (approved):

packages/user/TokenSwap/service/src/helpers.rs
```diff
@@ -0,0 +1,25 @@
...
+pub fn sqrt(n: u128) -> u128 {
+    if n == u128::MAX {
+        return u64::MAX as u128;
+    }
+    if n < 2 {
+        return n;
+    }
+
+    let mut x = n >> (n.ilog2() / 2);
+    let mut y = (x + 1) / 2;
```
swatanabe: The simplification of n/x = 1 is not valid anymore. In particular, this gives the wrong result for n = 4.

packages/user/TokenSwap/service/src/lib.rs
```diff
@@ -0,0 +1,631 @@
...
+                Tokens::call().getSubBal(self.token_id, self.pool_sub_account_id()),
+                "reserve balance not found",
+            )
+        }
+
+        pub fn debit_from_sender(&self, amount: Quantity) {
+            let sender = get_sender();
+            let tokens = Tokens::call();
+            tokens.debit(self.token_id, sender, amount, "memo".into());
+            tokens.reject(self.token_id, sender, "memo".into());
```
James-Mart: * Empty memo for debit
* "Change" for the reject?

packages/user/TokenSwap/service/src/lib.rs
```diff
@@ -0,0 +1,631 @@
...
+            Tokens::call().debit(
+                self.liquidity_token,
+                get_sender(),
+                liquidity_amount,
+                "Liquidity withdrawal".into(),
+            );
+        }
+
+        fn mint_pool_tokens(&self, amount: Quantity) {
+            Tokens::call().mint(self.liquidity_token, amount, "Pool tokens".into());
```
James-Mart: Empty memo

packages/user/TokenSwap/plugin/src/lib.rs
```diff
@@ -0,0 +1,341 @@
...
+        Low => "",
+        Medium => "",
+        High => "
+            - Performing token swaps
+            - Creating new liquidity pools
+            - Adding and removing liquidity to / from pools
+        ",
+    }
+    functions {
+        None => [quote, quote_pool_tokens, quote_add_liquidity, quote_remove_liquidity, fetch_pools],
```
James-Mart: Also `fetch_pool` no longer exists

packages/user/TokenSwap/service/src/lib.rs
```diff
@@ -0,0 +1,631 @@
...
+
+            let (reserve_a, reserve_b) = self.reserves_in_order(token_a_id, token_b_id);
+
+            let reserve_a_balance = reserve_a.balance();
+            let reserve_b_balance = reserve_b.balance();
+
+            check(
+                reserve_a_balance.value > 0 && reserve_b_balance.value > 0,
+                "pool does not have sufficient reserves",
+            );
```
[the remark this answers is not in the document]
Velua: No, given we burn some of the initial liquidity tokens.

packages/user/TokenSwap/service/src/lib.rs
Velua: I've been using "Dust"

Velua: Well we wanna reject any left over tokens not used when adding liquidity, however reject doesn't let us specify an amount so I suppose we will return any and all.

James-Mart:
From https://github.com/gofractally/psibase/pull/1661#discussion_r2823860452

> After all swaps, I think an event should be emitted:
> 
> `swap_completed(get_sender(), token_in, amount_in, token_out, amount_out)`
> 
> Indexed by column `0` 

Unless I'm missing it, I think you added the event, but never emit it.

James-Mart (approved):

PR #1676, opened 2026-01-16
Objective billing (CPU & NET only)

James-Mart:
# Objective billing

This PR introduces an initial system that can be used to measure and rate-limit the consumption of system resources on a network with multiple peers. "Objective" in this case means that the resources that are consumed must be consumed across all full infrastructure provider nodes on the network, as opposed to node-local resource consumption (such as those consumed by a read-query to a particular node). 

## Exclusions

This PR does not support rate-limiting Disk consumption for application state or temporary event log storage. This PR also does not address any rate-limiting for node-local resource consumption (e.g. query billing).

## The philosophy

A server that is part of the network of infra providers has some fixed amount of capacity available for servicing user transactions, as determined by the server's physical hardware constraints such as Disk capacity and network bandwidth capacity. The actual resource constraints must necessarily be less than or equal to the physical hardware constraints. The subset of the physical resources that is expected to be present on each infra provider node make up the specs of a "virtual server". This virtual server is the same regardless of which physical hardware it's running on, because all infra providers are expected to meet or exceed the virtual server specs.

After configuring the virtual server specs, we calculate a ratio of those actual resources that we can reliably consume for servicing user transactions. For example, total actual network bandwidth capacity must be divided by the number of full nodes involved in consensus, because each byte of bandwidth consumed in a transaction is multiplied by that number of nodes as a side-effect of the message propagation in our consensus algorithm. Or, similarly, if we desire for block replay (which is used when new infrastructure nodes peer with existing nodes) to proceed at some rate `R`-times faster than normal block production, then the theoretical *maximum* of total CPU time available per second to spend on user transactions is 1/R of one second.

This subset of virtual-server resources available for processing user-transactions define the *network specs*. It is the consumption of the resources of the network that are monitored, and against which users are rate-limited. 

### Rate limiting

To rate limit users network bandwidth consumption and consumed nanoseconds of CPU time, we use an algorithm inspired by the Bitcoin difficulty adjustment. Users must pay an amount of resource tokens equal to the amount of consumed resources multiplied by the spot price of the consumed resource at the time of the transaction. There is a target "rate" of consumption (like the block-time rate in Bitcoin) above which the resource is considered to be in "congestion" and below which the resource is considered to be "idle". When in congestion, the price (akin to "difficulty" in bitcoin) doubles at a configurable rate. When idle, the price of the resource halves at a configurable rate.

### Resource token

The resource token is exchangeable 1:1 for the system token at any time, but the resource token does not guarantee any particular amount of resource. Instead, a market for each individual resource (CPU time, network bandwidth, etc) vs. the resource token are used to purchase/consume some amount of actual resources at the time of the transaction. Resource tokens are used for internal accounting related to rate-limiting of physical resources, and typically should not need to be manually managed by users. Accounts typically "fill" a buffer with resource tokens so as to not require constant exchanges of system tokens for resource tokens (Think of a "gas-tank" as an analogy for the resource buffer, or a highway toll-card, both of which have occasionally "refills" but from which you can consume/bill on an as-needed basis).

The resource token is itself untradeable, and it cannot be exchanged back into system tokens. This limits is use as a vehicle for speculation, which helps ensure that applications can subsidize resources on behalf of users, which improves user-experience by further eliminating the need for users to manually manage resources.

### Client-side resource management

This PR adds some logic into the client-side supervisor infrastructure that automatically injects the necessary resource recharge transactions on behalf of users when their resource buffer gets below a configurable threshold. This only works if the user maintains a minimum balance of system tokens from which to exchange and acquire resource tokens.

### DIFF
--- a/packages/system/VirtualServer/plugin/src/lib.rs
+++ b/packages/system/VirtualServer/plugin/src/lib.rs
@@ -24,7 +24,8 @@ use psibase::AccountNumber;
 
 use psibase::services::tokens::{Precision, Quantity};
 use virtual_server::action_structs as Actions;
-use virtual_server::tables::tables as ServiceTables;
+use virtual_server::tables::tables::{self as ServiceTables};
+use virtual_server::DEFAULT_AUTO_FILL_THRESHOLD_PERCENT;
 
 struct VirtualServerPlugin;
 
@@ -242,7 +243,13 @@ impl Billing for VirtualServerPlugin {
         let capacity = TokensPlugin::helpers::decimal_to_u64(sys_id, &new_capacity)?;
         add_action_to_transaction(
             Actions::conf_buffer::ACTION_NAME,
-            &Actions::conf_buffer { capacity }.packed(),
+            &Actions::conf_buffer {
+                config: Some(ServiceTables::BufferConfig {
+                    capacity,
+                    threshold_percent: DEFAULT_AUTO_FILL_THRESHOLD_PERCENT,
+                }),
+            }
+            .packed(),
         )?;
 
         refill_to_capacity(Some(capacity), true)
--- a/packages/system/VirtualServer/service/src/lib.rs
+++ b/packages/system/VirtualServer/service/src/lib.rs
@@ -2,6 +2,7 @@
 
 pub mod rpc;
 pub mod tables;
+pub use crate::tables::DEFAULT_AUTO_FILL_THRESHOLD_PERCENT;
 
 /// Virtual Server Service
 ///
@@ -198,22 +199,13 @@ mod service {
         buy_res_for(amount, get_sender(), None);
     }
 
-    /// Allows the sender to request client-side tooling to automatically attempt to
-    /// reserve additional resources when the user is at or below the specified threshold
-    /// (specified in integer percentage values).
+    /// Allows the sender to manage the behavior of client-side tooling with respect to the
+    /// automatic management of the sender's resource buffer.
     ///
-    /// A threshold of 0 means that the client should not attempt to automatically manage
-    /// the user's reserved tokens.
+    /// If `config` is None, the account will use a default configuration
     #[action]
-    fn conf_auto_fill(threshold_percent: u8) {
-        UserSettings::get(get_sender()).set_auto_fill(threshold_percent);
-    }
-
-    /// Allows the sender to specify the capacity of their buffer of reserved system tokens.
-    /// The larger the buffer, the less often the sender will need to refill it.
-    #[action]
-    fn conf_buffer(capacity: u64) {
-        UserSettings::get(get_sender()).set_capacity(capacity);
+    fn conf_buffer(config: Option<BufferConfig>) {
+        UserSettings::get(get_sender()).configure_buffer(config);
     }
 
     fn bill(user: AccountNumber, amount: u64) {
@@ -224,7 +216,7 @@ mod service {
         if let Some(config) = BillingConfig::get() {
             let sys = config.sys;
 
-            let balance = UserSettings::get_resource_balance(user);
+            let balance = UserSettings::get(user).get_resource_balance();
             let amt = Quantity::new(amount);
 
             if balance < amt {
@@ -425,7 +417,7 @@ mod service {
         }
 
         let cpu_pricing = CpuPricing::get_assert();
-        let res_balance = UserSettings::get_resource_balance(account).value;
+        let res_balance = UserSettings::get(account).get_resource_balance().value;
 
         let price_per_unit = cpu_pricing.price();
         let ns_per_unit = cpu_pricing.billable_unit;
--- a/packages/system/VirtualServer/service/src/tables/user_settings.rs
+++ b/packages/system/VirtualServer/service/src/tables/user_settings.rs
@@ -5,7 +5,7 @@ use psibase::*;
 
 // Relatively arbitrary initial values
 const EXAMPLE_LARGE_TX_COST: u64 = 3_000_000; // e.g. Net cost of a 3MB upload when NET is at minimum cost
-const DEFAULT_AUTO_FILL_THRESHOLD_PERCENT: u8 = 20;
+pub const DEFAULT_AUTO_FILL_THRESHOLD_PERCENT: u8 = 20;
 
 impl UserSettings {
     pub fn get(user: AccountNumber) -> Self {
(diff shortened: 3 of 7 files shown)


packages/system/VirtualServer/plugin/src/lib.rs
```diff
@@ -0,0 +1,302 @@
...
+use psibase::fracpack::Pack;
+use psibase::AccountNumber;
+
+use psibase::services::tokens::{Precision, Quantity};
+use virtual_server::action_structs as Actions;
+use virtual_server::tables::tables as ServiceTables;
+
+struct VirtualServerPlugin;
+
+fn assert_caller(allowed: &[&str], context: &str) {
```
swatanabe: Why a direct check instead of permissions?

packages/system/Credentials/service/src/lib.rs
```diff
@@ -15,6 +20,23 @@ pub mod tables {
...
+                return;
+            }
+            table.put(&InitRow {}).unwrap();
+
+            Tokens::call().setUserConf(BalanceFlags::MANUAL_DEBIT.index(), true);
+            Nft::call().setUserConf(NftHolderFlags::MANUAL_DEBIT.index(), true);
+
+            // Turn off client-side auto-fill for resources when submitting tx
+            // using a credential
+            VirtualServer::call().conf_auto_fill(0);
```
swatanabe: Does this actually work? `virtual-server::auto-fill-gas-tank` checks `accounts::get_current_user`, but users of credentials (Invite, at least) call `transact::hook_action_sender` instead of setting the logged in user.

packages/system/VirtualServer/service/src/lib.rs
```diff
@@ -0,0 +1,516 @@
...
+    fn net_rates(doubling_time_sec: u32, halving_time_sec: u32) {
+        check(get_sender() == get_service(), "Unauthorized");
+        NetPricing::set_change_rates(doubling_time_sec, halving_time_sec);
+    }
+
+    /// Sets the number of blocks over which to compute the average network usage.
+    /// This average usage is compared to the network capacity (every block) to determine
+    /// the price of network bandwidth.
+    #[action]
+    fn net_blocks_avg(num_blocks: u8) {
```
swatanabe: 0 is not valid, and will brick the chain.

packages/system/VirtualServer/service/src/tables/cpu_pricing.rs
```diff
@@ -0,0 +1,134 @@
...
+    }
+
+    pub fn get() -> Option<Self> {
+        CpuPricingTable::read().get_index_pk().get(&())
+    }
+
+    pub fn set_num_blocks_to_average(num_blocks: u8) {
+        let table = CpuPricingTable::read_write();
+        let mut cpu_time = check_some(table.get_index_pk().get(&()), "CPU time not initialized");
+        cpu_time.num_blocks_to_average = num_blocks;
```
swatanabe: If `num_blocks` is reduced, `cpu_time.usage_history` needs to be truncated, either here or in `new_block`.

packages/system/VirtualServer/service/src/tables/cpu_pricing.rs
```diff
@@ -0,0 +1,134 @@
...
+        table.put(&cpu_time).unwrap();
+    }
```
Velua:
```suggestion
    fn update_cpu<F: FnOnce(&mut CpuPricing)>(f: F) {
        let table = CpuPricingTable::read_write();
        let mut cpu = check_some(table.get_index_pk().get(&()), "CPU time not initialized");
        f(&mut cpu);
        table.put(&cpu).expect("CpuPricing save failed");
    }

    pub fn set_num_blocks_to_average(num_blocks: u8) {
        Self::update_cpu(|cpu| cpu.num_blocks_to_average = num_blocks);
    }
```

packages/system/VirtualServer/service/src/tables/pricing_common.rs
```diff
@@ -0,0 +1,123 @@
...
+pub const PPM: u128 = 1_000_000;
+
+pub fn ratio_to_ppm(num: u64, den: u64) -> u128 {
+    (num as u128 * PPM) / (den as u128)
+}
+
+pub const LN2_PPM: u32 = 693_147;
+
+pub fn time_to_rate_ppm(t: u32) -> u32 {
+    (LN2_PPM + t / 2) / t
```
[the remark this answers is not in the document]
James-Mart: Added doc comments

packages/system/VirtualServer/service/src/tables/billing_config.rs
```diff
@@ -0,0 +1,92 @@
...
+             - 3MB = 3,000,000 bytes = 300.0000 resource tokens in network bandwidth cost alone when cost
+               is at minimum
+             - Therefore default capacity of 10,000 resource tokens makes 20% == 2,000, which is
+               plenty to afford a single large transaction
+        */
+        const DEFAULT_RESOURCE_BUFFER_SIZE: u64 = 10_000;
+
+        let tok = Tokens::call();
+        let sys = check_some(tok.getSysToken(), "System token not found");
+        DEFAULT_RESOURCE_BUFFER_SIZE * 10u64.pow(sys.precision.value() as u32)
```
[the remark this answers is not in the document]
James-Mart: Fixed (removed) the dependency on precision.

There are really two concepts, a default and a minimum. The logic is related to coming up with a default (even though the function is called `get_min_...`). I've now separated it into two functions:
* `get_min_resource_buffer`
* `get_default_resource_buffer`

packages/system/VirtualServer/service/src/tables/user_settings.rs
```diff
@@ -10,22 +15,43 @@ impl UserSettings {
...
+            abort_message(&err_msg);
+        }
+
         let mut settings = self;
         settings.buffer_capacity = Some(capacity);
         UserSettingsTable::read_write().put(&settings).unwrap();
     }
 
+    pub fn get_buffer_capacity(&self) -> u64 {
+        let default_capacity = self.get_min_buffer_capacity() * 5;
```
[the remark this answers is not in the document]
James-Mart: Reworked so there's a single action that sets both, or setting to `None` can be used to restore to the defaults.

libraries/psibase/common/include/psibase/Actor.hpp
```diff
@@ -68,7 +68,13 @@ namespace psibase
          {
             psio::shared_view_ptr<exportType<R>> result(psio::size_tag{result_size});
             raw::getResult(result.data(), result_size, 0);
-            check(result.validate(), "value returned was not serialized as expected");
+            if (!result.validate())
+            {
+               auto action      = psio::from_frac<Action>(psio::prevalidated{packed_action});
```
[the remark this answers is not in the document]
James-Mart: Done

PR #1681, opened 2026-01-28
Rework socket management

swatanabe:
This makes is possible to implement an async close that doesn't race with `recv`. The core problem before was that the only way to close a socket was by setting `autoClose`, which would fail if `recv` was running. Detecting this and asking `recv` to close also didn't work, because the lock held by recv was released by native after `recv` finished.

### DIFF
--- a/programs/psinode/init.cpp
+++ b/programs/psinode/init.cpp
@@ -18,6 +18,9 @@ namespace
    struct TempSocket : AutoCloseSocket, std::enable_shared_from_this<TempSocket>
    {
       TempSocket() {}
+      // TempSocket is used synchronously, so instead of posting remove
+      // to an executor in onClose, the creator is responsible for
+      // calling remove.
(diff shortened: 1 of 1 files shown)


programs/psinode/init.cpp
```diff
@@ -127,7 +127,21 @@ void load_local_packages(TransactionContext&             tc,
...
       }
       tc.transactionTrace.actionTraces.clear();
+      tc.ownedSockets.close(*tc.blockContext.writer, *tc.blockContext.systemContext.sockets,
+                            std::nullopt);
+      if (!socket->closed)
+         throw std::runtime_error("Socket was not closed");
+      if (socket->notifyClose)
+         throw std::runtime_error(
+             "Not implemented: socket notifyClose during initial service load");
+      socket->closed = false;
```
James-Mart: I don't really understand what's going on here. Socket is closed, then reopened, then closed again?
Consider adding a comment

PR #1686, opened 2026-01-30
Accounts app: add and remove accounts

brandonfancher:
- User can now remove accounts from login prompt page
- Adds the accounts management UI to the Accounts homepage (which was previously blank)
    - User can import accounts
    - User can remove accounts
    - UI features which only make sense within the prompt context are sensibly adjusted

### DIFF
--- a/.github/workflows/ubuntu-builder.yml
+++ b/.github/workflows/ubuntu-builder.yml
@@ -26,7 +26,7 @@ jobs:
         id: ccache_cache_timestamp
         run: echo timestamp=`date -u +"%Y-%m-%d-%H-%M-%S"` >>$GITHUB_OUTPUT
       - name: Maximize build space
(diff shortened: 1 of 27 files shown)


packages/system/Accounts/ui/src/router.tsx
```diff
@@ -13,13 +13,21 @@ const Router = createBrowserRouter([
...
         children: [
             {
-                path: "/plugin/web/prompt/connect",
+                path: "/",
                 element: <ConnectPrompt />,
             },
             {
-                path: "/plugin/web/prompt/import",
+                path: "/import",
                 element: <ImportPrompt />,
```
cool-ant: These look like a loss of context that, perhaps, James was planning for multi-platform support. Has he weighed in/aligned with this?

cool-ant (approved):
Just the one question; rest looks good

PR #1691, opened 2026-02-03
Config: simplify login/logout UI

brandonfancher:
This removes account switching from _within_ the Config app, instead relying on the Accounts app for account switching. We already implemented this simplification in Homepage. Mitigates potential user confusion.

This also hoists a query hook to shared-ui from Homepage. And while I was in there, I fixed a component that was badly breaking the React rules for hooks (`contact-details.tsx`).


James-Mart:
~~Console~~ -> Config?

PR #1692, opened 2026-02-05
Integration of invites with resource-billing

James-Mart:
# Background

With the recent introduction of [billing](https://github.com/gofractally/psibase/pull/1676), one of the drawbacks of the initial implementation was that the account who authorized the first action in the transaction was responsible for paying the resource costs for the full transaction. In most cases this works, but it does not work for services that provide alternative transaction authorization mechanisms, such as the credentials service. 

In the case of credential-authorized transactions, the sender account is a fixed account, and the credential being used is derived from the tx claims/proofs. 

# Solution

As of this PR, the account responsible for billing may, at any time, call into the virtual-service service and direct any subsequent billing to a subaccount. 

If this is done by the auth service of the billed account during the initial authorization (as implemented in this PR for credentials), then it is early enough that *all* transaction billing will be directed to the specified subaccount.

# Invites

The aforementioned solution to billing subaccounts was a prerequisite to the integration of our invite system with resource billing (since invites use credentials under-the-hood). Therefore, this PR also integrates invites with billing such that an invite now costs some minimum amount of system tokens to generate, and those tokens are used for two purposes:

1. Subsidizing the underlying credential so that it can be used by the receiver of the invite
1. Subsidizing resources for accounts created by the invite

### DIFF
--- a/packages/user/Tokens/service/src/tables.rs
+++ b/packages/user/Tokens/service/src/tables.rs
@@ -896,26 +896,6 @@ pub mod tables {
             SubAccountTable::read_write().erase(&(&self.pk()));
         }
 
-        pub fn force_delete(&mut self) {
-            let sub_bal_table = SubAccountBalanceTable::read_write();
-
-            let sub_balances: Vec<_> = sub_bal_table
-                .get_index_pk()
-                .range((self.id, 0)..(self.id, u32::MAX))
-                .collect();
-
-            for sub_balance in sub_balances {
-                // Move any sub-account balances to the primary balance
-                if sub_balance.balance.value > 0 {
-                    Balance::get_or_new(self.account, sub_balance.token_id)
-                        .add_balance(sub_balance.balance);
-                }
-                sub_bal_table.erase(&sub_balance.pk());
-            }
-
-            SubAccountTable::read_write().erase(&(&self.pk()));
-        }
-
         pub fn add_balance(&mut self, token_id: TID, quantity: Quantity) {
             Balance::get_assert(self.account, token_id).sub_balance(quantity);
             SubAccountBalance::get_or_new(self.id, token_id).add_balance(quantity);
@@ -990,10 +970,6 @@ pub mod tables {
 
         fn save(&self) {
             SubAccountBalanceTable::read_write().put(&self).unwrap();
-        }
-
-        pub fn delete(&mut self) {
-            SubAccountBalanceTable::read_write().erase(&self.pk());
         }
 
         fn add_balance(&mut self, quantity: Quantity) {


packages/system/VirtualServer/service/src/lib.rs
```diff
@@ -199,6 +287,35 @@ mod service {
...
+    /// the current transaction.
+    ///
+    /// This will only succeed if the sender has already been set by `Transact` to be the
+    /// billable account.
+    #[action]
+    fn bill_to_sub(sub_account: String) {
+        let billable_account = tx_cache::get_billable_account();
+        if billable_account.is_none() {
+            // Only allow overriding billable account after Transact has initialized it
+            return;
```
swatanabe: Should this be an error? If the caller intended `bill_to_sub` to take effect but it silently does nothing, it could be an exploitable bug.

packages/system/VirtualServer/service/src/lib.rs
```diff
@@ -199,6 +287,35 @@ mod service {
...
+    /// Allows the sender to specify one of their sub-accounts as the billable account for
+    /// the current transaction.
+    ///
+    /// This will only succeed if the sender has already been set by `Transact` to be the
+    /// billable account.
+    #[action]
+    fn bill_to_sub(sub_account: String) {
+        let billable_account = tx_cache::get_billable_account();
+        if billable_account.is_none() {
+            // Only allow overriding billable account after Transact has initialized it
```
swatanabe: `bill_to_sub` doesn't override billable account, though.

rust/psibase/src/sha256.rs
```diff
@@ -0,0 +1,6 @@
+use sha2::{Digest, Sha256};
+pub fn sha256(data: &[u8]) -> [u8; 32] {
```
swatanabe: If we're going to have a `sha256` function in the library, it should return `Checksum256`.

packages/user/Tokens/include/services/user/tokenTypes.hpp
```diff
@@ -132,4 +138,156 @@ namespace UserService
...
+         uint64_t fractional_part = quantity.value % divisor;
+
+         std::string result = std::to_string(integer_part);
+         result.reserve(result.length() + 1 + precision.value);
+         result += '.';
+         result += format_fractional_part(fractional_part, precision.value);
+         return result;
+      }
+
+      static Decimal fromStr(std::string_view s)
```
swatanabe: Is "." valid?

packages/user/Invite/src/Invite.cpp
```diff
@@ -163,10 +262,24 @@ void Invite::delInvite(uint32_t inviteId)
...
+   if (sysRecord.has_value())
+   {
+      auto sys           = sysRecord->id;
+      auto sub_account   = std::to_string(invite->id);
+      auto balanceRecord = to<Tokens>().getSubBal(sys, sub_account);
+
+      if (balanceRecord.has_value())
+      {
+         auto balance = *balanceRecord;
+         to<Tokens>().fromSub(sys, sub_account, balance);
```
swatanabe: `tokens::fromSub` leaves a 0 balance. It doesn't remove the record.

packages/user/Invite/src/Invite.cpp
```diff
@@ -85,23 +78,91 @@ void Invite::init()
...
+                                      ONE_WEEK.count(),  //
+                                      std::vector<MethodNumber>{"createAccount"_m});
 
-   auto         inviteTable = Tables().open<InviteTable>();
+   if (isBilling)
+   {
+      auto sys                   = getSysToken();
+      auto create_action_reserve = get_create_action_reserve();
+      auto gas_tank_reserve      = get_std_buffer_reserve();
+      auto needed = (create_action_reserve + gas_tank_reserve) * static_cast<uint64_t>(numAccounts);
```
swatanabe: Can this overflow?

packages/system/Credentials/service/src/tables.rs
```diff
@@ -0,0 +1,117 @@
...
+            let table = CredentialTable::read_write();
+            check(
+                table.get_index_by_pkh().get(&pubkey_hash_hex).is_none(),
+                "Credential already exists",
+            );
+
+            table
+                .put(&Credential {
+                    id,
+                    pubkey_hash_hex: pubkey_hash_hex.clone(),
```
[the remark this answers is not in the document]
James-Mart: Removed

packages/system/VirtualServer/service/src/lib.rs
James-Mart: At one point speculative execution was the only method by which this could be called without transact having first set the billable account. But now that the `FIRST_AUTH_FLAG` flag is checked in credentials `checkAuthSys`, this is now safe to be an error.

Changed.

rust/psibase_macros/psibase-macros-lib/src/service_macro/mod.rs
```diff
@@ -382,6 +402,32 @@ fn process_mod(
...
+                    #psibase_mod::RunAsCaller {
+                        sender,
+                        service: Self::#constant,
+                        allowed_actions: vec![],
+                    }
+                    .into()
+                }
+
+                #[doc = #call_as_with_doc]
+                pub fn call_as_with(
```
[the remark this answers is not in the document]
James-Mart: Changed to `call_as_extend` - how's that?

packages/system/VirtualServer/service/src/tables/user_settings.rs
```diff
@@ -42,9 +42,19 @@ impl UserSettings {
         }
     }
 
-    pub fn get_resource_balance(&self) -> Quantity {
+    pub fn to_sub_account_key(account: AccountNumber, sub_account: &str) -> String {
+        format!("{}-{}", account, sub_account)
```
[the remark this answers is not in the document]
swatanabe: If the delimiter is a valid character, the subaccount key could be the same as the regular key for a different account.

packages/system/VirtualServer/service/cpp/include/services/system/VirtualServer.hpp
```diff
@@ -269,9 +269,12 @@ namespace SystemService
 
       void notifyBlock(psibase::BlockNum block_num);
 
-      /// This actions sets the CPU limit for the specified account.
-      /// If the current tx exceeds the limit (ns), then the tx will timeout and fail.
-      void setCpuLimit(psibase::AccountNumber account);
+      /// This actions specifies which account is primarily responsible for
```
swatanabe: "actions" doesn't agree with "specifies"

packages/user/Tokens/service/src/tables.rs
```diff
@@ -887,8 +922,19 @@ pub mod tables {
...
         pub fn sub_balance(&mut self, token_id: TID, quantity: Quantity) {
-            SubAccountBalance::get_assert(self.id, token_id).sub_balance(quantity);
+            let remaining = SubAccountBalance::get_assert(self.id, token_id).sub_balance(quantity);
             Balance::get_or_new(self.account, token_id).add_balance(quantity);
+
+            if remaining.value == 0 && !self.manual_deletion {
+                let keep = SubAccountBalanceTable::read()
+                    .get_index_pk()
+                    .range((self.id, 0)..(self.id, u32::MAX))
+                    .any(|sub_balance| sub_balance.balance.value > 0);
```
swatanabe: If `manual_deletion` is not set, then there should be no zero balances, so we can only need to check whether any balance exists.

swatanabe (approved):

PR #1699, opened 2026-02-10
Fixing graphql query error when billing not enabled

cool-ant:
Querying virtual-server for resources/gas tank info was causing an error when billing wasn't enabled. This handles that event properly.

The result of the query now looks like this
{
  "data": null,
  "errors": [
    {
      "message": "billing not enabled",
      "locations": [
        {
          "line": 33,
          "column": 3
        }
      ],
      "path": [
        "userResources"
      ]
    }
  ]
}

### DIFF
--- a/packages/system/VirtualServer/service/src/rpc.rs
+++ b/packages/system/VirtualServer/service/src/rpc.rs
@@ -253,7 +253,9 @@ impl Query {
     async fn user_resources(&self, user: AccountNumber) -> async_graphql::Result<UserResources> {
         self.check_user_auth(user)?;
 
-        let config = BillingConfig::get_assert();
+        let config = BillingConfig::get()
+            .filter(|c| c.enabled)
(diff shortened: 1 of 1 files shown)


packages/system/VirtualServer/service/src/rpc.rs
```diff
@@ -253,7 +253,9 @@ impl Query {
     async fn user_resources(&self, user: AccountNumber) -> async_graphql::Result<UserResources> {
         self.check_user_auth(user)?;
 
-        let config = BillingConfig::get_assert();
+        let config = BillingConfig::get()
+            .filter(|c| c.enabled)
```
swatanabe: Shouldn't user resources be accessible before billing is enabled? I thought we need to fill the resource buffers before enabling billing.

James-Mart (approved):
PR description isn't quite right anymore

PR #1711, opened 2026-02-15
Add COST package

Velua:
Adds a package implementing Common Ownership Self-Assessed Tax (COST). 

Enables any account to create and manage a namespace of unique identifiers (treated as assets). After initial creation, assets are always listed for sale and purchased by others at the current owner's self-set valuation. Ownership requires payment of an annual tax rate and maintaining payments via prepaid balances; unpaid taxes lead to arrears and a Dutch auction discount on the sale price.


Velua:
- [ ] Change the manager to just an administration NFT, allows easy sale of the entire namespace and its proceeds.

PR #1713, opened 2026-02-17
Modernise NFT service

### LINKED ISSUE #1632: Modernize NFT service
1. Rework events to follow newer conventions (similar to [tokens](https://github.com/gofractally/psibase/pull/1467))
2. Protect queries by logged-in-user as appropriate (similar to [tokens](https://github.com/gofractally/psibase/pull/1576))
3. Add a single index to make paginated queries for pending nft transfers (similar to [tokens](https://github.com/gofractally/psibase/pull/1588))

Velua:


packages/user/Nft/service/src/lib.rs
```diff
@@ -223,12 +242,14 @@ pub mod tables {
                 "Only the creditor may perform this action",
             );
 
-            super::Wrapper::emit().history().uncredited(
+            super::Wrapper::emit().history().ownerChange(
                 self.nftId,
-                self.creditor,
+                "uncredit".to_string(),
```
James-Mart: uncredited*

packages/user/Nft/service/src/lib.rs
```diff
@@ -405,19 +461,13 @@ pub mod service {
...
-
-    #[event(merkle)]
-    pub fn transferred(nftId: NID, creditor: AccountNumber, debitor: AccountNumber, memo: String) {}
+    pub fn ownerChange(
+        nftId: NID,
+        action: String,
+        prev_owner: AccountNumber,
+        new_owner: AccountNumber,
+        memo: Memo,
+    ) {}
```
James-Mart: This event isn't quite right.

Should be something like:
```
 pub fn ownerChange(
        nftId: NID,
        action: String,
        account: AccountNumber,
        counter_party: AccountNumber,
        memo: Memo,
    ) 
```

Notice that column 2 is not the previous owner, it needs to stably represent the account to whom the event applies, and column 3 is the counterparty.

### e.g. Credit

`ownerChange(id, "credited", creditor, debitor, memo)`

And if auto-debit is enabled, a second event fires:
`ownerChange(id, "debited" debitor, creditor, memo)`

packages/user/Nft/service/src/lib.rs
```diff
@@ -299,21 +365,11 @@ pub mod service {
...
             };
 
-            add_index("minted", 0, DbId::HistoryEvent);
-            add_index("minted", 1, DbId::HistoryEvent);
-            add_index("burned", 0, DbId::HistoryEvent);
-            add_index("burned", 1, DbId::HistoryEvent);
+            add_index("ownerChange", 0, DbId::HistoryEvent);
+            add_index("ownerChange", 1, DbId::HistoryEvent);
+            add_index("ownerChange", 2, DbId::HistoryEvent);
+            add_index("ownerChange", 3, DbId::HistoryEvent);
```
James-Mart, line 371: You shouldn't need an index for column 1.

I guess having an index for column 0 makes sense, but querying that is probably not allowed by default...

packages/user/Nft/query-service/src/lib.rs
James-Mart, line 1: Missing authenticated query checks

packages/user/Nft/query-service/src/events.rs
James-Mart, line 1: This doesn't really follow the modern patterns.


1. `userConfEvent` should deserialize the flag index into the flag name. Actually does tokens even have an event for conf changed? I don't think we need a history of that, we really just need reading the current state (for which tokens uses `pub async fn settings(&self) -> BalanceFlagsJson`)
2. The action in `OwnerChangeEvent` should be emitted as a u8, and similarly deserialized into a string here at query time.

James-Mart:
I'm going to convert this to a draft so it stops showing up in my "active PR's to review" list.

packages/user/Nft/service/src/lib.rs
```diff
@@ -108,9 +114,9 @@ pub mod tables {
 
             super::Wrapper::emit().history().ownerChange(
                 new_instance.id,
-                "minted".to_string(),
-                0.into(),
+                MINTED,
                 sender,
+                0.into(),
                 "Minted".into(),
             );
```
James-Mart, line 121: Is service 0 preferable to just using `Wrapper::service`? As in, you minted the nft from the `nft` service, as opposed to from `bobs-nfts`

packages/user/Nft/query-service/src/lib.rs
```diff
@@ -30,10 +38,55 @@ mod service {
...
     #[Object]
     impl Query {
+        /// Paginated history of ownership changes for `account`.
+        ///
+        /// Requires the queried account to authorize the request through
+        /// `nft:plugin/authorized::graphql`.
+        async fn ownerChanges(
+            &self,
+            account: AccountNumber,
+            nft_id: Option<NID>,
```
James-Mart, line 66: Could also add counter-party as an optional param here, would allow a front-end to have a filter in your NFT wallet to only show interactions with some particular counter-party. It's better than doing such filtering client-side for pagination reasons.

packages/user/Nft/query-service/src/lib.rs
```diff
@@ -83,6 +137,46 @@ mod service {
             .await
         }
 
+        /// Pending NFT credits/debits for `user`.
+        ///
+        /// Requires the queried account to authorize the request through
+        /// `nft:plugin/authorized::graphql`.
+        async fn user_pending(
```
James-Mart, line 149: It feels to me like there are redundancies with the available queries, but it's hard for me to know what the right query service shapes are - probably better to actually build the NFT wallet interface like we did with the token wallet, and let that govern the refactors of this query service.

packages/user/Nft/service/src/lib.rs
James-Mart: This is the only comment that needs a response before I approve

packages/user/Nft/query-service/src/lib.rs
Velua: Agreed

packages/user/Nft/query-service/src/lib.rs
Velua: Added

packages/user/Nft/service/src/lib.rs
Velua: Added

James-Mart:
Need to merge with `main`. `host:common` was split into multiple plugins with separate concerns

James-Mart (approved):

PR #1738, opened 2026-02-26
Nix Dev Env

cool-ant:
This PR creates a dev environment with nix.
It adds flake.nix, flake.lock, and nix/ to the repo.
nix/README.md lays out how to spin it up.
What I've tested so far:
- full build
- building of individual items: rust bins and every component of a package
- loading extensions
- launch/boot/serving of a chain


libraries/psibase/sdk/pack_service.cmake
```diff
@@ -381,17 +381,28 @@ function(psibase_schema target)
         target_link_libraries(${target}-schema-gen PRIVATE "$<FILTER:$<TARGET_PROPERTY:${target},LINK_LIBRARIES>,EXCLUDE,.*(Psibase::service).*>" Psibase::test)
     endif()
 
+    # Fix for parallel build race condition: write_service_info() depends on the schema JSON file,
+    # but cmake only tracks file dependencies when the file is listed in OUTPUT (not BYPRODUCTS).
+    # When a concrete path is provided, we use it in OUTPUT so cmake knows to wait for schema generation.
+    # The original code also incorrectly depended on ${target} instead of ${target}-schema-gen.
```
swatanabe: Comments should describe the current code. They shouldn't describe some other version of the code. The commit message is the right place to put a description of what was wrong with the old code.

libraries/psibase/sdk/pack_service.cmake
```diff
@@ -381,17 +381,28 @@ function(psibase_schema target)
         target_link_libraries(${target}-schema-gen PRIVATE "$<FILTER:$<TARGET_PROPERTY:${target},LINK_LIBRARIES>,EXCLUDE,.*(Psibase::service).*>" Psibase::test)
     endif()
 
+    # Fix for parallel build race condition: write_service_info() depends on the schema JSON file,
+    # but cmake only tracks file dependencies when the file is listed in OUTPUT (not BYPRODUCTS).
+    # When a concrete path is provided, we use it in OUTPUT so cmake knows to wait for schema generation.
```
swatanabe: Using two different implementations like this is dodgy. The fact that the file's location is automatically determined doesn't mean that nothing depends on it.

libraries/psibase/sdk/pack_service.cmake
```diff
@@ -381,17 +381,28 @@ function(psibase_schema target)
         target_link_libraries(${target}-schema-gen PRIVATE "$<FILTER:$<TARGET_PROPERTY:${target},LINK_LIBRARIES>,EXCLUDE,.*(Psibase::service).*>" Psibase::test)
     endif()
 
+    # Fix for parallel build race condition: write_service_info() depends on the schema JSON file,
+    # but cmake only tracks file dependencies when the file is listed in OUTPUT (not BYPRODUCTS).
```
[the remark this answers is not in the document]
cool-ant: I've made my best effort to fix this. I *think* this fix is consistent both with your comments as well as a minimal, cmake-consistent mechanism, but this is outside my wheel house. So let me know if this works better, and if not, we'll have to spend a few minutes on the phone resolving this one together.
Ultimately, the issue that needs attention here was a race condition:
`write_service_info()` didn't have a proper dependency declaration on the `*-schema.json` files produced by `psibase_schema()`.

Velua (changes requested):

nix/README.md
```diff
@@ -0,0 +1,187 @@
...
+curl --proto '=https' --tlsv1.2 -sSf -L https://install.determinate.systems/nix | sh -s -- install
+```
+
+#### Fallback: the official installer
+
+```bash
+sh <(curl -L https://nixos.org/nix/install) --daemon
+```
+
+Note: the default installer requires you manually enable flakes in `~/.config/nix/nix.conf` or `/etc/nix/nix.conf`:
```
Velua: I'd be tempted to not even bother with installing Nix instructions, I feel like they'd get outdated quick enough and it should be implied it's already there given you've listed it as a pre-req, I'd just leave the Nix flakes bit.

flake.nix
```diff
@@ -0,0 +1,283 @@
...
+              [[ -r __BASH_COMPLETION__ ]] && source __BASH_COMPLETION__
+            fi
+            [[ -f "$HOME/.bashrc" ]] && . "$HOME/.bashrc"
+            NIX_BASH_INIT_EOF
+                sed -i "s|__BASH_COMPLETION__|${pkgs.bash-completion}/share/bash-completion/bash_completion|g" "$_nix_bash_init"
+              fi
+              if command -v direnv &> /dev/null; then
+                eval "$(direnv hook bash)"
+              fi
+            fi
```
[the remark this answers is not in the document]
Velua: Note that this doesn't occur if I do something like

```
nix develop
nix develop
``` 
It's only when I go in fresh from outside the nix'd shell with `nix develop` does this happen.

flake.nix
```diff
@@ -0,0 +1,283 @@
...
+              echo "  Rust:  $(rustc --version)"
+              echo "  Node:  $(node --version)"
+              echo "  Yarn:  $(yarn --version)"
+              echo "  WASI SDK: ${wasiSdk}"
+              echo ""
+              _check_cargo_tools
+              echo "To build psibase (clean build required on first run):"
+              echo "  rm -rf build && mkdir build && cd build"
+              echo "  cmake .. -DCMAKE_BUILD_TYPE=Release"
+              echo "  cmake --build . -j\$(nproc)"
```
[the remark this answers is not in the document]
cool-ant: among other things, yes. Updated to mirror what's in .vscode/scripts/build.sh

nix/README.md
```diff
@@ -0,0 +1,187 @@
...
+# git credentials configured in .git/config
+#[credential]
+#    helper = 
+#    helper = store --file=/home/mike/.git-credentials-work
+EOF
+direnv allow
+```
+
+#### 2.4 extensions
+Copy nix/docs/extensions.json.example to .vscode/extensions.json to pick up the standard dev extensions.
```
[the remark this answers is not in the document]
cool-ant: Noted as a VS Code and forks thing.

Velua (approved):

nix/README.md
cool-ant: NOTE: README.md is being reviewed in this PR: https://github.com/gofractally/psibase/pull/1913/

.gitignore
```diff
@@ -8,6 +8,7 @@
 .cursor
 .cursorignore
 .cursorrules
+.direnv
```
James-Mart: This needed?

James-Mart:
You're missing 
* mdbook-pagetoc
* mdbook-tabs

Also
* your mdbook-mermaid is at 0.15.0, whereas contributor pins it at 0.16.2
* your mdbook is slightly old (0.4.49 vs 0.4.52 in contributor)

None of this is critical, but your mdbook docs won't render correctly, if at all.

flake.nix
```diff
@@ -0,0 +1,318 @@
...
+          zstd
+          rustToolchain
+          cargoComponent
+          cargoGenerate
+          cargoEdit
+          binaryen
+          wabt
+          wasm-pack
+          wasm-tools
+          nodejs_20
```
James-Mart: Might not matter, but this is node `20.19.6` and currently `psibase-contributor` pins to `20.11.0`

nix/README.md
```diff
@@ -0,0 +1,160 @@
...
+- **Docs**: mdbook with mermaid plugin
+
+You do **not** need to pre-install Rust, Node, or other dev tools; `nix develop` provides everything.
+
+# Prerequisites
+
+- **Nix** (the pkg mgr, not the OS) *with flakes enabled*
+- **direnv** (optional but recommended)
+
+## Install Nix (Linux)
```
James-Mart, line 25: IMO this section shouldn't be included. If the instructions ever change, these instructions will be wrong - At most, link to [nix installation docs](https://nix.dev/install-nix.html)

nix/README.md
```diff
@@ -0,0 +1,160 @@
...
+experimental-features = nix-command flakes
+```
+
+### Configuring for NixOS
+
+```nix
+nix.settings.experimental-features = [ "nix-command" "flakes" ];
+```
+
+## Install direnv (Recommended but optional)
```
James-Mart: See comment about nix installation instructions [here](https://github.com/gofractally/psibase/pull/1738/changes#r3507042152).

Same comment but even moreso for direnv. Definitely don't need platform-specific installation instructions for optional software.

nix/README.md
```diff
@@ -0,0 +1,160 @@
...
+### 2.1 For VS Code and forks: run .vscode/scripts/env-setup.sh
+This will ensure your build and dev envs can find everything they need
+
+### 2.2 Initializing/Configuring the Environment (and IDE)
+
+#### Launching IDE within the nix-shell
+It's recommended to launch Cursor in the nix-shell so Cursor (and its terminals) have the normal dev environment (that `nix develop` sets up)
+So an easy option is to write a launch script that does the following
+- `nix develop`
+- HOST_IP=127.0.0.1 #This is determined automatically in a `psibase-contributor` container, but that obviously doesn't apply here
```
James-Mart: In this case, shouldn't the HOST IP be 127.0.0.1 on all platforms, in which case can't we just set that environment variable in the nix config?

nix/README.md
```diff
@@ -0,0 +1,160 @@
...
+- cursor [cursor workspace file] (or other IDE)
+
+Recommended extensions: clangd, rust-analyzer, wit-idl, ESLint, Prettier, direnv.
```
James-Mart: We can probably eliminate this line in favor of adding a `.vscode/extensions.json` file with the extensions listed in psibase-contributor's `devcontainer.json`.

nix/README.md
```diff
@@ -0,0 +1,160 @@
...
+From the psibase repo root:
+
+```bash
+cat >> .vscode/.envrc <<'EOF'
+# set IP for Launch/Continue scripts
+HOST_IP=127.0.0.1
+# git credentials configured in .git/config
+#[credential]
+#    helper = 
+#    helper = store --file=/path/to/.git-credentials-work
```
James-Mart: Is `[credential]` table needed or not? If not, let's just delete it instead of keeping it and having it commented out

nix/README.md
```diff
@@ -0,0 +1,160 @@
...
+
+Recommended extensions: clangd, rust-analyzer, wit-idl, ESLint, Prettier, direnv.
+
+#### (using direnv/.envrc, which activates the environment automatically when you `cd` into the repo)
+From the psibase repo root:
+
+```bash
+cat >> .vscode/.envrc <<'EOF'
+# set IP for Launch/Continue scripts
+HOST_IP=127.0.0.1
```
James-Mart: See [this comment](https://github.com/gofractally/psibase/pull/1738/changes#r3507061313) for an idea that should let us remove this line

nix/README.md
```diff
@@ -0,0 +1,160 @@
...
+#### Launching IDE within the nix-shell
+It's recommended to launch Cursor in the nix-shell so Cursor (and its terminals) have the normal dev environment (that `nix develop` sets up)
+So an easy option is to write a launch script that does the following
+- `nix develop`
+- HOST_IP=127.0.0.1 #This is determined automatically in a `psibase-contributor` container, but that obviously doesn't apply here
+- cursor [cursor workspace file] (or other IDE)
+
+Recommended extensions: clangd, rust-analyzer, wit-idl, ESLint, Prettier, direnv.
+
+#### (using direnv/.envrc, which activates the environment automatically when you `cd` into the repo)
```
James-Mart: 1. A parenthetical heading is odd
2. "actives the environment" is a bit confusing, makes it sound like you're talking about the .envrc in the repo causing `direnv` to automatically use the nix environment, which it doesn't - it just manages some of your path variables.

nix/README.md
```diff
@@ -0,0 +1,160 @@
+Nix-Based Development Environment for Psibase (Linux)
+
+How to use Nix as an alternative to the Docker-based development environment (psibase-contributor).
+**Supported platforms: Linux x86_64 and Linux aarch64.**
+Not (yet) supported: macOS
+
+# Overview
+
+The Nix configuration, emulating the current psibase-contributor config provides:
```
James-Mart: Not sure "emulating the current psibase-contributor config" is accurate. Plenty of deviations from it, all of which (I believe) are fine.

nix/README.md
```diff
@@ -0,0 +1,160 @@
...
+```bash
+cat >> .vscode/.envrc <<'EOF'
+# set IP for Launch/Continue scripts
+HOST_IP=127.0.0.1
+# git credentials configured in .git/config
+#[credential]
+#    helper = 
+#    helper = store --file=/path/to/.git-credentials-work
+EOF
+direnv allow
```
James-Mart: A better solution than `direnv allow` here is to add 
`~/.config/direnv/config.toml`
```
[whitelist]
prefix = ["/path/to/psibase"]
```

Because `direnv allow` is non-recursive so managed path variables will disappear if shells are not at the root of the workspace. It would also have to be run per-worktree in worktree-based workflows.

nix/README.md
```diff
@@ -0,0 +1,160 @@
...
+```
+
+## 3. Build psibase
+### Clean build required on first run (or after shell changes):
+Manually wipe your build dir, then run the build script.
```
James-Mart: When wiping the build dir is necessary, I consider that a bug

nix/README.md
```diff
@@ -0,0 +1,160 @@
...
+
+```bash
+.vscode/scripts/build.sh
+```
+
+Click the VS Code custom Build button or run the build manually.
```
James-Mart: No custom build button exists if the user doesn't install the `actboy168.tasks` extension, which also isn't listed in your recommended extensions list.

nix/README.md
```diff
@@ -0,0 +1,160 @@
...
+
+## 4. Launch a chain
+### Launch
+Press Launch to launch a change (manual hacky command below).
```
James-Mart: 1. No launch button exists if the user doesn't install the `actboy168.tasks` extension, which also isn't listed in your recommended extensions list.
2. Using the CLI isn't "hacky", it's the official method. If anything is hacky, it's the launch button.

nix/README.md
```diff
@@ -0,0 +1,160 @@
...
+
+Within the nix shell, replace the PROJECT_ROOT below with your repo root path:
+```bash
+PSIBASE_ADMIN_IP=127.0.0.1 psinode "$PROJECT_ROOT/db" -p myprod -l 7777
```
James-Mart: 1. Can PSIBASE_ADMIN_IP=127.0.0.1 be set from the nix config?
2. This command is pretty opinionated, prescribing db name, producer name, port... If you're going to have a sample command listed here at all, you should probably identify it as a "example launch command" (similar to other example commands in `.vscode/tasks.json`).

nix/README.md
```diff
@@ -0,0 +1,160 @@
...
+```
+
+### Boot
+Boot as you normally would from the cli or use the provided x-admin link and boot with the UI.
```
James-Mart: if "as you normally would" is acceptable here, I think it should be acceptable for launch as well. So you could just say, "launch and boot as you normally would with the CLI/X-admin"

nix/README.md
```diff
@@ -0,0 +1,160 @@
...
+
+# Optional: HTTPS and SoftHSM
+SKIP if you're doing a quick build and http test.
+
+- **HTTPS**: The shell includes `mkcert`. Run `mkcert -install`, create certs, and set `VITE_SECURE_LOCAL_DEV=true` and `VITE_SECURE_PATH_TO_CERTS` if needed.
```
James-Mart: 1. `VITE_SECURE_LOCAL_DEV=true` and `VITE_SECURE_PATH_TO_CERTS` are never needed
2. These instructions for https are incomplete. I'd think you can probably just remove them entirely, since it's not particularly related to nix. Someday we can have an https guide in the docs.

nix/README.md
```diff
@@ -0,0 +1,160 @@
...
+- **SoftHSM2**: `softhsm2-util --init-token --slot 0 --label "psibase SoftHSM" --pin "Ch4ng#Me!" --so-pin "Ch4ng#Me!"`
```
James-Mart: Since we aren't running `softhsm-util` automatically for the user, we don't need to prescribe a pin (`Ch4ng#Me!`).
* Example softhsm setup command: `softhsm2-util --init-token --slot 0 --label "psibase SoftHSM" --pin "<PRIVATE_PIN>" --so-pin "<PRIVATE_PIN>"`

nix/README.md
```diff
@@ -0,0 +1,160 @@
...
+# Environment variables set by the shell
+
+| Variable | Description |
+|----------|-------------|
+| `WASI_SDK_PREFIX` | Path to WASI SDK |
+| `WASI_SYSROOT` | WASI sysroot |
+| `CC` / `CXX` | Clang paths |
+| `LIBCLANG_PATH` | For rust-analyzer |
+| `RUST_SRC_PATH` | For rust-analyzer |
+| `IN_NIX_SHELL` | Set to `1` in the shell |
```
James-Mart: Out of curiosity, what's the point of `IN_NIX_SHELL`?

nix/README.md
```diff
@@ -0,0 +1,160 @@
...
+- **HTTPS**: The shell includes `mkcert`. Run `mkcert -install`, create certs, and set `VITE_SECURE_LOCAL_DEV=true` and `VITE_SECURE_PATH_TO_CERTS` if needed.
+- **SoftHSM2**: `softhsm2-util --init-token --slot 0 --label "psibase SoftHSM" --pin "Ch4ng#Me!" --so-pin "Ch4ng#Me!"`
+
+# Environment variables set by the shell
+
+| Variable | Description |
+|----------|-------------|
+| `WASI_SDK_PREFIX` | Path to WASI SDK |
+| `WASI_SYSROOT` | WASI sysroot |
+| `CC` / `CXX` | Clang paths |
```
James-Mart: @swatanabe won't this cause native to build with the wrong compiler?

nix/README.md
```diff
@@ -0,0 +1,160 @@
...
+
+- **HTTPS**: The shell includes `mkcert`. Run `mkcert -install`, create certs, and set `VITE_SECURE_LOCAL_DEV=true` and `VITE_SECURE_PATH_TO_CERTS` if needed.
+- **SoftHSM2**: `softhsm2-util --init-token --slot 0 --label "psibase SoftHSM" --pin "Ch4ng#Me!" --so-pin "Ch4ng#Me!"`
+
+# Environment variables set by the shell
+
+| Variable | Description |
+|----------|-------------|
+| `WASI_SDK_PREFIX` | Path to WASI SDK |
+| `WASI_SYSROOT` | WASI sysroot |
```
James-Mart: Is this needed? CMake derives it from `WASI_SDK_PREFIX`

nix/rust-toolchain.toml
James-Mart, line 1: Is this file generally useful everywhere? And if so, should it just be at `psibase` root? @cool-ant @swatanabe

nix/README.md
James-Mart: Also, it's not clear from these docs when this is needed. It's only needed if you want to use a softhsm for the producer's block-signing key,  and providing psinode with a `pkcs11-mod` path. Not needed for development networks, so probably not even really needed in these instructions.

nix/rust-toolchain.toml
cool-ant: I mean psibase-contributor doesn't need it, so... it's not useful unless someone's doing something odd (aside from Nix). So I wouldn't vote to move it to the root.

.gitignore
cool-ant: Hmm, at some point, but it appears no longer. removed.

James-Mart: Nothing currently creates a `.direnv` file (in psibase-contributor), does the nix env?

nix/README.md
cool-ant: Part of the reason this is included is just that the docs aren't the best option. Otherwise, I'd normally completely agree.

nix/README.md
cool-ant: Generally, I agree, but I assume you're not requiring here that a build dir built some other way (for instance using `psibase-contributor` reconfigured to using a bind mount) must be reliably buildable without wiping the build/ dir right? This was originally added just to be sure, give small changes in version are very possible switching to this environment, and I don't believe we've thoroughly tested the capacity of our CMake config to detect small version difference and reliably rebuild everything needed.
Besides CMake cache may have all kinds of things cached it just seemed simpler and more practical to wipe the dir and invest the 10 min in a fresh build.

nix/README.md
cool-ant: gives the user (and customized cli prompt) a way to know whether this shell is a nix-customized env or a standard shell.

nix/README.md
cool-ant: nope; removed

nix/README.md
James-Mart: > the docs aren't the best option. 

Why?

nix/README.md
James-Mart: I notice you didn't include a step for running `git clean -dxf` to make sure there's no bad source-tree files that will mess up the build, and you shouldn't - because it shouldn't be necessary. Same as deleting the build dir.

If anything this is a troubleshooting tip.

nix/README.md
```diff
@@ -0,0 +1,113 @@
...
+Launch Cursor (or another IDE) from the dev shell so the editor and its terminals inherit the flake environment (including `HOST_IP` for Launch/Continue tasks):
+
+```bash
+nix develop -c cursor /path/to/your.code-workspace
+```
+
+Or use a wrapper script (see `~/repos/cursor-workspaces/cursor-psibase-via-nix.sh`).
+
+## 3. Build and Launch psibase
+Build and Launch with the tasks.json buttons of by running the same command at the nix shell
```
James-Mart: or*

nix/README.md
James-Mart: @cool-ant FYI resolution on this comment is the only one preventing my approval. I thought cc/cxx had to be gcc, this says they are clang

cool-ant: cc/cxx is unset in `psibase-contributor`, leading it to pick up `gcc` from `build-essential` on ubuntu (base image). So `psibase-contributor` certainly doesn't enforce `gcc` as a requirement. However, this nix config grabbing `clang` I think was an unintended fallout of starting this project on my Mac, which defaults to `clang`. I can take the same approach as `psibase-contributor` and simply not specify a requirement. That'll lead to each person's system providing a (potentially different) compiler, or we can set it to `gcc` in an opinionated way, given that's how `psibase-contributor` will be as long as ubuntu underlies it.

James-Mart (dismissed):

libraries/triedent/CMakeLists.txt
```diff
@@ -12,6 +12,11 @@ add_library(triedent
 target_include_directories(triedent PUBLIC ${CMAKE_CURRENT_SOURCE_DIR}/include ${Boost_INCLUDE_DIRS})
 target_link_libraries(triedent PUBLIC Threads::Threads)
 
+if(CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
+   set(CMAKE_AR ${CMAKE_CXX_COMPILER_AR})
+   set(CMAKE_RANLIB ${CMAKE_CXX_COMPILER_RANLIB})
+endif()
+
```
cool-ant: @swatanabe @James-Mart This was the final, simplest fix I could come up with for a linker error I ran into. gcc 11 allowed the use of `ar` (more lenient), but gcc 15 forces the use of gcc-ar (which also works on gcc-11). So I'm just switching us to use `gcc-ar`. Wanted to make both of you aware of this last change, given one of your approvals is already in...

nix/README.md
Velua: Nix docs in general is notoriously dreadful, however, I don't think it's up to us to supplement that. 
I vote for deletion of this section as well as the following section, we mention Flakes as a requirement in `- **Nix** (the pkg mgr, not the OS) *with flakes enabled*`. 

```
Note: the default installer requires you manually enable flakes in `~/.config/nix/nix.conf` or `/etc/nix/nix.conf`:

experimental-features = nix-command flakes
```

libraries/triedent/CMakeLists.txt
swatanabe: The right fix is to use the [`INTERPROCEDURAL_OPTIMIZATION`](https://cmake.org/cmake/help/latest/prop_tgt/INTERPROCEDURAL_OPTIMIZATION.html#prop_tgt:INTERPROCEDURAL_OPTIMIZATION) property, instead of passing `-flto` directly.

cool-ant: Ok, I'll add that. Are you ok with the move to `gcc-ar` to unify the implementation between the 2 version?

swatanabe: The property tells CMake that it needs to use `gcc-ar`

cool-ant: Updated.

James-Mart (approved):
I approve with a noted exception that I don't understand the `INTERPROCEDURAL_OPTIMIZATION_RELEASE` CMake property so suggest waiting for swatanabe to confirm its used correctly

swatanabe (approved):

PR #1776, opened 2026-03-17
add initial UI for x-admin local package list

James-Mart:
This adds a list view in X-Admin with local packages from the index, design patterned after the config UI.

It lets you see installed local packages, versions, and easily update them.


brandonfancher (approved):
LGTM!

See my comment about moving the shared file. And make sure prettier runs on these files. There are some formatting issues on the `node-local-packages.tsx` file for sure, but check the others as well.

PR #1789, opened 2026-03-24
Update vite configs for x-proxy and hot reloading

brandonfancher:
This simplifies the shared vite config. Now running `yarn dev` on an app client works with x-proxy, giving us hot reloading nirvana.

I've updated a few app's vite configs to use the new shared config for demonstration and feedback-gathering purposes. If we're happy, I'll add a commit updating the remaining vite configs.

- [x] Test with fractal-core, given subdomain is dynamic (fixed by removing subdomain/service name from vite config entirely)
- [x] Roll out changes to app vite configs

## Subsequent PRs
- Figure out what we're going to do about homepage (no sub-sub-domain)
- Update x-proxy UI to support origin removal


cool-ant (approved):
LGTM

Velua (approved):

PR #1790, opened 2026-03-27
Homepage accessible on configurable subdomain

James-Mart:
# Overview

* This PR introduces a **canonical network-name subdomain** (default `network`) for the homepage UI, while keeping the `homepage` account as the deployment target. 
* The `branding` service’s `setNetworkName` creates the network-name account, and configures **Sites** to **reverse-proxy** that subdomain’s HTTP serving to the `homepage` service. The `homepage` service (and previously configured homepage subdomains) are automatically configured to `redirect` permanently to the new homepage subdomain.
* This PR also adds generic per-account **HTTP redirect** (`setRedirect` / `clearRedirect`) capabilities to the **HttpServer** app, so `homepage.<root>` can redirect to `network.<root>`. Requests to the root domain are also redirected to the homepage subdomain.
* Network rebranding through the **Config** plugin will coordinate all the necessary wiring: update branding service (which sets the reverse proxy / redirects).
## Concepts

- **`homepage`**  
  Account that actually holds deployed Homepage UI assets.

- **Network name** (e.g. `network`)  
  Account name used as the subdomain on which to access the homepage. Configred in `branding`.

- **Redirect (`HttpServer.setRedirect`)**  
  Per-account HTTP **308** redirect to another **sibling** subdomain (preserves the same path/query).

- **Reverse proxy (`Sites.setProxy`)**  
  Per-account config for a subdomain’s requests to be served using **content** from another account.
  This PR simply uses the pre-existing reverse proxy functionality to proxy the network name subdomain to the homepage service.

---

## Installing the homepage package

If the network installs the homepage package, then it will automatically:
* Set the network name to a default: "network"
  * Which reverse proxies requests to "network" to "homepage"
* Redirects requests to "homepage" to the network name.

## Overall flow

```mermaid
flowchart TB
  E_R(( )) --> R
  E_H(( )) --> H
  E_N(( )) --> N
  E_P(( )) --> P

  subgraph result[" "]
    R["Request root domain"]
    H["Request 'homepage' subdomain"]
    N["Request network subdomain"]
    P["Request previously configured network subdomain"]
    U["Homepage UI content"]
  end
  R -->|HttpServer 308| H
  H -->|HttpServer 308| N
  P -->|HttpServer 308| N
  N -->|Reverse proxy to homepage.rootDomain| U
```


brandonfancher (commented):
Code looks good. But after playing with it, there are a couple issues I want to highlight:

1. If the `homepage` app's name changes to something other than `homepage`, the cross-app whitelisting no longer works. For example, when it's `homepage`, opening the tokens UI doesn't prompt for permission. But it does if you change it.
2. After changing the network name a couple times, my browser keeps trying to redirect me to a old, nonexistent subdomain.

packages/user/Supervisor/ui/src/supervisor.ts
```diff
@@ -81,14 +81,18 @@ export class Supervisor implements AppInterface {
             "Redundant setting parent origination",
         );
 
-        if (callerOrigin === rootDomain) {
+        const service = serviceFromOrigin(callerOrigin);
+
+        if (service === "homepage") {
+            // TODO: we should check if service == the dynamically configured "homepage" service in httpserver.
+            // If it is, then we can internally namespace this app as "homepage".
```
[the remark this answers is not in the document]
James-Mart: Fixed

packages/local/XHttp/src/XHttp.cpp
```diff
@@ -559,6 +559,16 @@ void XHttp::startSession()
    recurse().to<XAdmin>().startSession();
 }
 
+bool XHttp::is_secure(int32_t socket)
```
swatanabe: Consider checking that the sender is the owner of the socket, to prevent information leakage.

brandonfancher (approved):
Working well!

packages/user/Homepage/postinstall.json
```diff
@@ -1 +1,34 @@
...
+        "service": "sites",
+        "method": "enablespa",
+        "data": {
+            "enable": true
+        }
+    },
+    {
+        "sender": "homepage",
+        "service": "sites",
+        "method": "setRedirect",
```
swatanabe: The interaction between `homepage => setRedirect(network)` and `network => setProxy(homepage)` is confusing. Just going by the action descriptions, I would expect it to cause a redirect loop.

libraries/psibase/common/src/HttpHeaders.cpp
```diff
@@ -220,3 +223,22 @@ std::vector<std::optional<IPAddress>> psibase::forwardedFor(const HttpRequest& r
    }
    return result;
 }
+
+std::optional<std::string> psibase::forwardedProto(const HttpRequest& request)
+{
+   for (const auto& value : request.getHeaderValues("x-forwarded-proto"))
```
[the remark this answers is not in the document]
James-Mart: Good point, fixed

packages/local/XHttp/src/XHttp.cpp
James-Mart: I think this is how I do that. 
Did I check the necessary tables/fields?

packages/system/HttpServer/src/HttpServer.cpp
```diff
@@ -322,6 +356,45 @@ namespace SystemService
...
+      {
+         redirectUrlHost = destination.str();
+         redirectUrlHost.push_back('.');
+         redirectUrlHost.append(rootHost);
+      }
+      else
+      {
+         auto prefixLen  = req.host.size() - rootHost.size() - 1;
+         redirectUrlHost = req.host;
+         redirectUrlHost.replace(0, prefixLen, destination.str());
```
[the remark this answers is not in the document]
James-Mart: Good point, updated - I just assert if the request isn't to roothost or a subdomain

packages/system/HttpServer/src/HttpServer.cpp
```diff
@@ -322,6 +356,45 @@ namespace SystemService
...
+
+   std::string HttpServer::getSiblingUrl(HttpRequest                 req,
+                                         std::optional<std::int32_t> socket,
+                                         AccountNumber               destination)
+   {
+      auto rootHost = to<XHttp>().rootHost(req.host);
+
+      check(req.host.size() >= rootHost.size(), "request host invalid: \"" + req.host + "\"");
+
+      std::string redirectUrlHost;
```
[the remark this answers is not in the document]
James-Mart: Copypasta from before I moved it into `getSiblingUrl` - fixed

packages/system/HttpServer/src/HttpServer.cpp
```diff
@@ -322,6 +356,38 @@ namespace SystemService
...
+      if (auto proto = forwardedProto(req))
+         scheme = *proto + "://";
+      else if (socket)
+         scheme = to<XHttp>().isSecure(*socket) ? "https://" : "http://";
+      else
+         scheme = isLocalhost(req) ? "http://" : "https://";
+
+      std::string url = scheme + siblingUrlHost;
+
+      if (auto suf = hostHeaderPortSuffix(req); !suf.empty())
```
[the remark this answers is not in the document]
James-Mart: Fixed

packages/system/HttpServer/include/services/system/HttpServer.hpp
```diff
@@ -107,7 +114,8 @@ namespace SystemService
                 method(registerServer, server),
                 method(recv, socket, data, flags),
                 method(serve, socket, req),
-                method(rootHost, host))
+                method(rootHost, host),
+                method(getSiblingUrl, req, socket, destination))
```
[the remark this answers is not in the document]
James-Mart: Fixed

packages/user/Homepage/postinstall.json
James-Mart: Addressed

swatanabe: Why does `/common` need to take precedence over the redirect? The reason I concluded that HttpServer needs to handle the redirect is precisely because I thought that `/common` should be redirected, too. If we don't want to redirect `/common`, then a separate redirect service has the advantage that it will automatically be overridden by a new `registerServer`.

I've looked through what's in common:
- common-lib won't be accessed via a redirect, because the pages that include it should be redirected
- static content in general will be the same whether redirected or not. The only difference is the current URL, which I think will always make more sense if it is redirected. The plugin tester, in particular should be redirected.
- `thisservice` could be interpreted to redirect or not. If everything else is redirected, I think it should be too.
- auth cookie manipulation will fail if redirected or be useless if not redirected
- The rest of CommonApi is the same whether redirected or not

James-Mart: > auth cookie manipulation will fail if redirected or be useless if not redirected

Great point

I've now made redirect take precedence even over common api requests.

packages/system/HttpServer/src/HttpServer.cpp
```diff
@@ -271,8 +288,41 @@ namespace SystemService
...
 
-      // First we check the registered server
-      auto                     service    = getTargetService(req, to<XHttp>().rootHost(req.host));
+      // Extract subdomain
+      auto root        = to<XHttp>().rootHost(req.host);
+      auto service_opt = getTargetService(req, root);
+
+      auto redirect = [&](const psibase::AccountNumber& subdomain)
+      {
+         auto loc  = getSiblingUrl(req, std::optional{sock}, subdomain, true);
```
swatanabe, line 297: Makes an unnecessary copy of `req`

packages/system/HttpServer/src/HttpServer.cpp
```diff
@@ -322,6 +373,54 @@ namespace SystemService
...
+          .account     = sender,
+          .destination = destination,
+      });
+   }
+
+   void HttpServer::clearRedirect()
+   {
+      auto table = Tables{getReceiver()}.open<RedirectTable>();
+      if (auto row = table.get(getSender()))
+         table.remove(*row);
```
swatanabe: `table.erase` is slightly more efficient for tables with only one index.

packages/user/Branding/service/src/lib.rs
```diff
@@ -27,18 +27,43 @@ mod tables {
...
             get_sender() == SERVICE,
             "Only the 'branding' account is authorized to set network branding",
         );
-        NetworkNameTable::new()
-            .put(&NetworkName { name: name.clone() })
-            .unwrap();
+
+        let account = AccountNumber::from_exact(name.as_str()).expect("Network name invalid");
+
+        AuthDelegate::call().newAccount(account, SERVICE, true);
```
swatanabe: Did you test changing the network name back to a previous value? I suspect it will cause a redirect loop.

packages/system/HttpServer/src/HttpServer.cpp
James-Mart: fixed

packages/system/HttpServer/src/HttpServer.cpp
James-Mart: fixed

packages/user/Branding/service/src/lib.rs
James-Mart: fixed

swatanabe (approved):

PR #1800, opened 2026-04-01
Unpack action arguments and results in traces

swatanabe:
Implemented in the CLI and the rust tester.

Schemas are cached, so each schema only needs to be fetched once in a each command. Schemas in packages being installed, and schemas loaded for packing actions will be reused, without being fetched again.


James-Mart (approved):

James-Mart:
This is so useful :1st_place_medal:

PR #1804, opened 2026-04-03
PremAccounts Market implementation

cool-ant:
Partial impl of https://github.com/gofractally/psibase-pm/issues/15

Includes:
- buy, claim premium account
- market creation, param update, enabe/disable


CMakeLists.txt
```diff
@@ -708,6 +708,14 @@ cargo_psibase_package(
     DEPENDS ${Identity_js_DEP}
 )
 
+cargo_psibase_package(
+    OUTPUT ${SERVICE_DIR}/PremAccounts.psi
+    PATH packages/user/PremAccounts
+    TARGET_DIR ${USER_WORKSPACE_TARGET}
+    DEPENDS ${PremAccounts_js_DEP}
+    PACKAGE_DEPENDS "HttpServer(^${PSIBASE_VERSION}.0)" "Nft(^${PSIBASE_VERSION}.0)" "Tokens(^${PSIBASE_VERSION}.0)" "Sites(^${PSIBASE_VERSION}.0)" "Permissions(^${PSIBASE_VERSION}.0)" "Transact(^${PSIBASE_VERSION}.0)" "AuthSig(^${PSIBASE_VERSION}.0)"
```
James-Mart: package dependencies for `cargo_psibase_packages` are managed in the `Cargo.toml`

packages/shared-ui/lib/schemas/account.ts
```diff
@@ -1,9 +1,12 @@
 import { z } from "zod";
 
+export const MIN_PREMIUM_NAME_LENGTH = 1;
+export const MAX_PREMIUM_NAME_LENGTH = 15;
```
James-Mart: Did you confirm with Steven if this is the max of the new name encoding?

packages/shared-ui/lib/quantity.ts
James-Mart, line 1: If what you're after is a string `"1234.2340"` conversion into a scaled integer representation (`12342340`), you can also use the pre-existing `tokens:plugin/helpers::decimal-to-u64`. That returns a bigint.

packages/user/PremAccounts/README.md
James-Mart, line 1: Should we keep these autogenerated READMEs?

packages/user/PremAccounts/Cargo.toml
```diff
@@ -0,0 +1,27 @@
+[package]
+name = "prem-accounts-pkg"             # Crate name
+description = "An example application"
```
James-Mart: Improve description

packages/user/PremAccounts/plugin/src/lib.rs
```diff
@@ -0,0 +1,377 @@
...
+        if length < MIN_PREMIUM_NAME_LENGTH || length > MAX_PREMIUM_NAME_LENGTH {
+            return Err(ErrorType::CreateMarketLengthInvalid.into());
+        }
+        let packed = prem_accounts::action_structs::enable { length }.packed();
+        add_action_to_transaction("enable", &packed)?;
+        Ok(())
+    }
+
+    fn disable(length: u8) -> Result<(), Error> {
+        assert_authorized_with_whitelist(FunctionName::enable, vec!["config".into()])?;
```
James-Mart: Wrong `FunctionName`

Btw, using the new `psibase_plugin` crate eliminates this kind of issue

packages/user/PremAccounts/service/src/lib.rs
```diff
@@ -0,0 +1,479 @@
...
+    }
+
+    /// Buy a premium account name
+    ///
+    /// # Arguments
+    /// * `account` - The account name to purchase
+    /// * `max_cost` - The maximum cost to pay for the account name (to account for during-call fluctuations)
+    #[action]
+    fn buy(account: String, max_cost: u64) {
+        check_init();
```
James-Mart: You're already using `#[pre_action(exclude(init))]`, so this isn't needed (in any of the actions)

packages/user/PremAccounts/service/src/lib.rs
```diff
@@ -0,0 +1,479 @@
...
+
+        purchased_accounts_table.remove(&purchased_account);
+
+        crate::Wrapper::emit()
+            .history()
+            .premAcctEvent(get_sender(), acct_to_claim, CLAIMED);
+    }
+
+    #[action]
+    fn create(
```
James-Mart, line 198: Missing a check for `getSender() == Wrapper::SERVICE`

Also true in `configure`, `enable`, and `disable`

packages/user/Config/ui/src/hooks/premium-name-markets/use-add-market.ts
```diff
@@ -0,0 +1,28 @@
...
+import QueryKey from "../../lib/query-keys";
+
+import { usePluginMutation } from "@/hooks/use-plugin-mutation";
+
+const PREM_ACCOUNTS = zAccount.parse("prem-accounts");
+
+export const useAddPremiumNameMarket = () =>
+    usePluginMutation<[number, string, number, string, number, number]>(
+        {
+            service: PREM_ACCOUNTS,
```
James-Mart: This is incorrect, the Config UI should not call the prem-accounts plugin directly. It should be calling the `config` plugin, which would propose that the action is called by `prem-accounts` (Same with all other administrative functions)

packages/user/PremAccounts/plugin/src/lib.rs
```diff
@@ -0,0 +1,377 @@
...
+
+        let packed_buy_args = prem_accounts::action_structs::buy {
+            account: account.clone(),
+            max_cost: max_cost_u64,
+        }
+        .packed();
+
+        let result = add_action_to_transaction("buy", &packed_buy_args);
+        if result.is_err() {
+            // TODO: if tx fails, pull back credit; if succeeds, pull back change
```
James-Mart: That's not what result.is_err means. That would mean it fails to add the action to the transaction currently being built. 

Furthermore, if the tx fails, the credit (which is in the tx) would not go through, so nothing to `uncredit`.

packages/user/Tokens/plugin/src/lib.rs
```diff
@@ -121,15 +121,15 @@ impl TokensPlugin {
...
+    #[psibase_plugin::authorized(High, whitelist = ["homepage", "virtual-server", "invite", "token-swap", "prem-accounts"])]
     fn credit(token_id: u32, debitor: String, amount: String, memo: String) -> Result<(), Error> {
         let amount = Self::non_zero(token_id, amount)?;
         let memo = memo.try_into().unwrap();
         Tokens::add_to_tx().credit(token_id, debitor.as_str().into(), amount, memo);
         Ok(())
     }
 
-    #[psibase_plugin::authorized(High, whitelist = ["homepage"])]
+    #[psibase_plugin::authorized(High, whitelist = ["homepage", "prem-accounts"])]
```
James-Mart: Your use of `uncredit` in the prem-accounts plugin isn't needed, so this whitelist isn't either

CMakeLists.txt
```diff
@@ -136,8 +136,8 @@ if(DEFINED WASI_SDK_PREFIX)
 
     set(SERVICES Accounts AuthDelegate AuthSig AuthAny CommonApi
         CpuLimit Db Events EventIndex Explorer Invite
-        Nop Packages Producers HttpServer Sites RAccounts
-        RAuthSig RAuthDelegate REvents RInvite RPackages RProducers
+        Nop Packages PremAccounts Producers HttpServer Sites RAccounts
+        RAuthSig RAuthDelegate REvents RInvite RPackages RPremAccounts RProducers
```
James-Mart: I don't think `RPremAccounts` is a thing

packages/user/PremAccounts/query-service/src/lib.rs
```diff
@@ -0,0 +1,215 @@
...
+        async fn name_events(
+            &self,
+            owner: AccountNumber,
+            first: Option<i32>,
+            last: Option<i32>,
+            before: Option<String>,
+            after: Option<String>,
+        ) -> async_graphql::Result<Connection<u64, PremiumAccountEvent>> {
+            self.require_authenticated()?;
+            query_name_events(self, owner, first, last, before, after)
```
James-Mart: This indirection hides the fact that you're doing double validation
```
self.require_authenticated()?;
q.check_user_auth(owner.clone())?;
```

packages/user/PremAccounts/service/src/lib.rs
```diff
@@ -0,0 +1,479 @@
...
+        Transact::call().runAs(set_owner, vec![]);
+        Transact::call().runAs(set_auth, vec![]);
+    }
+
+    #[action]
+    fn init() {
+        let table = InitTable::new();
+        table.put(&InitRow {}).unwrap();
+        Tokens::Wrapper::call().setUserConf(BalanceFlags::MANUAL_DEBIT.index(), true);
+        Nfts::Wrapper::call().setUserConf(BalanceFlags::MANUAL_DEBIT.index(), true);
```
James-Mart: You're using a Tokens flag on the Nft service here, even though it happens to be the same underlying index.

packages/user/PremAccounts/service/src/lib.rs
```diff
@@ -0,0 +1,479 @@
...
+    ) {
+        check(target > 0, "target must be positive");
+        check(window_seconds > 0, "window seconds must be positive");
+        DiffAdjust::call().set_window(nft_id, window_seconds);
+        DiffAdjust::call().set_targets(nft_id, target, target);
+        DiffAdjust::call().set_floor(nft_id, floor_price);
+        DiffAdjust::call().set_ppm(nft_id, increase_ppm, decrease_ppm);
+    }
+
+    fn new_account(name: AccountNumber) {
```
James-Mart: This entire function can just be 

```
AuthDelegate::Wrapper::call().newAccount(name, get_service());
```

packages/user/PremAccounts/service/src/lib.rs
```diff
@@ -0,0 +1,479 @@
...
+        check_init();
+
+        let length = account.len() as u8;
+        check(
+            length >= MIN_PREMIUM_NAME_LENGTH && length <= MAX_PREMIUM_NAME_LENGTH,
+            &format!(
+                "account name must be {}-{} characters",
+                MIN_PREMIUM_NAME_LENGTH, MAX_PREMIUM_NAME_LENGTH,
+            ),
+        );
```
James-Mart: Already enforced when the account is created

packages/user/PremAccounts/service/src/lib.rs
```diff
@@ -0,0 +1,479 @@
...
+
+        let acct_to_buy = check_some(
+            AccountNumber::from_exact(&account).ok(),
+            "invalid account name: start with a letter; use only lowercase letters, numbers, and hyphens; may not end with '-' or start with 'x-'",
+        );
```
James-Mart: Already enforced when the account is created

packages/user/PremAccounts/service/src/lib.rs
```diff
@@ -0,0 +1,479 @@
...
+
+        check(
+            !Accounts::Wrapper::call().exists(acct_to_buy),
+            "account already exists",
+        );
```
James-Mart: Already enforced when the account is created

packages/user/PremAccounts/service/src/lib.rs
```diff
@@ -0,0 +1,479 @@
...
+    }
+
+    /// Buy a premium account name
+    ///
+    /// # Arguments
+    /// * `account` - The account name to purchase
+    /// * `max_cost` - The maximum cost to pay for the account name (to account for during-call fluctuations)
+    #[action]
+    fn buy(account: String, max_cost: u64) {
```
James-Mart: Is `max_cost` needed in the service? I think the plugin can just credit the `max_cost`, and then this service can debit the current cost, and reject the change.
`max_cost` is just whatever the user put in the shared balance

packages/user/PremAccounts/service/src/lib.rs
```diff
@@ -0,0 +1,479 @@
...
+
+        check(auction.enabled, "market is disabled");
+
+        let current_price = DiffAdjust::call().get_diff(auction.nft_id);
+
+        let shared_bal = Tokens::Wrapper::call().getSharedBal(sys_token_id, sender, get_service());
+
+        check(
+            current_price <= max_cost,
+            "max cost is below the current price; raise max cost to at least the displayed ask",
```
James-Mart: Service shouldn't assume it knows about the UI.
But also this check isn't really needed, already enforced when you debit.

packages/user/PremAccounts/service/src/lib.rs
```diff
@@ -0,0 +1,479 @@
...
+        );
+        check(
+            current_price <= shared_bal.value,
+            "insufficient balance allocated for this purchase; increase max cost so it covers the current price",
```
James-Mart: Not strictly needed because it is enforced when you debit, but I suppose this could still be useful if your goal is to define a nicer error message than what the tokens service provides by default.

packages/user/PremAccounts/service/src/lib.rs
```diff
@@ -0,0 +1,479 @@
...
+                owner: sender,
+            })
+            .unwrap();
+
+        let cost = Quantity::from(current_price);
+        Tokens::Wrapper::call().debit(
+            sys_token_id,
+            sender,
+            cost,
+            "premium account purchase".to_string().try_into().unwrap(),
```
James-Mart: * You can just use `"something".into()` if you're hardcoding the memo
* But also you're already setting an identical memo for this transaction in the plugin on credit, so you should just save some space and leave this debit memo empty (`"".into()`)

packages/user/PremAccounts/service/src/lib.rs
James-Mart: Also `account` can just be an AccountNumber

packages/user/PremAccounts/service/src/lib.rs
```diff
@@ -0,0 +1,479 @@
...
+        // Alert the DiffAdjust service to the purchase
+        DiffAdjust::call().increment(auction.nft_id, 1);
+
+        crate::Wrapper::emit()
+            .history()
+            .premAcctEvent(sender, acct_to_buy, BOUGHT);
+    }
+
+    #[action]
+    fn claim(account: String, pub_key: SubjectPublicKeyInfo) {
```
James-Mart: `account` can be an AccountNumber

packages/user/PremAccounts/service/src/lib.rs
```diff
@@ -0,0 +1,479 @@
...
+        check_init();
+
+        let purchased_accounts_table = PurchasedAccountsTable::new();
+        let acct_to_claim = check_some(
```
James-Mart: Check not needed, it should already come in as an AccountNumber

packages/user/PremAccounts/service/src/lib.rs
```diff
@@ -0,0 +1,479 @@
...
+            purchased_accounts_table.get_index_pk().get(&acct_to_claim),
+            "account not purchased",
+        );
+
+        check(
+            purchased_account.owner == get_sender(),
+            "account not purchased by sender",
+        );
+
+        let set_key_action = Action {
```
James-Mart: Can just be 
`AuthSig::Wrapper::call_as(account).setKey(pub_key);`

packages/user/PremAccounts/service/src/lib.rs
```diff
@@ -0,0 +1,479 @@
...
+            method: AuthSig::action_structs::setKey::ACTION_NAME.into(),
+            rawData: AuthSig::action_structs::setKey {
+                key: pub_key.into(),
+            }
+            .packed()
+            .into(),
+        };
+        Transact::call().runAs(set_key_action, vec![]);
+
+        let set_auth_serv_action = Action {
```
James-Mart: Can just be 
`Accounts::Wrapper::call_as(account).setAuthServ(AuthSig::Wrapper::SERVICE);`

packages/user/PremAccounts/service/src/lib.rs
```diff
@@ -0,0 +1,479 @@
...
+                MIN_PREMIUM_NAME_LENGTH, MAX_PREMIUM_NAME_LENGTH,
+            ),
+        );
+        check(
+            increase_ppm > 0
+                && increase_ppm < 1_000_000
+                && decrease_ppm > 0
+                && decrease_ppm < 1_000_000,
+            "increase and decrease ppm must be between 1 and 999999",
+        );
```
James-Mart: This check is already done by `DiffAdjust::create`

packages/user/PremAccounts/service/src/lib.rs
```diff
@@ -0,0 +1,479 @@
...
+        let auctions_table = AuctionsTable::new();
+        check(
+            auctions_table.get_index_pk().get(&length).is_none(),
+            "market already exists",
+        );
+        check(
+            initial_price >= floor_price,
+            "initial price must be at least floor price",
+        );
```
James-Mart: I'd expect this behavior to be enforced by `DiffAdjust` if it's actually necessary

packages/user/PremAccounts/service/src/lib.rs
```diff
@@ -0,0 +1,479 @@
...
+        check(target > 0, "target must be positive");
```
James-Mart: This check is already done by `DiffAdjust::create`

packages/user/PremAccounts/service/src/lib.rs
```diff
@@ -0,0 +1,479 @@
...
+            MARKET_WINDOW_SECONDS,
+            1,
+            target,
+            floor_price,
+            increase_ppm,
+            decrease_ppm,
+        );
+        Nfts::Wrapper::call().debit(
+            nft_id,
+            "prem accounts rate limit NFT"
```
James-Mart: No one really sees inter-service debit memos, so I'd probably just leave this empty to make the tx a bit cheaper

packages/user/PremAccounts/service/src/lib.rs
```diff
@@ -0,0 +1,479 @@
...
+    fn apply_diff_adjust_configure(
+        nft_id: u32,
+        window_seconds: u32,
+        target: u32,
+        floor_price: u64,
+        increase_ppm: u32,
+        decrease_ppm: u32,
+    ) {
+        check(target > 0, "target must be positive");
+        check(window_seconds > 0, "window seconds must be positive");
```
James-Mart: Both of the above checks I would expect to be enforced by `DiffAdjust`

packages/user/PremAccounts/plugin/src/lib.rs
```diff
@@ -0,0 +1,377 @@
...
+            None => return Err(ErrorType::SystemTokenNotDefined.into()),
+        };
+        let floor_price = floor_price.trim().to_string();
+        if floor_price.is_empty() {
+            return Err(ErrorType::MaxCostNotCanonicalDecimal.into());
+        }
+        let floor_price_u64 = match TokensHelpers::decimal_to_u64(sys_token_id, &floor_price) {
+            Ok(v) => v,
+            Err(_) => return Err(ErrorType::MaxCostNotCanonicalDecimal.into()),
+        };
```
James-Mart: In various places in this plugin you do a whole bunch of work converting string token amounts into a u64, but I don't really see what it buys you.

Can't you just do
`let floor_price: u64 = TokensHelpers::decimal_to_u64(sys_token_id, &floor_price)?;`

packages/user/PremAccounts/plugin/src/lib.rs
```diff
@@ -0,0 +1,377 @@
...
+        Ok(())
+    }
+
+    fn claim(account: String) -> Result<String, Error> {
+        assert_authorized_with_whitelist(FunctionName::claim, vec!["accounts".into()])?;
+
+        let account = account.trim().to_string();
+        validate_premium_account_name(&account)?;
+
+        let keypair = AuthSigKeyVault::generate_unmanaged_keypair()?;
```
James-Mart: I think we should probably prefer
```
let keypair = host::crypto::generate_unmanaged_keypair().unwrap();
```

packages/user/PremAccounts/plugin/src/lib.rs
```diff
@@ -0,0 +1,377 @@
...
+
+        if pem_data.tag() != "PUBLIC KEY" {
+            return Err(ErrorType::InvalidPublicKey(format!(
+                "Expected PUBLIC KEY, got {}",
+                pem_data.tag()
+            ))
+            .into());
+        }
+
+        AuthSigKeyVault::import_key(&keypair.private_key)?;
```
James-Mart: You can use
```
let pubkey = host::crypto::import_key(&keypair.private_key).unwrap();
```

packages/user/PremAccounts/plugin/src/lib.rs
```diff
@@ -0,0 +1,377 @@
...
+        let pem_data = pem::parse(keypair.public_key.trim())
+            .map_err(|e| ErrorType::InvalidPublicKey(format!("Failed to parse PEM: {}", e)))?;
+
+        if pem_data.tag() != "PUBLIC KEY" {
+            return Err(ErrorType::InvalidPublicKey(format!(
+                "Expected PUBLIC KEY, got {}",
+                pem_data.tag()
+            ))
+            .into());
+        }
```
James-Mart: Using `pem::parse` and checking the tag is not needed

packages/user/PremAccounts/plugin/src/lib.rs
```diff
@@ -0,0 +1,377 @@
...
+
+        AuthSigKeyVault::import_key(&keypair.private_key)?;
+
+        let packed_claim_args = prem_accounts::action_structs::claim {
+            account: account.clone(),
+            pub_key: SubjectPublicKeyInfo::from(pem_data.contents().to_vec()),
```
James-Mart: Can be
`pub_key: host::crypto::keyvault::to_der(&pubkey).unwrap().into()`

packages/user/PremAccounts/plugin/src/lib.rs
```diff
@@ -0,0 +1,377 @@
...
+    };
+
+    let price_str = row.price.to_string();
+    TokensHelpers::decimal_to_u64(sys_token_id, &price_str)
+        .map_err(|_| ErrorType::QueryResponseParseError("invalid price".into()).into())
+}
+
+impl Queries for PremAccountsPlugin {
+    // TODO: Should this query also be an action for srv-to-srv calls?
+    fn get_prices() -> Result<Vec<MarketPrice>, Error> {
```
James-Mart: I don't necessarily think this needs to be a plugin function. Probably should prefer a `prem-accounts:plugin/authenticated::graphql`  interface, similar to tokens, that allows the UI to make authenticated queries.

packages/shared-ui/lib/schemas/account.ts
cool-ant: It's not. The max of the new encoding is 10, but that's a future PR when the encoding takes effect. Otherwise, there will be numerous things that need to change today. I can do them pre-emptively if you like

packages/shared-ui/lib/quantity.ts
cool-ant: ok, I'll do all the Quantity stuff in a follow-up PR

packages/shared-ui/lib/schemas/account.ts
cool-ant: I propose name length change as a consequence of  the new name algo be a follow-on PR

packages/user/PremAccounts/README.md
cool-ant: I had the same question. Probably useful but probably requires some discussion

packages/user/PremAccounts/service/src/lib.rs
cool-ant: derp. good catch; thanks

packages/user/Config/ui/src/pages/premium-name-market-config.tsx
```diff
@@ -0,0 +1,683 @@
...
+    const increasePpm = parsePpm(increasePpmRaw);
+    const decreasePpm = parsePpm(decreasePpmRaw);
+    const pricesOk =
+        floorPrice.trim() !== "" &&
+        target !== null &&
+        increasePpm !== null &&
+        decreasePpm !== null;
+
+    return (
+        <div className="flex max-w-lg flex-col gap-4 pt-2">
```
brandonfancher: Recommend using the shared `<PageContainer />` component here, as is used on other Config pages, for consistency.

packages/user/Config/ui/src/hooks/premium-name-markets/use-configured-markets.ts
```diff
@@ -0,0 +1,49 @@
+import { useQuery } from "@tanstack/react-query";
+import { z } from "zod";
+
+import { graphql } from "@shared/lib/graphql";
+import QueryKey from "../../lib/query-keys";
```
brandonfancher, line 10: Make sure prettier formatting is working, as these imports are out of order. (And double check other files in this PR are formatted properly.)

packages/user/Config/ui/src/pages/premium-name-market-config.tsx
brandonfancher, line 1: This one is really long. Recommend breaking up into a few components.

packages/user/PremAccounts/ui/src/pages/purchased-page.tsx
```diff
@@ -0,0 +1,196 @@
...
+                    )}
+                    {error && <p className="text-red-500">{error}</p>}
+                </div>
+                {claimSuccessNotification && (
+                    <div className="mt-4 rounded-lg border border-green-200 bg-green-50 p-4 dark:border-green-900 dark:bg-green-950/50">
+                        <p className="mb-2 font-medium text-green-700 dark:text-green-300">
+                            Account &apos;{claimSuccessNotification.accountName}
+                            &apos; claimed successfully
+                        </p>
+                        <p className="text-muted-foreground mb-1 text-sm">
```
brandonfancher: For now, might be best to format this the same way we format out passwords. We drop the "(BEGIN|END) PRIVATE KEY" lines and remove line breaks (without adding spaces). Right now, we can copy the body of the private key, but we have to manually remove the spaces inserted before it can be used as the password for sign in.

brandonfancher (approved):
Approved pending fix for the button that creates markets for names with lengths 1-15 instead of 1-9. And see other, more minor comments.

packages/user/PremAccounts/service/src/lib.rs
cool-ant: I suppose this is a matter of how much do we want the service to satnd on its own? given we do this validation in the UI. Perhaps we let people trying to use the plugin/service wtihout the prescribed UI get a slightly uglier message? I'd be fine with that.

packages/system/Accounts/src/Accounts.cpp
```diff
@@ -80,7 +81,12 @@ namespace SystemService
       {
          check(!strName.starts_with("x-"),
                "The 'x-' account prefix is reserved for infrastructure providers");
-         check(strName.length() >= 10, "account name must be at least 10 characters: " + strName);
+         // Allow PremAccounts service to create accounts with < 10 characters
+         if (getSender() != AccountNumber("prem-accounts"))
```
James-Mart: Actually I think this change is not needed, the code that was there before was already correct (no custom `prem-accounts` logic is needed in this action, only in preapprove)

packages/user/PremAccounts/README.md
James-Mart: I guess the text in the README is really meant for the developer of the new package more than it's meant to be documentation for third parties about the generated package. 

Which makes me think the info should be moved to `/package-templates/README.md` and removed from the generated package.

packages/user/PremAccounts/service/src/lib.rs
Velua: Yeah lets remove this and just debit, I haven't seen this pattern anywhere else and don't see a need to introduce it now.

packages/user/PremAccounts/service/src/lib.rs
```diff
@@ -0,0 +1,378 @@
...
+                floor_price: 0,
+                window_seconds: 0,
+                increase_ppm: 0,
+                decrease_ppm: 0,
+            }
+        }
+    }
+
+    impl Auction {
+        #[primary_key]
```
Velua: Can just move this macro to the struct definition, delete this `impl` block entirely.

packages/user/PremAccounts/service/src/lib.rs
```diff
@@ -0,0 +1,378 @@
...
+    pub struct PurchasedAccount {
+        pub account: AccountNumber,
+        pub owner: AccountNumber,
+    }
+
+    impl PurchasedAccount {
+        #[primary_key]
+        fn pk(&self) -> AccountNumber {
+            self.account
+        }
```
Velua: Same thing here

packages/user/PremAccounts/service/src/lib.rs
```diff
@@ -0,0 +1,378 @@
...
+        decrease_ppm: u32,
+    ) {
+        DiffAdjust::call().set_window(nft_id, window_seconds);
+        DiffAdjust::call().set_targets(nft_id, target, target);
+        DiffAdjust::call().set_floor(nft_id, floor_price);
+        DiffAdjust::call().set_ppm(nft_id, increase_ppm, decrease_ppm);
+    }
+
+    fn require_caller_is_self() -> bool {
+        return get_sender() == Wrapper::SERVICE;
```
Velua: Would there be a difference is this was...? @James-Mart 
```suggestion
        return get_sender() == get_service();
```

packages/user/PremAccounts/service/src/lib.rs
```diff
@@ -0,0 +1,378 @@
...
+    }
+
+    #[action]
+    fn init() {
+        let table = InitTable::new();
+        table.put(&InitRow {}).unwrap();
+        Tokens::Wrapper::call().setUserConf(BalanceFlags::MANUAL_DEBIT.index(), true);
+        Nfts::Wrapper::call().setUserConf(NftHolderFlags::MANUAL_DEBIT.index(), true);
+
+        register_prem_acct_event_indices();
```
Velua, line 116: From what I understand, this block can be called over and over each time someone calls `init`. 
I think it's worth wrapping in something like...  
```
if InitTable::read().get_index_pk().get(&()).is_none() {
    table.put(&InitRow {}).unwrap();
    ...
}
```

which would give us more confidence in idempotence

packages/user/PremAccounts/service/src/lib.rs
```diff
@@ -0,0 +1,378 @@
...
+        let sender = get_sender();
+
+        let auction = check_some(
+            AuctionsTable::new().get_index_pk().get(&length),
+            "auction not found for this length",
+        );
+
+        check(auction.enabled, "market is disabled");
+
+        let current_price = DiffAdjust::call().get_diff(auction.nft_id);
```
Velua: `DiffAdjust` is designed to be called with `increment` immediately in this case as it returns the new difficulty as u64. Otherwise this isn't consistent with the `Symbol` approach.

Velua:
```suggestion
        let current_price = Quantity::from(DiffAdjust::call().increment(auction.nft_id, 1));
```

packages/user/PremAccounts/service/src/lib.rs
```diff
@@ -0,0 +1,378 @@
...
+        AuthSig::Wrapper::call_as(account).setKey(pub_key);
+
+        Accounts::Wrapper::call_as(account).setAuthServ(AuthSig::Wrapper::SERVICE);
+
+        purchased_accounts_table.remove(&purchased_account);
+
+        crate::Wrapper::emit()
+            .history()
+            .premAcctEvent(get_sender(), account.clone(), CLAIMED);
+    }
```
Velua, line 195: I think this smells. 

1. Recording a `PurchasedAccount` against an `AccountNumber` creates way more friction for ongoing speculation of the asset when it could just be an `NFT` instead. Such an NFT minted by this service sells a "brand new account still in it's packaging" appeal which is lost if we hard code it to a single owner who's only choice is to eject. 
2. The ejection of an account going from it's sealed packaging immediately into `AuthSig` with a public key feels overly opinionated when we could instead just do `AuthDelegate::setOwner(get_sender())` and then let the unwrapping owner do whatever he wants, AuthSig makes it awkward especially considering that it could still be a service hitting this `claim` action too.

packages/user/PremAccounts/service/src/lib.rs
```diff
@@ -0,0 +1,378 @@
...
+            floor_price,
+            increase_ppm,
+            decrease_ppm,
+        );
+        auction.target = target;
+        auction.floor_price = floor_price;
+        auction.window_seconds = window_seconds;
+        auction.increase_ppm = increase_ppm;
+        auction.decrease_ppm = decrease_ppm;
+        auctions_table.put(&auction).unwrap();
```
Velua: Why do we bother to store these properties on this service?

packages/user/PremAccounts/service/src/lib.rs
```diff
@@ -0,0 +1,378 @@
...
+            ),
+        );
+        let auctions_table = AuctionsTable::new();
+        let mut auction = check_some(
+            auctions_table.get_index_pk().get(&length),
+            "auction not found for this length",
+        );
+        auction.enabled = false;
+        auctions_table.put(&auction).unwrap();
+    }
```
Velua: I'd be tempted to just have one action, `toggle_auction` with a `is_active: bool` param, save all this repetition.

packages/user/PremAccounts/ui/src/lib/utils.ts
```diff
@@ -0,0 +1,37 @@
...
+    const parsObject = new URLSearchParams(urlObject.search);
+
+    Object.entries(params).forEach(([key, value]) => {
+        parsObject.set(key, value);
+    });
+
+    urlObject.search = parsObject.toString();
+
+    return urlObject.toString();
+};
```
Velua: This appears to be re-introducing things we've either moved to shared or deleted entirely. 
```suggestion

```

packages/shared-ui/lib/quantity.ts
```diff
@@ -230,3 +236,83 @@ export class Quantity {
...
+        frac.length === precision
+    );
+}
+
+export function formatCanonicalTokenAmount(
+    canonicalDecimal: string,
+    symbol?: string,
+): string {
+    return symbol ? `${canonicalDecimal} ${symbol}` : canonicalDecimal;
+}
```
Velua: This is a repeat of `Quantity.format({ includeLabel: true })` isn't it?

packages/user/PremAccounts/ui/src/pages/buy-page.tsx
```diff
@@ -0,0 +1,399 @@
...
+                                        Account name already exists
+                                    </p>
+                                )}
+                                {accountExists === false && price !== null && (
+                                    <p className="text-green-600">
+                                        Buy for{" "}
+                                        {formatCanonicalTokenAmount(
+                                            price,
+                                            systemToken?.symbol,
+                                        )}
```
Velua: This is a hardcore function name for what it's achieving and doubt it's value in a shared library, I'd argue for it's deletion. 

```suggestion
                                       {price} {systemToken?.symbol ?? ''}
```

packages/user/PremAccounts/ui/src/pages/buy-page.tsx
```diff
@@ -0,0 +1,399 @@
...
+                    query: `
+                        query {
+                            token(tokenId: "${sysTid}") {
+                                precision
+                                symbol
+                            }
+                        }
+                    `,
+                }),
+            });
```
Velua: Stuff like this is already done in `packages/shared-ui/lib/graphql.ts`

packages/user/PremAccounts/ui/src/pages/buy-page.tsx
```diff
@@ -0,0 +1,399 @@
...
+            const tokenData = (await tokenResponse.json()) as {
+                data?: { token?: { precision?: number; symbol?: string } };
+                errors?: Array<{ message: string }>;
+            };
```
Velua: Types like this should be asserted with Zod, it's much cleaner and we can trust it's actually legit during runtime. 
See `packages/user/Homepage/ui/src/apps/token-swap/hooks/use-token.ts` for what a clean query + typing looks like.

packages/user/PremAccounts/ui/src/pages/buy-page.tsx
```diff
@@ -0,0 +1,399 @@
...
+    }, []);
+
+    useEffect(() => {
+        const checkAccount = async () => {
+            const trimmed = accountName.trim();
+            if (!trimmed) {
+                setAccountExists(null);
+                setPrice(null);
+                return;
+            }
```
Velua: This is somewhat redundant given it would be caught and treated the same way below with the `zAccount.safeParse` right? 
```suggestion
```

packages/user/PremAccounts/plugin/src/lib.rs
```diff
@@ -0,0 +1,270 @@
...
+            high: &["Purchase premium account names"],
+        }
+    }
+}
+
+struct PremAccountsPlugin;
+
+impl MarketAdmin for PremAccountsPlugin {
+    #[psibase_plugin::authorized(Max, whitelist = ["config"])]
+    fn create(
```
Velua, line 32:
```suggestion
    fn create(
        length: u8,
        initial_price: String,
        target: u32,
        floor_price: String,
        increase_ppm: u32,
        decrease_ppm: u32,
    ) -> Result<(), Error> {
        let sys_token_id =
            TokensHelpers::fetch_network_token()?.ok_or(ErrorType::SystemTokenNotDefined)?;

        let initial_price_u64 = TokensHelpers::decimal_to_u64(sys_token_id, &initial_price)?;
        let floor_price_u64 = TokensHelpers::decimal_to_u64(sys_token_id, &floor_price)?;

        let packed = prem_accounts::action_structs::create {
            length,
            initial_price: initial_price_u64,
            target,
            floor_price: floor_price_u64,
            increase_ppm,
            decrease_ppm,
        }
        .packed();

        add_action_to_transaction("create", &packed)
    }
```

packages/user/PremAccounts/service/src/lib.rs
```diff
@@ -0,0 +1,378 @@
...
+
+        crate::Wrapper::emit()
+            .history()
+            .premAcctEvent(get_sender(), account.clone(), CLAIMED);
+    }
+
+    #[action]
+    fn create(
+        length: u8,
+        initial_price: u64,
```
Velua: Prices / costs should be represented in `Quantity` for uniformity with other services.

packages/user/PremAccounts/ui/src/components/history-section.tsx
```diff
@@ -0,0 +1,183 @@
...
+                    null,
+                    PREM_ACCOUNTS_SERVICE,
+                    "/graphql",
+                );
+
+                const allEvents: { account: string; action: string }[] = [];
+                let hasNextPage = true;
+                let afterCursor: string | null = null;
+
+                while (hasNextPage) {
```
Velua, line 42: Are we ignoring pagination on the UI here and just downloading everything?

packages/shared-ui/lib/schemas/account.ts
```diff
@@ -1,9 +1,16 @@
 import { z } from "zod";
 
+export const MIN_ACCOUNT_NAME_LENGTH = 1;
+export const MAX_ACCOUNT_NAME_LENGTH = 10;
```
Velua: Why the change from `18` to `10` ?

packages/shared-ui/lib/quantity.ts
```diff
@@ -230,3 +236,83 @@ export class Quantity {
...
+        return null;
+    }
+    const intPart = m[1];
+    let frac = m[2] ?? "";
+    if (frac.length > precision) {
+        return null;
+    }
+    frac = frac.padEnd(precision, "0");
+    return `${intPart}.${frac}`;
+}
```
Velua: We've already got the Quantity class + the format method for this don't we? 
And we rarely need it as we can just pass stuff like `1.2` directly to the plugin as a string which will then internally parse it with `decimal_to_u64`.

Velua (changes requested):

packages/user/PremAccounts/service/src/lib.rs
James-Mart: Using `get_service` is generic, allows the service to be moved. But a lot of our services hardcode the expected service, so not particularly important.

packages/user/PremAccounts/service/src/lib.rs
James-Mart: Or just
```
if table.get_index_pk().get(&()).is_some() {
    return;
}
```

As written it is technically already idempotent. But this just makes it an explicit no-op on re-init

packages/user/PremAccounts/query-service/Cargo.toml
```diff
@@ -0,0 +1,18 @@
...
+publish.workspace = true
+
+[dependencies]
+psibase.workspace = true
+psibase_service.workspace = true
+serde.workspace = true
+serde_json.workspace = true
+serde-aux.workspace = true
+prem-accounts = { path = "../service", version = "0.23.0" }
+diff-adjust = { path = "../../DiffAdjust/service", version = "0.23.0" }
```
James-Mart: You should link to `DiffAdjust` via your existing psibase dep (psibase/services/...) not via a direct crate dependency

packages/user/PremAccounts/service/src/lib.rs
swatanabe: I disagree somewhat. The main way to call `init` again is to upgrade a package. In this case, any new parts of `init` do need to be run. As long as everything that `init` does is idempotent, it's better just to run it all again.

packages/user/PremAccounts/plugin/src/lib.rs
```diff
@@ -0,0 +1,270 @@
...
+        let packed = prem_accounts::action_structs::create {
+            length,
+            initial_price: initial_price_u64,
+            target,
+            floor_price: floor_price_u64,
+            increase_ppm,
+            decrease_ppm,
+        }
+        .packed();
+        add_action_to_transaction("create", &packed)?;
```
James-Mart: FYI, since you added `psibase_plugin`, you can also use another simplification

`prem_accounts::Wrapper::add_to_tx().create(...);`

No separate action struct packing :)

packages/user/PremAccounts/plugin/src/lib.rs
```diff
@@ -0,0 +1,270 @@
...
+        .into());
+    };
+
+    let price_str = row.price.to_string();
+    TokensHelpers::decimal_to_u64(sys_token_id, &price_str)
+        .map_err(|_| ErrorType::QueryResponseParseError("invalid price".into()).into())
+}
+
+impl Authorized for PremAccountsPlugin {
+    #[psibase_plugin::authorized(Medium, whitelist = ["accounts", "prem-accounts"])]
```
James-Mart: Don't need to whitelist yourself (implicitly whitelisted)

packages/user/PremAccounts/service/src/lib.rs
James-Mart: I don't think (1) particularly matters. It's a reasonable internal change that can be done later.

I do agree with (2). It would be nice if a claimed account used `AuthDelegate`, as it would make it easier to use for services. I think we should do this - Although there's already a planned fast-follow PR (or 2), so I think we can do this refactor at that point.

packages/shared-ui/lib/schemas/account.ts
cool-ant: new name algo

packages/user/PremAccounts/service/src/lib.rs
Velua: If we did do 2 then I think without 1 you may as well drop the PurchasedAccountTable entirely and just immediately delegate the account to the sender during purchase.

James-Mart: No, once you've claimed it, the account is not available for resale because we can't verify that the account hasn't been tampered with.

packages/shared-ui/lib/schemas/account.ts
Velua: Does that get implemented in this PR though? The change you're making here is across all UI's
Unless this PR makes it so that I can no longer create an account of 11 chars on I'd say this change should be added elsewhere.

Reply: I've reversed this. We originally discussed making the account length changes in this PR, but as that's become more questioned and clearly a separate issue that can just as easily be a follow-on PR, I've reversed my changes to the chain's account upper limit on length. Steven (or one of us) will make those changes when his naming algo changes drop,

packages/user/PremAccounts/service/src/lib.rs
Velua: There's no resale anyway is there? 
We're hard-coding ownership to the original purchaser with no transfer abilities prior to unwrapping so one must lose tamper guarantees in order to transfer / resell it, unless you've made it part of a greater staged tx / service deployed by the owner or something.

packages/user/PremAccounts/service/src/lib.rs
cool-ant: great call.

packages/shared-ui/lib/quantity.ts
cool-ant: This version isn't lossy (Quantity.format(...) parses string as Number(), which can by lossy for large numbers. So once now that we have these proper conversions in shared, we can tweak and share them. This'll take a separate PR to flesh these out further.

packages/user/PremAccounts/ui/src/pages/buy-page.tsx
cool-ant: I'll simplify it to current standards (to the degree there are such), but I want to push finalizing/polishing this file until a future PR dedicated to it

packages/shared-ui/lib/quantity.ts
cool-ant: Current implementation rlies on Javascript's Number type, which can cause trouble in some cases (not big enough for Quantity; should be BigInt), so these functions handle quantities with precision

packages/user/PremAccounts/service/src/lib.rs
cool-ant: wow. great catch.

packages/user/PremAccounts/ui/src/components/history-section.tsx
cool-ant: lol. put them in the query; forgot the UI

packages/shared-ui/lib/quantity.ts
cool-ant: Number vs BigInt

packages/shared-ui/lib/quantity.ts
```diff
@@ -230,3 +236,83 @@ export class Quantity {
...
+    precision: number,
+): string | null {
+    const t = amount.trim();
+    if (t.length === 0) {
+        return null;
+    }
+    if (precision <= 0) {
+        return /^\d+$/.test(t) ? t : null;
+    }
+    const m = /^(\d+)(?:\.(\d*))?$/.exec(t);
```
swatanabe: This allows `9.`, but not `.125`. Is that intended?

packages/shared-ui/lib/quantity.ts
```diff
@@ -230,3 +236,83 @@ export class Quantity {
...
+    }
+    const m = /^(\d+)\.(\d+)$/.exec(t);
+    if (!m) {
+        return false;
+    }
+    const [, intPart, frac] = m;
+    return (
+        intPart.length > 0 &&
+        /^\d+$/.test(intPart) &&
+        /^\d+$/.test(frac) &&
```
swatanabe: Don't these tests duplicate the regex that creates `m`?

packages/user/PremAccounts/service/src/lib.rs
```diff
@@ -0,0 +1,327 @@
...
+        decrease_ppm: u32,
+    ) {
+        DiffAdjust::call().set_window(nft_id, window_seconds);
+        DiffAdjust::call().set_targets(nft_id, target, target);
+        DiffAdjust::call().set_floor(nft_id, floor_price.value);
+        DiffAdjust::call().set_ppm(nft_id, increase_ppm, decrease_ppm);
+    }
+
+    fn require_caller_is_self() -> bool {
+        return get_sender() == get_service();
```
James-Mart: Security bug: This doesn't assert, just returns a bool, which is promptly ignored

packages/user/PremAccounts/service/src/lib.rs
```diff
@@ -0,0 +1,327 @@
...
+        );
+        let auctions_table = AuctionsTable::new();
+        check(
+            auctions_table.get_index_pk().get(&length).is_none(),
+            "market already exists",
+        );
+        let nft_id = DiffAdjust::call().create(
+            initial_price.value,
+            MARKET_WINDOW_SECONDS,
+            1,
```
James-Mart: Wait, is `target_min == 1` what we want? That means that the price only decreases if there are 0 sales. I think `target_min == target_max == target` makes more sense

packages/user/PremAccounts/plugin/src/lib.rs
```diff
@@ -0,0 +1,225 @@
...
+        };
+
+        let max_cost = max_cost.trim().to_string();
+        let max_cost_u64 = TokensHelpers::decimal_to_u64(sys_token_id, &max_cost)?;
+        let canonical = TokensHelpers::u64_to_decimal(sys_token_id, max_cost_u64)?;
+        if canonical != max_cost {
+            return Err(ErrorType::MaxCostNotCanonicalDecimal.into());
+        }
+
+        let length = account.len() as u8;
```
James-Mart, line 102: Should be `acct_name.to_string().len()`, which could be different from `account.len()` because of `account.trim()`

packages/user/PremAccounts/service/src/lib.rs
James-Mart: @cool-ant you thumbed-up steven's comment, which usually means you completed it, but your code still has the early return:
```
if table.get_index_pk().get(&()).is_some() {
    return;
}
```

packages/user/PremAccounts/service/src/lib.rs
James-Mart: John's suggestion was to use `toggle_auction` as the action directly and delete `enable` and `disable` actions.

But anyway, as is at least the main redundancy was removed, now it's just a slightly overly verbose action/plugin interface.

packages/user/Config/ui/src/hooks/premium-name-markets/use-configured-markets.ts
```diff
@@ -0,0 +1,52 @@
...
+import { graphql } from "@shared/lib/graphql";
+
+import QueryKey from "../../lib/query-keys";
+
+const zRow = z.object({
+    length: z
+        .number()
+        .int()
+        .min(DEFAULT_MIN_PREMIUM_NAME_LENGTH_MARKET)
+        .max(DEFAULT_MAX_PREMIUM_NAME_LENGTH_MARKET),
```
James-Mart: This means you expect all premium markets that are fetched to be between 1 and 7 characters in length, but I don't think anything in the create-market flow restricts you from creating premium markets up to the max allowed account name length.

packages/user/PremAccounts/service/src/lib.rs
James-Mart: Also in the `market_params` query I'm noticing that you also assume `target_min == target_max == target`
```
target: rate_limit.target_min, // target_min == target_max in our usage
```

packages/shared-ui/lib/quantity.ts
cool-ant: Rather than fix these, I found the regex largely just makes this harder to read for many (especially with the further tweaks that were needed), so I went with a much more readable split() operation.
Good catch. Thanks

packages/user/PremAccounts/service/src/lib.rs
cool-ant: derp

packages/user/Config/ui/src/hooks/premium-name-markets/use-configured-markets.ts
cool-ant: I found `toggle` with a bool param to be semantically awkward. toggle toggles (flips a value). `enable_market(enable: bool)` is also a bit awkward, so I opted for the more explicit api you and I originally discussed.
@James-Mart @Velua  aligned with that still?

brandonfancher (approved):

packages/user/Config/ui/src/hooks/premium-name-markets/use-configured-markets.ts
James-Mart: We do have other services that use `enable_x(enabled: bool)`
...but the explicit interface is totally fine with me

James-Mart: Also this comment is on the wrong chain :P

packages/user/PremAccounts/plugin/src/lib.rs
```diff
@@ -0,0 +1,225 @@
...
+    fn create(
+        length: u8,
+        initial_price: String,
+        target: u32,
+        floor_price: String,
+        increase_ppm: u32,
+        decrease_ppm: u32,
+    ) -> Result<(), Error> {
+        let sys_token_id = TokensHelpers::fetch_network_token()?
+            // .map_err(|_| ErrorType::SystemTokenNotDefined)?
```
Velua:
```suggestion
```

packages/user/PremAccounts/plugin/src/lib.rs
```diff
@@ -0,0 +1,225 @@
...
+        window_seconds: u32,
+        target: u32,
+        floor_price: String,
+        increase_ppm: u32,
+        decrease_ppm: u32,
+    ) -> Result<(), Error> {
+        let sys_token_id = match TokensHelpers::fetch_network_token()? {
+            Some(id) => id,
+            None => return Err(ErrorType::SystemTokenNotDefined.into()),
+        };
```
Velua:
```suggestion
        let sys_token_id =
            TokensHelpers::fetch_network_token()?.ok_or(ErrorType::SystemTokenNotDefined)?;
```

packages/user/PremAccounts/plugin/src/lib.rs
```diff
@@ -0,0 +1,225 @@
...
+    #[psibase_plugin::authorized(High, whitelist = ["accounts"])]
+    fn buy(account: String, max_cost: String) -> Result<(), Error> {
+        let acct_name = AccountNumber::from_exact(account.trim()).unwrap();
+
+        let service_account = prem_accounts::SERVICE.to_string();
+
+        let sys_token_id = match TokensHelpers::fetch_network_token()? {
+            Some(id) => id,
+            None => return Err(ErrorType::SystemTokenNotDefined.into()),
+        };
```
Velua:
```suggestion
        let sys_token_id =
            TokensHelpers::fetch_network_token()?.ok_or(ErrorType::SystemTokenNotDefined)?;
```

packages/user/PremAccounts/plugin/src/lib.rs
```diff
@@ -0,0 +1,225 @@
...
+    fn disable(length: u8) -> Result<(), Error> {
+        prem_accounts::Wrapper::add_to_tx().disable(length);
+        Ok(())
+    }
+}
+
+impl Api for PremAccountsPlugin {
+    #[psibase_plugin::authorized(High, whitelist = ["accounts"])]
+    fn buy(account: String, max_cost: String) -> Result<(), Error> {
+        let acct_name = AccountNumber::from_exact(account.trim()).unwrap();
```
Velua: I don't like the use of `trim` here, if we get white space from the UI or another plugin function then we should recognize it as bad input and treat it as a bug from the caller rather than hide it and silently fix it.

packages/user/PremAccounts/plugin/src/lib.rs
```diff
@@ -0,0 +1,225 @@
...
+
+        let service_account = prem_accounts::SERVICE.to_string();
+
+        let sys_token_id = match TokensHelpers::fetch_network_token()? {
+            Some(id) => id,
+            None => return Err(ErrorType::SystemTokenNotDefined.into()),
+        };
+
+        let max_cost = max_cost.trim().to_string();
```
Velua: Same deal here, caller either gives us a decent string or nothing at all, we shouldn't silently fix, better to throw and force the caller to get it together.

packages/user/PremAccounts/plugin/src/lib.rs
```diff
@@ -0,0 +1,225 @@
...
+        let max_cost_u64 = TokensHelpers::decimal_to_u64(sys_token_id, &max_cost)?;
+        let canonical = TokensHelpers::u64_to_decimal(sys_token_id, max_cost_u64)?;
+        if canonical != max_cost {
+            return Err(ErrorType::MaxCostNotCanonicalDecimal.into());
+        }
```
Velua: This check seems overly defensive and strange we do it here but not in the other several places this same operation is performed.

packages/user/PremAccounts/plugin/src/lib.rs
```diff
@@ -0,0 +1,225 @@
...
+        prem_accounts::Wrapper::add_to_tx().buy(acct_name);
+
+        Ok(())
+    }
+
+    #[psibase_plugin::authorized(Medium, whitelist = ["accounts"])]
+    fn claim(account: String) -> Result<String, Error> {
+        let account = AccountNumber::from_exact(account.trim()).unwrap();
+
+        let keypair = AuthSigKeyVault::generate_unmanaged_keypair().unwrap();
```
Velua: This returns a Result with a compatible error, so no need to panic when we can return an error if there is one. 
```suggestion
let keypair = AuthSigKeyVault::generate_unmanaged_keypair()?;
```

packages/user/PremAccounts/plugin/src/lib.rs
```diff
@@ -0,0 +1,225 @@
...
+
+        AuthSigKeyVault::import_key(&keypair.private_key).unwrap();
```
Velua:
```suggestion
AuthSigKeyVault::import_key(&keypair.private_key)?;
```

packages/user/PremAccounts/plugin/src/lib.rs
```diff
@@ -0,0 +1,225 @@
...
+
+        prem_accounts::Wrapper::add_to_tx().claim(
+            account,
+            SubjectPublicKeyInfo::from(HostCryptoKeyVault::to_der(&keypair.public_key).unwrap()),
```
Velua:
```suggestion
SubjectPublicKeyInfo::from(HostCryptoKeyVault::to_der(&keypair.public_key)?),
```

packages/user/PremAccounts/query-service/src/lib.rs
```diff
@@ -0,0 +1,206 @@
...
+        }
+
+        fn require_authenticated(&self) -> async_graphql::Result<AccountNumber> {
+            if self.user.is_none() {
+                return Err(async_graphql::Error::new(
+                    "permission denied: an authorized session is required for this query.",
+                ));
+            }
+            Ok(self.user.unwrap().clone())
+        }
```
Velua, line 82:
```suggestion
        fn require_authenticated(&self) -> async_graphql::Result<AccountNumber> {
            self.user.ok_or_else(|| {
                async_graphql::Error::new(
                    "permission denied: an authorized session is required for this query.",
                )
            })
        }
```

packages/user/PremAccounts/query-service/src/lib.rs
```diff
@@ -0,0 +1,206 @@
...
+            user.append_key(&mut prefix);
+            let sub = TableIndex::<AccountNumber, PurchasedAccount>::new(
+                &owner_idx.db,
+                prefix,
+                owner_idx.is_secondary,
+            );
+            let mut names: Vec<String> = sub.iter().map(|r| r.account.to_string()).collect();
+            names.sort();
+            Ok(names)
+        }
```
Velua:
```suggestion
        async fn unclaimed_names(
            &self,
            first: Option<i32>,
            last: Option<i32>,
            before: Option<String>,
            after: Option<String>,
        ) -> async_graphql::Result<Connection<RawKey, PurchasedAccount>> {
            let user = self.require_authenticated()?;

            TableQuery::subindex::<AccountNumber>(
                PurchasedAccountsTable::with_service(prem_accounts::SERVICE).get_index_by_owner(),
                &(user),
            )
            .first(first)
            .last(last)
            .before(before)
            .after(after)
            .query()
            .await
        }
```

packages/user/PremAccounts/query-service/src/lib.rs
```diff
@@ -0,0 +1,206 @@
...
+                        price_raw = rate_table
+                            .get_index_pk()
+                            .get(&auction.nft_id)
+                            .map(|r| r.floor_difficulty)
+                            .unwrap_or(0);
+                    }
+                    MarketPrice {
+                        length: auction.length,
+                        price: Decimal::new(Quantity::from(price_raw), precision),
+                    }
```
Velua, line 114: I don't see a filter for auctions where enabled = false, therefore sending back a price for something that's not on the market feels misleading.

packages/user/PremAccounts/query-service/src/lib.rs
```diff
@@ -0,0 +1,206 @@
...
+                    // Delegate effective difficulty to DiffAdjust (same as `get_diff` action).
+                    let mut price_raw = DiffAdjust::call().get_diff(auction.nft_id);
+                    // Defensive: in edge cases active can read 0; fall back to stored floor.
+                    if auction.enabled && price_raw == 0 {
+                        price_raw = rate_table
+                            .get_index_pk()
+                            .get(&auction.nft_id)
+                            .map(|r| r.floor_difficulty)
+                            .unwrap_or(0);
+                    }
```
Velua, line 110: This is too defensive, and not expected behaviour from `DiffAdjust` so if there's really an edge case found it should be patched on `DiffAdjust` itself.

packages/user/PremAccounts/service/src/lib.rs
```diff
@@ -0,0 +1,326 @@
...
+
+    impl Default for Auction {
+        fn default() -> Self {
+            Self {
+                length: 0,
+                nft_id: 0,
+                enabled: true,
+            }
+        }
+    }
```
Velua: Is this Default needed? When do we lean on it?

packages/user/PremAccounts/service/src/lib.rs
```diff
@@ -0,0 +1,326 @@
...
+    ) {
+        require_caller_is_self();
+
+        check(
+            length >= MIN_ACCOUNT_NAME_LENGTH && length <= MAX_ACCOUNT_NAME_LENGTH,
+            &format!(
+                "market name length must be {}-{}",
+                MIN_ACCOUNT_NAME_LENGTH, MAX_ACCOUNT_NAME_LENGTH,
+            ),
+        );
```
Velua: This check is repeated 3 times, I think it's worth moving to it's own function.

packages/user/PremAccounts/service/src/lib.rs
```diff
@@ -0,0 +1,326 @@
...
+        let auctions_table = AuctionsTable::new();
+        check(
+            auctions_table.get_index_pk().get(&length).is_none(),
+            "market already exists",
+        );
```
Velua:
```suggestion
        check_none(
            auctions_table.get_index_pk().get(&length),
            "market already exists",
        );
```

packages/user/PremAccounts/service/src/tests.rs
```diff
@@ -0,0 +1,347 @@
...
+            );
+            assert!(
+                err.contains(&expected),
+                "unexpected error for length {length}: {err}"
+            );
+        }
+
+        // --- buy on max premium length (7) ---
+        Tokens::push_from(&chain, alice)
+            .credit(1, PremAccounts::SERVICE, 1000_0000u64.into(), "".into())
```
Velua, line 283: Is it a safe assumption that the sys token will always be `1`? I'm not actually sure, to be extra safe one could have the sys token returned from `setup_tokens` 
```
    fn setup_tokens(chain: &psibase::Chain) -> Result<(), psibase::Error> {
```

packages/user/PremAccounts/ui/src/pages/buy-page.tsx
```diff
@@ -0,0 +1,369 @@
...
+            price != null
+                ? canonicalTokenAmountToRaw(price, token.precision)
+                : null;
+        if (maxRaw != null && priceRaw != null && maxRaw < priceRaw) {
+            setError(
+                "Max cost is below the current price. Raise max cost to at least the displayed price.",
+            );
+            setIsLoading(false);
+            return;
+        }
```
Velua: Isn't this check already done on the plugin? 
```
        if max_cost_u64 < ask_u64 {
            return Err(ErrorType::MaxCostBelowCurrentAsk.into());
        }
```
Why repeat it on the UI?

packages/user/PremAccounts/ui/src/lib/prem-service.ts
```diff
@@ -0,0 +1,20 @@
...
+            method: "getAccount",
+            params: [accountName],
+            service: "accounts",
+            intf: "api",
+        });
+        return Boolean(res && typeof res === "object" && "accountNum" in res);
+    } catch {
+        return false;
+    }
+}
```
Velua: Should remove and use `import { isAccountAvailable } from "@shared/lib/get-account";` instead.

Velua (changes requested):
I'm against the introduction of `expandToCanonicalTokenDecimal` and `canonicalTokenAmountToRaw` . Especially given from what I can see their only use in this case is to replicate throwing an error the plugin is already doing anyway. I don't doubt there's some use cases where doing some things in JS over plugin is warranted but I don't see this as one of them.

I think we're better off leaning on plugin functions more and working with the real `Quantity` and `Decimal` implementations whenever possible, rather than make further attempts of mimicking their behaviour in JavaScript. I also fear that any JS helpers like this makes it easier for a developer to perform checks in his isolated UI code rather than the shared plugin code. I think the lack of our JS support with these kind of things is actually a pro not a con.

My blockers are 
- Weird edge case handling in `DiffAdjust`
- `expandToCanonicalTokenDecimal` and `canonicalTokenAmountToRaw` being in a shared library, at minimum I argue for their containment in `PremAccounts/ui` until a broader conversation over number handling and responsibilities of UI vs plugin with the team.

packages/shared-ui/lib/quantity.ts
cool-ant: https://github.com/gofractally/psibase/issues/1844

packages/user/PremAccounts/service/src/lib.rs
cool-ant: https://github.com/gofractally/psibase/issues/1845

packages/user/PremAccounts/plugin/src/lib.rs
James-Mart: I don't necessarily agree, we overuse results.

* “this might reasonably fail” --> Result
* “this must not fail unless there’s a bug / invariant violation” --> panic is fine

In this case, we are generating a keypair. It's reasonable to expect that should never fail. In fact, `generate_unmanaged_keypair` itself should not even return a result.

---------------------
The current exception to this advice is:
https://github.com/gofractally/psibase/issues/1846

packages/user/PremAccounts/service/src/tests.rs
James-Mart: It is not a safe assumption in general, but also this is a unit test so it is probably safe if it's constructed to set up the system token immediately

packages/user/PremAccounts/plugin/src/lib.rs
James-Mart: Would it make you feel better if the doc comments explicitly mention that whitespace on the edges of the input is automatically trimmed, so it wouldn't be a silent fix?

Otherwise I don't really get why you think we have to push trimming to the caller? If it's a common enough user error that can be trivially handled here, why not?

packages/user/PremAccounts/service/src/lib.rs
James-Mart: > one must lose tamper guarantees in order to transfer / resell it

Claiming is the loss of tamper guarantees. So once you claim, you can't resell. When we add resale to premium-accounts, only unclaimed accounts are eligible.

CMakeLists.txt
```diff
@@ -136,7 +136,7 @@ if(DEFINED WASI_SDK_PREFIX)
 
     set(SERVICES Accounts AuthDelegate AuthSig AuthAny CommonApi
         CpuLimit Db Events EventIndex Explorer Invite
-        Nop Packages Producers HttpServer Sites RAccounts
+        Nop Packages PremAccounts Producers HttpServer Sites RAccounts
```
swatanabe: This list is for C++ services

rust/psibase/src/services/diff_adjust.rs
```diff
@@ -36,9 +36,32 @@
...
 #[crate::service(name = "diff-adjust", dispatch = false, psibase_mod = "crate")]
 #[allow(non_snake_case, unused_variables)]
 pub mod Service {
+    use crate::{AccountNumber, Fracpack, TimePointSec, ToSchema};
+
+    use async_graphql::SimpleObject;
+    use serde::{Deserialize, Serialize};
+
+    #[table(name = "RateLimitTable", index = 0)]
+    #[derive(Default, Fracpack, ToSchema, SimpleObject, Serialize, Deserialize, Debug)]
```
swatanabe: The `Fracpack` macro is marked `// TODO: remove`. Use `Pack` and `Unpack`.

packages/user/PremAccounts/plugin/src/lib.rs
cool-ant: falls in same category James called out above https://github.com/gofractally/psibase/pull/1804#discussion_r3161739026

packages/user/PremAccounts/plugin/src/lib.rs
cool-ant: falls in same category James called out above https://github.com/gofractally/psibase/pull/1804#discussion_r3161739026

packages/user/PremAccounts/query-service/src/lib.rs
cool-ant: why `with_service(SERVICE) rather than just `read()`?

packages/user/PremAccounts/query-service/src/lib.rs
cool-ant: The floor_difficulty.max(...) clamp only runs inside the windows_elapsed > 0 && below_target branch — the return value itself is never re-clamped. So if active_difficulty is 0 and either no full window has elapsed, or the counter wasn't below target_min, get_diff returns 0 regardless of the configured floor.

Perhaps we should 1) disallow 0 as a valid `initial_price`, or 2) clamp `get_diff()`'s return to be at minimum `target_floor`?

packages/user/PremAccounts/service/src/lib.rs
cool-ant: fyi, it was just timing. I thumb-up once I've made the change. then there's 0 < x < 3 hours before I push those changes

packages/user/PremAccounts/query-service/src/lib.rs
Velua: I thought that because this is in the query directory one needs to specify the service but maybe that's not needed.

packages/user/Config/ui/src/hooks/premium-name-markets/use-configured-markets.ts
cool-ant: I can't respond on the original chain for some reason

packages/user/PremAccounts/plugin/src/lib.rs
```diff
@@ -0,0 +1,206 @@
...
+    fn buy(account: String, max_cost: String) -> Result<(), Error> {
+        let acct_name = AccountNumber::from_exact(&account)
+            .map_err(|err| ErrorType::InvalidAccountName(err.to_string()))?;
+
+        let service_account = prem_accounts::SERVICE.to_string();
+
+        let sys_token_id =
+            TokensHelpers::fetch_network_token()?.ok_or(ErrorType::SystemTokenNotDefined)?;
+
+        let length = account.to_string().len() as u8;
```
James-Mart: `to_string()` isn't needed (account is already a string)

James-Mart (approved):
Noted one additional minor fix

packages/user/PremAccounts/service/src/tests.rs
```diff
@@ -0,0 +1,350 @@
...
+        );
+
+        let (pub_pem, _priv_pem) = generate_keypair()?;
+        let der = pem::parse(pub_pem.trim())
+            .map_err(|e| anyhow::anyhow!(e))?
+            .into_contents();
+
+        PremAccounts::push_from(&chain, alice)
+            .claim(account!("test"), SubjectPublicKeyInfo::from(der))
+            .get()?;
```
Velua, line 161: Claim no longer requires 2nd param.

packages/user/PremAccounts/service/src/lib.rs
```diff
@@ -0,0 +1,306 @@
...
+            purchased_accounts_table.get_index_pk().get(&account),
+            "account not purchased",
+        );
+
+        check(
+            purchased_account.owner == get_sender(),
+            "account not purchased by sender",
+        );
+
+        AuthDelegate::Wrapper::call().setOwner(get_sender());
```
Velua: We're mutating the owner of `prem-accounts` not `*account being purchased*` , so Alice ends up becoming owner of `prem-accounts` then if Bob buys and claims he'll end up becoming owner of `prem-accounts` according to AuthDelegate too. 

Ideally something like `call_from(account)` would get you out of trouble but I don't think we can get away with that without additional privileges, need to check in with @James-Mart or @swatanabe on how to tackle this.

Velua (changes requested):

packages/user/PremAccounts/service/Cargo.toml
```diff
@@ -0,0 +1,27 @@
...
+rust-version.workspace = true
+edition.workspace = true
+publish.workspace = true
+
+[package.metadata.psibase]
+server = "r-prem-accts"
+plugin = "prem-accounts-plugin"
+data = [{ src = "../ui/dist", dst = "/" }]
+postinstall = [
+    { sender = "prem-accounts", service = "prem-accounts", method = "init", data = {} },
```
Velua, line 13:
```suggestion
    { sender = "prem-accounts", service = "sites", method = "enablespa", data = { enable = true } },
    { sender = "prem-accounts", service = "prem-accounts", method = "init", data = {} },
```

Velua (changes requested):

packages/user/PremAccounts/service/Cargo.toml
James-Mart: The UI is temporary

packages/user/PremAccounts/service/src/lib.rs
James-Mart: Good catch!
This should be `AuthDelegate::Wrapper::call_as(account).setOwner(get_sender());`

packages/user/PremAccounts/query-service/src/lib.rs
cool-ant: @James-Mart ? I don't know when with_service() is relevant. Pls advise.

Velua (approved):

Issue #1267, opened 2025-07-01
Add onBlock event


James-Mart:
We just need to create an event for a block, then we can take advantage of the fact that event ids are monotonically increasing.

_Originally posted by @swatanabe in https://github.com/gofractally/psibase/pull/1254#discussion_r2175814973_

This should allow us to, for any event, determine the block number in which the event was emitted.

James-Mart:
https://github.com/gofractally/psibase/pull/1812

PR #1812, opened 2026-04-08
Gql events with block

James-Mart:
Closes 
https://github.com/gofractally/psibase/issues/1267


libraries/psibase/common/include/psibase/serveGraphQL.hpp
```diff
@@ -1022,12 +1064,81 @@ namespace psibase
...
          return connection;
       }
 
+      template <typename EdgeType>
+      void addBlockContext(std::vector<EdgeType>& edges) const
+      {
+         uint64_t minRowId = extract_cursor(edges.front().cursor);
+         uint64_t maxRowId = extract_cursor(edges.back().cursor);
+
+         auto markers = fetchBlockStartMarkers(minRowId, maxRowId);
```
swatanabe: This has unbounded runtime cost if the events are sparse.

James-Mart: Changed the strategy

libraries/psibase/common/include/psibase/serveGraphQL.hpp
```diff
@@ -1022,12 +1065,84 @@ namespace psibase
...
+      template <typename EdgeType>
+      void addBlockContext(std::vector<EdgeType>& edges) const
+      {
+         auto edgeBlockStarts = fetchEdgeBlockStarts(edges);
+         if (edgeBlockStarts.empty())
+            return;
+
+         std::unordered_map<BlockNum, BlockTime> blockTimeCache;
+
+         auto db            = proxyKvOpen(DbId::blockLog, {}, KvMode::read);
```
swatanabe: Better to call `to<Db>().open`. `proxyKvOpen` is something of a `Table` implementation detail. The main reason it exists is to allow the tables to work in the tester, too.

libraries/psibase/common/include/psibase/serveGraphQL.hpp
```diff
@@ -1022,12 +1065,84 @@ namespace psibase
...
+         auto blockLogIndex = TableIndex<Block, uint32_t>{db, {}, false};
+
+         auto lookupBlockTime = [&](BlockNum bn) -> std::optional<BlockTime>
+         {
+            auto it = blockTimeCache.find(bn);
+            if (it != blockTimeCache.end())
+               return it->second;
+            auto block = blockLogIndex.get(bn);
```
swatanabe: This can be quite an expensive lookup, because blocks can be large. Consider putting the time in the blocks start event. At least, use `getView` to avoid unpacking the entire block.

libraries/psibase/common/include/psibase/serveGraphQL.hpp
```diff
@@ -1022,12 +1065,84 @@ namespace psibase
...
       }
 
+      template <typename EdgeType>
+      void addBlockContext(std::vector<EdgeType>& edges) const
+      {
+         auto edgeBlockStarts = fetchEdgeBlockStarts(edges);
+         if (edgeBlockStarts.empty())
+            return;
+
+         std::unordered_map<BlockNum, BlockTime> blockTimeCache;
```
swatanabe: Because `edgeBlockStarts` is ordered, only the most recent block time added to the cache can ever be reused.

libraries/psibase/common/include/psibase/serveGraphQL.hpp
```diff
@@ -1084,16 +1199,20 @@ namespace psibase
...
-         std::string_view c = *cursor;
-         auto [ptr, ec]     = std::from_chars(c.data(), c.data() + c.size(), b);
+         uint64_t b{};
+         auto [ptr, ec] = std::from_chars(c.data(), c.data() + c.size(), b);
          check(ec == std::errc{}, "Invalid cursor");
          check(ptr == c.data() + c.size(), "Invalid cursor: not all characters consumed");
          return b;
       }
 
+      uint64_t extract_cursor(const std::string& cursor) const
```
swatanabe: This shouldn't be needed except to disambiguate, and I don't see another overload of `extract_cursor`, besides this and the `string_view` one.

libraries/psibase/common/include/psibase/serveGraphQL.hpp
```diff
@@ -129,6 +130,41 @@ namespace psibase
...
    }
 
+   /// Additional fields on each event edge providing block context.
+   template <typename Node, psio::FixedString EdgeName>
+   struct EventEdge
+   {
+      Node                              node;
+      std::string                       cursor;
+      std::optional<psibase::BlockNum>  blockNum;
+      std::optional<psibase::BlockTime> blockTime;
```
swatanabe, line 136: This isn't really standard usage of graphql cursors. Fields should be part of the node.

rust/psibase/src/graph_ql.rs
```diff
@@ -673,8 +715,80 @@ impl<T: DeserializeOwned + OutputType> EventQuery<T> {
...
+        for (i, (edge_rowid, _)) in edges.iter().enumerate() {
+            if i > 0 {
+                values.push_str(", ");
+            }
+            let _ = write!(values, "({}, {})", i, edge_rowid);
+        }
+
+        let sql = format!(
+            "WITH edges(edgeIndex, edgeRowId) AS (VALUES {}) \
+            SELECT e.edgeIndex, m.blockNum \
```
swatanabe: I've been looking through the sqlite reference and I think this can be done without a join, by putting the nested `SELECT` in the result:
```
SELECT (SELECT blockNum FROM blockStart ...)
```

- https://www.sqlite.org/syntax/result-column.html
- https://www.sqlite.org/lang_expr.html#subquery_expressions

James-Mart: I'm taking your advice from another comment and just emitting the blocktime in the event, but I think doing so obviates this comment
```
A subquery that returns a single column is a scalar subquery and can be used most anywhere. A subquery that returns two or more columns is a [row value](https://www.sqlite.org/rowvalue.html) subquery and can only be used as an operand of a comparison operator or as the value in an UPDATE SET clause whose column name list has the same size.
```

libraries/psibase/common/include/psibase/serveGraphQL.hpp
James-Mart: Removed the cache

libraries/psibase/common/include/psibase/serveGraphQL.hpp
James-Mart: Added the blocktime to the event

libraries/psibase/common/include/psibase/serveGraphQL.hpp
James-Mart: Removed

libraries/psibase/common/include/psibase/serveGraphQL.hpp
James-Mart: No longer needed now that I'm reading blocktime from the event

libraries/psibase/common/include/psibase/serveGraphQL.hpp
James-Mart: Moved to the node (flattened)
`edges { cursor node { block { blockNum blockTime ...} field1 field2 } }`

libraries/psibase/common/include/psibase/serveGraphQL.hpp
```diff
@@ -129,6 +129,66 @@ namespace psibase
...
+   template <typename T>
+   struct WithBlockContext
+   {
+      EventBlockInfo block;
+      T              data;
+      PSIO_REFLECT(WithBlockContext, block, data)
+   };
+
+   template <typename T>
+   constexpr auto get_type_name(const WithBlockContext<T>*)
```
swatanabe: This should return `const char*`, not `FixedString`. Make a static constant and return `.c_str()`.

libraries/psio/include/psio/graphql.hpp
```diff
@@ -1456,6 +1454,38 @@ namespace psio
...
+   template <typename T, typename OS, typename E>
+   auto make_field_resolver(const T&    value,
+                            gql_stream& input_stream,
+                            OS&         output_stream,
+                            const E&    error,
+                            bool        allow_unknown_members,
+                            bool&       first)
+   {
+      return [&, allow_unknown_members](std::string_view field_name,
+                                       std::string_view alias) -> std::pair<bool, bool>
```
swatanabe: - I don't think a pair is a reasonable return type. A simple bool with processing of not found handled by the resolver is easier to understand.
- Distinguishing between "not found" and "error" is only necessary for `allow_unknown_members`, which is non-conforming GraphQL, and is only used by the obsolete `EventDecoder`.

rust/psibase/src/graph_ql.rs
```diff
@@ -414,6 +419,132 @@ impl<'de> Deserialize<'de> for SqlRow {
...
+
+fn fetch_block_context(rowids: &[u64]) -> async_graphql::Result<HashMap<usize, EventBlockInfo>> {
+    let mut result = HashMap::new();
+    if rowids.is_empty() {
+        return Ok(result);
+    }
+
+    let block_starts = fetch_block_start_rows(rowids)?;
+    for block_start in block_starts {
+        if block_start.row_index >= rowids.len() {
```
swatanabe: Is this condition needed?

James-Mart: Removed

libraries/psio/include/psio/graphql.hpp
James-Mart: * now returns bool
* removed allow_unknown_members handling

libraries/psibase/common/include/psibase/serveGraphQL.hpp
James-Mart: Done

rust/psibase/src/graph_ql.rs
swatanabe: The rust version doesn't seem to depend on `row_index` being in bounds for safety. The C++ version has undefined behavior if `rowIndex` is out of bounds. It always *should* be in bounds, but it comes from the events service and is not checked. I'd prefer to verify the data, even though it comes from a system service.

James-Mart: I see, restored the check only in c++

swatanabe (approved):

PR #1815, opened 2026-04-13
Fix FractalGen and create FractalTester package

brandonfancher:
Bootstrap the system fractal with 18 test accounts as core guild members. This helps with testing evaluations and other fractal facilities.

### DIFF
--- a/packages/user/FractalTester/Cargo.toml
+++ b/packages/user/FractalTester/Cargo.toml
@@ -1,0 +1,25 @@
+[package]
+name = "frac-tester-pkg"
+description = "Creates 18 auth-any test accounts and adds them as members of the system guild. Depends on FractalGen. Not for production use."
+version.workspace = true
+edition.workspace = true
+repository.workspace = true
+homepage.workspace = true
+rust-version.workspace = true
+publish.workspace = true
+
+[package.metadata.psibase]
+package-name = "FractalTester"
+services = ["frac-tester"]
+
+[package.metadata.psibase.dependencies]
+Sites = "0.23.0"
+HttpServer = "0.23.0"
+FractalGen = "0.23.0"
+
+[lib]
+crate-type = ["rlib"]
+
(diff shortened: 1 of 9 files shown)


packages/user/FractalTester/service/Cargo.toml
```diff
@@ -0,0 +1,35 @@
...
+Sites = "0.23.0"
+HttpServer = "0.23.0"
+Tokens = "0.23.0"
+Symbol = "0.23.0"
+Transact = "0.23.0"
+Nft = "0.23.0"
+VirtualServer = "0.23.0"
+Fractals = "0.23.0"
+Accounts = "0.23.0"
+Producers = "0.23.0"
```
James-Mart, line 26: I think most of these deps aren't needed. Maybe just 
`Accounts`
`Producers`
`Fractals`

And `FractalGen`?

packages/user/FractalTester/Cargo.toml
```diff
@@ -0,0 +1,25 @@
...
+publish.workspace = true
+
+[package.metadata.psibase]
+package-name = "FractalTester"
+services = ["frac-tester"]
+
+[package.metadata.psibase.dependencies]
+Sites = "0.23.0"
+HttpServer = "0.23.0"
+FractalGen = "0.23.0"
```
James-Mart: We union the deps from the root package `Cargo.toml` and any service `Cargo.toml`s. So you can probably delete this whole section as long as you make sure the `service/Cargo.toml` contains all deps your service needs (add FractalGen).

James-Mart:
Nice! 
Minor dependency cleanup, but otherwise lgtm

brandonfancher:
> Nice! Minor dependency cleanup, but otherwise lgtm

@James-Mart Ah, right...because they're already dependencies of `FractalGen`. How's this now?

James-Mart (approved):

PR #1817, opened 2026-04-14
Relax install order

swatanabe:


rust/psibase/src/main.rs
```diff
@@ -1928,6 +1940,20 @@ impl<'a> From<&'a PackageInfo> for LocalPackage<'a> {
...
+impl<'a, R: Read + Seek> From<&'a PackagedService<R>> for LocalPackage<'a> {
+    fn from(other: &'a PackagedService<R>) -> Self {
+        let meta = other.meta();
+        LocalPackage {
+            name: &meta.name,
+            version: &meta.version,
+            description: &meta.description,
+            depends: &meta.depends,
+            accounts: &meta.accounts,
+            sha256: other.hash(),
```
cool-ant, line 1975: Any reason not to include `services` here?

rust/psibase/src/main.rs
cool-ant, line 1992: `services` missing here too

packages/user/Packages/plugin/src/lib.rs
```diff
@@ -372,12 +420,10 @@ impl PrivateApi for PackagesPlugin {
         assert_caller_config_or_self("resolve");
 
         let index = index.into_iter().map(|p| p.into()).collect();
-        let essential = get_essential_packages(&index, &EssentialServices::new());
```
cool-ant, line 375: it looks like this is the only ref to this function; worth removing?

rust/psibase/src/package.rs
```diff
@@ -1138,6 +1186,297 @@ pub fn validate_dependencies<T: Read + Seek>(
...
+            }
+            PackageOpFull::Remove(_) => {
+                // There is no postrm for now, so remove order doesn't matter
+            }
+        }
+    }
+    Ok(result)
+}
+
+fn topological_sort_impl(
```
cool-ant, line 1393: `topological_sort_impl` panics on `indexes.get(vertex).unwrap()` for an unknown vertex; for `Replace`, `indexes` is keyed by `package.name()` only while `updated_packages` (used to filter edges) admits both `meta.name` and `package.name()`.

libraries/psibase/common/src/package.cpp
```diff
@@ -101,6 +119,127 @@ namespace psibase
...
+               if (info.server)
+               {
+                  if (!local)
+                     services.push_back(HttpServer::service);
+                  // x-http is always present
+                  break;
+               }
+            }
+
+            for (const auto& act : readPostinstall(package))
```
cool-ant, line 165: The JSON is parsed again at L671; perhaps we memoize? (Rust already caches)

cool-ant (changes requested):

libraries/psibase/common/src/package.cpp
swatanabe: The only reason that rust caches it is to avoid requiring a mutable reference to the package (for read). C++ isn't as picky about references. I don't expect that the performance matters at all.

rust/psibase/src/package.rs
swatanabe: `meta` in `PackageOp::Replace(meta, info)` indicates the old package. Right now, `meta.name` and `package.name` will always be the same, so there's no issue. If they are ever different, none of the packages being installed can depend on a package that is being removed, so there's still no problem.

cool-ant (approved):

PR #1820, opened 2026-04-15
Add guild service

Velua:
This PR moves all our Guild / Evaluation related logic to a `Guild` service which then interacts with `Fractals`. 

The `fractal-core` plugin and its UI largely remains the same with minimal changes, either removing now redundant plugin functions or calling the new ones presented in `Guilds` and removed from `Fractals`. 

In order to keep UI changes minimal, the UI flow for joining a Fractal remains the same by calling `FractalCore::join` requiring only membership of a guild in order to join the fractal. 

## Levies
Levies are also introduced here, a levy can be placed on any fractal member which taxes any tokens they redeem from their token streams, one of 5% to their recruiter is created upon joining the fractal. 

Further actions addressing their administration can be added later.


brandonfancher:
I'm getting this when trying to load the fractal overview:

image

packages/user/Fractals/service/src/lib.rs
```diff
@@ -1,113 +1,52 @@
...
-    /// * `rep_role` - Representative role account.
     #[action]
-    fn create_fractal(
+    fn create_frac(
         fractal_account: AccountNumber,
-        guild_account: AccountNumber,
+        legislature: AccountNumber,
+        judiciary: AccountNumber,
+        executive: AccountNumber,
+        recruitment: AccountNumber,
```
James-Mart, line 88: You could also randomly generate these account names using the fractal_account as entropy (and then the interfaces wouldn't need to change when the new AccountNumber algo with subaccounts comes along.

packages/user/Guilds/ui/src/App.tsx
brandonfancher, line 1: Not sure yet what this app does, but we shouldn't be reintroducing all of this old UI code that we have (recently) banished from our code base. Can this entire `ui` directory inside `Guilds` go away? Looks like it's all old example boilerplate.

packages/user/PremAccounts/service/src/lib.rs
James-Mart, line 1: This shouldn't be in this pr

rust/psibase/src/services/defacto.rs
James-Mart, line 1: Arguably we don't even need this wrapper, and can just use the exported `Occupations` interface from the fractals service to call into an occupation.

I guess this would only be needed if there is a defacto plugin that needs to schedule actions.

CMakeLists.txt
```diff
@@ -530,6 +530,13 @@ cargo_psibase_package(
     TARGET_DIR ${USER_WORKSPACE_TARGET}
 )
 
+cargo_psibase_package(
+    OUTPUT ${SERVICE_DIR}/Guilds.psi
+    PATH packages/user/Guilds
+    TARGET_DIR ${USER_WORKSPACE_TARGET}
+    DEPENDS ${Guilds_js_DEP}
```
James-Mart: Is this UI needed?

packages/user/Fractals/plugin/src/lib.rs
```diff
@@ -49,22 +53,60 @@ define_trust! {
...
+            add_action_to_transaction(
+                fractals::action_structs::set_r_occ::ACTION_NAME,
+                &packed_args,
+            )
+        };
+
+        set_role_occ(1)?; // Legislature
+        set_role_occ(2)?; // Judiciary
+        set_role_occ(3)?; // Executive
+        set_role_occ(4) // Recruitment
```
James-Mart: You should export named constants from the fractals service, and then you wouldn't need such comments.

James-Mart (commented):
Just a few initial comments. Not a full review

packages/user/Fractals/service/src/helpers/create_managed_account.rs
```diff
@@ -0,0 +1,14 @@
+use psibase::services::{accounts, auth_dyn};
+use psibase::AccountNumber;
+
+pub fn create_managed_account<F>(new_account: AccountNumber, f: F)
+where
+    F: Fn(),
+{
+    accounts::Wrapper::call().newAccount(new_account, "auth-any".into(), true);
```
James-Mart: This should actually use `auth-delegate`. We are slowly stripping uses of `auth-any` from services and the hope is that we can actually not even install `auth-any` in a production boot.

Velua:
From feedback from @brandonfancher 

- [x] Guild metadata form breaks when approaching from /legistlative URL, likely query params failure.
- [x] Cannot register for an evaluation, "sender origin app must be evaluation owner or evaluations"
- [x] Fetching past evaluations seems to throw an error.
- [x] Cannot see guild applications despite being recognised on the service
- [ ] UI seems to not build properly
- [ ] Fix `FractalGen` and `FractalTester` packages

CMakeLists.txt
Velua: Removed

rust/psibase/src/services/defacto.rs
Velua: Removed the wrapper

brandonfancher (changes requested):
`FractalGen` and `FractalTester` packages aren't working. And I cannot create a new fractal from the `Fractals` app UI. I get the following error:

Image

Velua:
@brandonfancher lets check on a call, I've fixed the UI creation bug but cannot replicate the `FractalTester` or `FractalGen` one

James-Mart:
> FractalGen and FractalTester packages aren't working

`FractalGen` must be installed at boot. If you don't, you will get an authorization error when you try to install it. If installed during boot, I cannot replicate any error with it, and `FractalTester` installs as well.

Velua:
I think now there's a graphql PF error issue after that merge, will put back in draft until it's tested again.

brandonfancher (commented):
After my latest review of UI-driven functionality, here's what I've found:

- If a non-member tries to Apply to Join from the FractalCore home/overview page (for the fractal, not the guild), this error follows:

Image

- If they apply to join the guild, we properly see the "Application Open" UI on the guild overview page. But the fractal overview page still has the join fractal CTA UI. Should we even have a Join Fractal CTA on the fractal overview page?
- When I try exiling a member (acting as the guild representative) from the fractal judiciary page, I get this:

Image

- When I try initializing the token (acting as the guild representative) from the fractal legislative page, I get this:

Image

> Mind you that in both of those latter two bullet points, I was acting not as the original founder/representative, but as a representative that the original rep appointed (set as representative). Not sure that matters though.

Velua:
> If a non-member tries to Apply to Join from the FractalCore home/overview page (for the fractal, not the guild), this error follows:
If they apply to join the guild, we properly see the "Application Open" UI on the guild overview page. But the fractal overview page still has the join fractal CTA UI. Should we even have a Join Fractal CTA on the fractal overview page?

As far as I know this is the only place we offer the opportunity to join the fractal, by default we will automatically join them to the fractal when they join an eligible guild but should they join the guild before the guild is ranked by the fractal they will remain outside of the fractal until they manually join. 

I've since improved the error message so it will actually explain what needs to happen next, I think this is sufficient for the purposes of this PR, we can plan where a better place to join might be in another PR I reckon. 

I've fixed the remaining errors you've raised with exile and init token.

packages/user/Fractals/service/src/helpers/create_managed_account.rs
James-Mart: This should actually use `auth-delegate`. We are slowly stripping uses of auth-any from services and the hope is that we can actually not even install `auth-any` in a production boot.

packages/user/Fractals/plugin/src/lib.rs
```diff
@@ -1,588 +1,129 @@
...
+        Ok(())
     }
 
-    fn close_eval(guild_account: String) -> Result<(), Error> {
-        let guild = get_guild(guild_account)?;
-        guild.assert_authorized(FunctionName::close_eval)?;
-
-        close(&"fractals".to_string(), guild.eval_id()?)
+    #[psibase_plugin::authorized(None)]
+    fn set_dist_interval(distribution_interval_secs: u32) -> Result<(), Error> {
```
James-Mart, line 106: Should be `High` trust

packages/user/Fractals/plugin/src/lib.rs
```diff
@@ -1,588 +1,129 @@
...
-            guild_account: guild_account.as_str().into(),
-        }
-        .packed();
-
-        add_action_to_transaction(
-            fractals::action_structs::start_eval::ACTION_NAME,
-            &packed_args,
-        )
+    #[psibase_plugin::authorized(None)]
+    fn exile_member(member: String) -> Result<(), Error> {
```
James-Mart, line 100: Should be `High` trust

packages/user/Fractals/service/src/lib.rs
```diff
@@ -116,475 +105,127 @@ pub mod service {
...
-        FractalMember::add(fractal, sender, member_status);
-
-        Wrapper::emit().history().joined_fractal(fractal, sender);
+    fn set_dist_int(distribution_interval_secs: u32) {
+        let mut fractal = Fractal::by_sender();
+        fractal.set_distribution_interval(distribution_interval_secs);
     }
 
-    /// Sets the schedule for an evaluation type within a fractal.
+    /// Set genesis time for a fractal
```
James-Mart, line 128: This comment is not very helpful, I don't really know what the "genesis time" for a fractal even means, the comment should probably explain it.

While you're at it I recommend doing a once-over on the other action comments for similar types of clarity improvements.

James-Mart (commented):
Just a couple more minor things

brandonfancher:
image

@James-Mart: I get this prompt when clicking to join a fractal from the `FractalsCore` fractal overview (home) page. Weren't PRs intended to fix that merged into this branch? Or maybe it's one we missed.

brandonfancher (changes requested):
- Exiling a member from the fractal does not remove them from their guilds (shouldn't it?)
- Initialise token modal should be dismissed after successful init

Velua:
> * Exiling a member from the fractal does not remove them from their guilds (shouldn't it?)

No, under this new approach occupations decide who can / can't engage internally, the Fractal simply decides if it's going to pay them or worry about them on any judicial level. One future use case might be that a guild member needs to get a certain score or 3 month active participant status before he's accepted as a fractal member. 

> * Initialise token modal should be dismissed after successful init

Added handling

brandonfancher (approved):

James-Mart (approved):

PR #1832, opened 2026-04-20
add pretty action data field to staged-tx GraphQL

James-Mart:
Adds a new `data` field to the staged tx graphql schema that returns a pretty/pruned view of the action data in the staged transaction. This can help front-ends display information that is useful to the act of reviewing staged transactions.

Is this the right way to do this?

### DIFF
--- a/packages/system/StagedTx/query-service/src/lib.rs
+++ b/packages/system/StagedTx/query-service/src/lib.rs
@@ -1,10 +1,15 @@
 #[psibase::service]
 #[allow(non_snake_case)]
 mod service {
-    use async_graphql::{connection::Connection, *};
+    use async_graphql::{
+        connection::{Connection, Edge},
+        ComplexObject, SimpleObject,
+    };
+    use psibase::services::packages::{self, InstalledSchemaTable};
     use psibase::*;
     use serde::Deserialize;
     use staged_tx::service::*;
+    use std::fmt::Write;
     use std::str::FromStr;
 
     #[derive(SimpleObject, Deserialize, Debug)]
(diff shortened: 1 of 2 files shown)


packages/system/StagedTx/query-service/src/lib.rs
```diff
@@ -15,15 +20,107 @@ mod service {
...
+        pub method: MethodNumber,
+        #[graphql(name = "rawData")]
+        pub raw_data: Hex<Vec<u8>>,
+    }
+
+    #[ComplexObject]
+    impl ActionView {
+        /// A pretty, truncated JSON representation of the action's arguments,
+        /// decoded using the service's installed schema.
+        async fn data(&self) -> String {
```
swatanabe, line 37: Returning not-quite-JSON as a string is kind of odd. JSON is only sort of human readable, and the non-JSON parts make it unsuitable for further processing. I think it would be better to manage unpacking and pruning the data on the client-side.  I would expect the UI to allow expanding and collapsing sections of the transaction, instead of having a fixed pruning scheme (the CLI doesn't provide an interactive view of traces, so it doesn't need this).

packages/system/StagedTx/query-service/src/lib.rs
```diff
@@ -15,15 +20,107 @@ mod service {
...
+                    for (i, (name, value)) in args.iter().enumerate() {
+                        if i > 0 {
+                            out.push('\n');
+                        }
+                        write!(out, "{}: {}", name, PrunedJsonPretty(value)).unwrap();
+                    }
+                    out
+                }
+                Some(value) => PrunedJsonPretty(&value).to_string(),
+                None if self.raw_data.is_empty() => String::new(),
```
swatanabe, line 57: This case only happens if the raw action data was pruned, which is relevant for traces, but not for staged transactions.

PR #1837, opened 2026-04-22
More supervisor cleanup

James-Mart:
# Standardizing on identifier casing in supervisor 

Converting between casing was a mess in the supervisor and was done at arbitrary points in the lifecycle of a call. This standardizes: Everything is using snake-case identifiers as is used in wit, unless and until a conversion is needed. For example, JCO expects camelCase identifiers so we convert to camelCase at the boundary. No longer does component-parser assume all downstream tooling wants camelCase.

# Simplifying privileged import loading

The imports for privileged plugins were provided in a way that required an additional mapping in the `loader.ts`. This instead defines the functions in the shape expected by the component loader, thereby removing that mapping layer. Adding future intrinsics is simpler because there is one fewer place we need to update.

# Statically import JCO shims

We were lazy importing the jco shims, but JCO internally already statically imports the shims. So Vite was emitting warnings that we were loading dynamically even though it was already loaded statically in the same chunk. So I just import the shims statically and the warnings went away.


cool-ant (approved):
This is a really nice refactor. ❤️

cool-ant (changes requested):
looks like plugin-loader.ts:ensureAllInstantiated() is no longer used

James-Mart:
> looks like plugin-loader.ts:ensureAllInstantiated() is no longer used

Removed

Also fixed some Vite warnings by statically importing jco shims

cool-ant (approved):

PR #1841, opened 2026-04-25
Disk billing

James-Mart:
# Disk billing

This adds billing for "objective" disk storage to complement the existing billing for CPU and NET. 
The `virtual-server` service now distinguishes between "rate limited" and "capacity limited" resources. Rate limited resources are those like network bandwidth and CPU, where there's a max throughput at any given time but that doesn't compound over time. Capacity limited resources are those with fixed limits over the lifetime of the entire network, like disk storage. Once disk is consumed, unless it is freed, it is permanently stored. Therefore, it deserves its own billing strategy.

## Capacity limited resource billing

Capacity limited resources are, rather than billed using a satoshi-proof-of-work-style price adjustment algo as we do for rate-limited resources, billed according to an AMM style relay. More precisely, as storage is consumed, it approaches the max amount of storage available for the network, and climbs up a price curve. At full disk consumption, the area under the curve (which is the total amount of tokens that would have had to be paid into the relay to purchase the disk consumption) is equal to the total token supply of the network token.

**Note**
I rewrote my git history to take ~70 commits and turn it into 6 that are more sanely grouped. Probably not compileable at the intermediate commits.


James-Mart:
One odd thing is that the cost to enable billing changes between boots, by sometimes a substantial amount. I think it's due to install order and therefore what gets written after virtual-server gets installed.
Ideally this would be more deterministic.

swatanabe:
> One odd thing is that the cost to enable billing changes between boots, by sometimes a substantial amount. I think it's due to install order and therefore what gets written after virtual-server gets installed. Ideally this would be more deterministic.

The right way to solve this is to get the initial RAM usage by scanning the entire database. That way, it doesn't depend on when virtual-server is installed. The scan may need to be split into multiple actions.

packages/system/Transact/src/Transact.cpp
```diff
@@ -579,13 +590,19 @@ namespace SystemService
...
             {
-               // Billing network usage *after* first action auth, because first action auth may set
-               //   a billable subaccount on the VirtualServer.
-               uint64_t      netUsage = t.size();
-               AccountNumber sender   = act.sender();
-               to<VirtualServer>().useNetSys(sender, netUsage);
+               // Anything that could be billed should be done after first auth, because first
+               //   auth may set a billable subaccount on the VirtualServer.
+
+               includedTable.put({tapos.expiration, id});
```
swatanabe: Moving this here (inside the `enforceAuth` check) means that the boot transactions are not added to `includedTable`.

packages/system/VirtualServer/service/cpp/include/services/system/VirtualServer.hpp
```diff
@@ -265,11 +287,34 @@ namespace SystemService
...
+      ///   their resource buffer, depending on the specified database and the
+      ///  amount of bytes consumed/freed.
+      void useDiskSys(psibase::AccountNumber user, psibase::DbId db_id, int64_t amount_bytes);
+
       UserService::Quantity get_resources(psibase::AccountNumber user);
 
       void notifyBlock(psibase::BlockNum block_num);
 
-      /// This action specifies which account is primarily responsible for
+      /// This actions specifies which account is primarily responsible for
```
swatanabe: ??

packages/system/VirtualServer/service/src/tables/capacity_limit_pricing.rs
```diff
@@ -0,0 +1,639 @@
...
+}
+
+/// This is a constant product curve, where the fundamental equation is XY = k. In this case,
+/// we derive X = (x+x0) and Y = (y+y0) where x0 and y0 are virtual offsets.
+///
+/// Plugging in our curve endpoints yields:
+/// k = x0(y_max+y0) = y0(x_max + x0)
+///
+/// Algebraically simplifying yields:
+/// k = x_max / x0 = y_max / y0
```
swatanabe: This is not `k`

packages/system/VirtualServer/service/src/tables/capacity_limit_pricing.rs
```diff
@@ -0,0 +1,639 @@
...
+///
+/// We refer to this single constant as the "shape parameter", or `D`. D controls how "aggressive"
+/// pricing is near the endpoints. Smaller D means larger offsets (smoother), larger D means
```
swatanabe: I wouldn't use the term "smooth" here as it has a mathematical definition. The curve is always smooth ($C^∞$).

packages/system/VirtualServer/service/src/tables/capacity_limit_pricing.rs
```diff
@@ -0,0 +1,639 @@
...
+    /// Substituting yields:
+    /// k = (x_max * y_max)(D+1) / D^2
+    pub(crate) fn new(max_reserves: u64, max_resources: u64, curve_d: u64) -> Self {
+        let xm = max_reserves as u128;
+        let ym = max_resources as u128;
+        let d = curve_d as u128;
+        Curve {
+            x0: (max_reserves / curve_d),
+            y0: (max_resources / curve_d),
+            k: (xm * ym * (d + 1)).div_ceil(d * d),
```
swatanabe, line 122: The multiplications can overflow, and if $d = 1$, the result can also overflow.

packages/system/VirtualServer/service/src/tables/capacity_limit_pricing.rs
```diff
@@ -0,0 +1,639 @@
...
+    ///
+    /// Substituting yields:
+    /// k = (x_max * y_max)(D+1) / D^2
+    pub(crate) fn new(max_reserves: u64, max_resources: u64, curve_d: u64) -> Self {
+        let xm = max_reserves as u128;
+        let ym = max_resources as u128;
+        let d = curve_d as u128;
+        Curve {
+            x0: (max_reserves / curve_d),
+            y0: (max_resources / curve_d),
```
swatanabe: It probably deserves calling out that both `y0` and `x0` are rounded down, as this is necessary to ensure that `pos_from_resources` reliably rounds up.

packages/system/VirtualServer/service/src/tables/capacity_limit_pricing.rs
```diff
@@ -0,0 +1,639 @@
...
+    pub(crate) x0: u64,
+    pub(crate) y0: u64,
+    pub(crate) k: u128,
+}
+
+pub(crate) struct CurvePosition {
+    pub(crate) reserves: u64,
+
+    x: u128,
+    y: u128,
```
swatanabe: These seem to be `X` and `Y`, rather than `x` and `y`

packages/system/VirtualServer/service/src/tables/capacity_limit_pricing.rs
```diff
@@ -0,0 +1,639 @@
...
+    // Cost of consuming `amount` units of resource.
+    // delta_x = ceil(k / (Y - delta_y)) - X
+    // Rounds UP (user pays more).
+    pub(crate) fn cost_of(&self, amount_consumed: u64) -> u64 {
+        if amount_consumed == 0 {
+            return 0;
+        }
+        let dy = amount_consumed as u128;
+
+        check(self.y > dy, "Insufficient disk capacity");
```
swatanabe: Shouldn't this check that `y` doesn't drop below `y0`? Otherwise `dx as u64` could overflow.

packages/system/VirtualServer/service/src/tables/capacity_limit_pricing.rs
```diff
@@ -0,0 +1,639 @@
...
+        }
+    }
+
+    /// Positions the curve by solving for reserves from resources.
+    /// reserves = ceil(k / (resources + y0)) - x0
+    ///
+    /// Rounds the required reserves up, keeping the pool fully capitalized.
+    pub(crate) fn pos_from_resources(&self, resources: u64) -> CurvePosition {
+        let y = resources as u128 + self.y0 as u128;
+        let reserves = (self.k.div_ceil(y) - self.x0 as u128) as u64;
```
swatanabe: If `x_max == u64::MAX`, is it possible for `reserves` to overflow a u64, due to rounding up?

packages/system/VirtualServer/service/src/lib.rs
```diff
@@ -538,9 +694,113 @@ mod service {
...
+            return;
+        }
+
+        if amount_bytes == 0 {
+            return;
+        }
+
+        const SYSTEM_USER_SENTINEL: AccountNumber = AccountNumber::new(0);
+        let user = if user == SYSTEM_USER_SENTINEL {
+            AccountNumber::from("system")
```
swatanabe: This requires the `"system"` account to be reserved. I think it would be better to propagate `AccountNumber{0}` through everywhere. Either that or make `Transact` use an appropriate account. I don't think translating from `0` to `system` in `useDiskSys` makes sense.

packages/system/Transact/src/Transact.cpp
James-Mart: Any reason I shouldn't move `includedTable.put({tapos.expiration, id});` to after all action processing?

packages/system/VirtualServer/service/src/tables/capacity_limit_pricing.rs
James-Mart: Fixed

packages/system/VirtualServer/service/src/tables/capacity_limit_pricing.rs
James-Mart: Fixed, use flat/steep instead of smooth/aggressive

packages/system/VirtualServer/service/src/tables/capacity_limit_pricing.rs
James-Mart: Fixed

packages/system/VirtualServer/service/src/tables/capacity_limit_pricing.rs
James-Mart: Done

packages/system/VirtualServer/service/src/tables/capacity_limit_pricing.rs
James-Mart: Fixed (in multiple places)

packages/system/VirtualServer/service/src/tables/capacity_limit_pricing.rs
James-Mart: Fixed

packages/system/VirtualServer/service/src/tables/capacity_limit_pricing.rs
James-Mart: Good catch. Fixed

packages/system/VirtualServer/service/src/lib.rs
James-Mart: Now using `AccountNumber{0}`

packages/system/Transact/src/Transact.cpp
swatanabe: I can't think of any reason why it would fail

James-Mart: Done

packages/user/Tokens/service/src/lib.rs
```diff
@@ -4,7 +4,7 @@ pub mod helpers;
 pub mod tables;
 pub use tables::tables::{BalanceFlags, TokenFlags};
 
-#[psibase::service(tables = "tables::tables")]
+#[psibase::service(tables = "tables::tables", recursive = true)]
```
swatanabe: This looks really dangerous. What happens if a user exactly consumes his resource balance, causing the subaccount row to be deleted and triggering a refund to the same subaccount? What if the fee receiver has auto-debit enabled, and triggers a billing event on his system token balance?

packages/system/Transact/src/Transact.cpp
```diff
@@ -531,6 +535,12 @@ namespace SystemService
                                      act.method().unpack().str()));
          }
       }
+
+      namespace
+      {
+         AccountNumber trxSender = {};
```
swatanabe: Consider making this optional and explicitly setting it to `{}` in the places where we want the special system billing to detect unexpected writes. We could also use `Transact::service` to get the same effect, I suppose. `transact` will not normally have any resources, so billing to it will fail.

packages/system/Transact/src/Transact.cpp
```diff
@@ -660,6 +677,42 @@ namespace SystemService
...
+         }
+         else
+         {
+            delta = static_cast<int64_t>(newValueLen) - static_cast<int64_t>(oldValueLen);
+         }
+         if (delta != 0)
+         {
+            if constexpr (enable_print)
+            {
+               psibase::writeConsole(
```
swatanabe: This doesn't add any information that isn't already in the transaction trace

packages/system/VirtualServer/service/src/tables.rs
```diff
@@ -183,14 +196,53 @@ pub mod tables {
         /// is rounded up to the nearest billable unit.
         pub billable_unit: u64,
     }
+
+    /// Tracks the token discrepancy between the relay curve state and the
+    /// actual relay reserve while billing is disabled. Using two u64s to ensure no
+    /// overflow can occur, but exposing a signed i128 interface to make it easier
+    /// to reason about.
+    #[derive(Default, Clone, Debug, Serialize, Deserialize, ToSchema, Fracpack)]
+    pub struct VirtualBalance {
```
swatanabe: We're way overusing the term "virtual".

- `VirtualServer` doesn't really explain what this service does.
- `VirtualBalance` is more an unpaid or outstanding balance.

packages/system/VirtualServer/service/src/tables.rs
```diff
@@ -183,14 +196,53 @@ pub mod tables {
...
+        #[graphql(skip)]
+        pub max_reserve: u64,
+        /// Total resource capacity
+        pub max_capacity: u64,
+        /// Curve / virtual-offset parameter
+        pub curve_d: u64,
+        /// Fee on refunds, parts per million
+        pub fee_ppm: u32,
+        /// Virtual balance tracking for billing-disabled periods
+        pub virtual_balance: VirtualBalance,
```
swatanabe: I think this method of tracking is overly complex. This is always the difference between the reserves dictated by the curve and the actual reserves. I think it would be better to track these two and calculate the difference on demand. The actual reserves are available from the relay subaccount.

packages/system/VirtualServer/service/src/tables.rs
```diff
@@ -183,14 +196,53 @@ pub mod tables {
...
+    #[table(name = "CapacityPricingTable", index = 7)]
+    #[derive(Serialize, Deserialize, ToSchema, Fracpack, Debug, SimpleObject, Clone)]
+    #[graphql(complex)]
+    pub struct CapacityPricing {
+        #[primary_key]
+        pub resource_id: u8,
+
+        /// Reserve tokens held in relay
+        #[graphql(skip)]
+        pub reserve: u64,
```
swatanabe: This is technically redundant, right? It should be possible to calculate it in terms of the other parameters.

packages/system/VirtualServer/service/src/tables/rate_limit_pricing.rs
```diff
@@ -66,20 +65,24 @@ fn create_diff_adjust(doubling_time_sec: u32, halving_time_sec: u32) -> u32 {
...
-    usage_history.truncate(num_blocks_to_average as usize);
+    let n = num_blocks_to_average as u64;
 
-    let avg = average(&usage_history);
-    let ppm = ratio_to_ppm(avg, capacity);
+    // Exponential moving average (https://en.wikipedia.org/wiki/Exponential_smoothing):
+    //   avg = α * sample + (1 − α) * avg_prev
+    // Setting α = 1/N and multiplying through by N to keep everything in integer arithmetic:
+    //   avg = (sample + (N − 1) * avg_prev) / N
+    *avg_usage = (*last_block_usage + (n - 1) * *avg_usage) / n;
```
swatanabe: `average` used `u128` to avoid overflow. Was that unnecessary?

packages/system/VirtualServer/service/src/tables/capacity_limit_pricing.rs
```diff
@@ -0,0 +1,675 @@
...
+
+    pub fn initialize(resource: ResourceType, max_capacity: u64, curve_d: u64, fee_ppm: u32) {
+        // Since reducing the reserve (burning) is supported, we can initialize
+        //   the reserve now at u64::MAX and then reduce it later (if needed) when
+        //   the total system token supply is known.
+        let max_reserve = u64::MAX;
+
+        check(max_capacity > 0, "max_capacity must be > 0");
+        check(curve_d > 0, "curve_d must be > 0");
+        check(fee_ppm > 0, "fee_ppm must be > 0");
```
swatanabe: Is there an explanation anywhere of why a fee of 0 isn't allowed?

packages/system/VirtualServer/service/src/tables/capacity_limit_pricing.rs
```diff
@@ -0,0 +1,675 @@
...
+use psibase::services::tokens::{Decimal, Precision, Quantity, Wrapper as Tokens};
+use psibase::*;
+use std::cell::Cell;
+
+use crate::math_utils::PPM;
+
+// Note: `pub(crate)` is used so the functions are available for the unit tests
+
+fn relay_sub(resource: ResourceType) -> String {
+    format!("{}.relay", resource.name().to_lowercase())
```
swatanabe: `"disk" => virtual-server::bill_to_sub("relay")`

packages/system/VirtualServer/service/src/tables/capacity_limit_pricing.rs
```diff
@@ -0,0 +1,675 @@
...
+    }
+
+    pub fn consume(
+        &self,
+        amount_consumed: u64,
+        user: AccountNumber,
+        sub_account: Option<String>,
+    ) -> u64 {
+        let resource = self.resource();
+        let cost = Cell::new(0u64);
```
swatanabe: This doesn't need to be a `Cell`. `FnOnce` can capture mutable references.

packages/system/VirtualServer/service/src/lib.rs
```diff
@@ -87,34 +165,49 @@ mod service {
...
-            _ => "UnknownResource",
-        }
-    }
+    use crate::resource_type::ResourceType;
+    use ResourceType::{Cpu, Disk, Net};
+
+    fn get_prealloc() -> u64 {
+        let table_size = |table: NativeTable| {
+            let prefix = (table, NATIVE_TABLE_PRIMARY_INDEX).to_key();
+            let handle = unsafe {
```
swatanabe: `KvHandle::new` is better because it uses RAII. You'll just need to add a way to extract the raw handle to pass to `native_raw`, or add a version of `kv_greater_equal_XXX` that returns the value size.

packages/system/VirtualServer/service/src/lib.rs
```diff
@@ -87,34 +165,49 @@ mod service {
...
+        };
 
-    pub fn get_measurement_unit(resource_id: u8) -> &'static str {
-        match resource_id {
-            CPU => "ns",
-            NET => "bytes",
-            _ => "UnknownUnit",
-        }
+        let code_table_size = table_size(CODE_TABLE);
+        let code_by_hash_table_size = table_size(CODE_BY_HASH_TABLE);
```
swatanabe: This isn't enough. To get accurate accounting, you want the total size of everything in the `native` and `service` databases, not just the code tables.

packages/system/VirtualServer/service/src/lib.rs
```diff
@@ -87,34 +165,49 @@ mod service {
...
+                // Returns value size without copying into memory
+                let size = unsafe {
+                    native_raw::kvGreaterEqual(handle, key.as_ptr(), key.len() as u32, 0)
+                };
+                if size == u32::MAX {
+                    break; // No more rows under this prefix
+                }
+                total += size as u64;
+                // Advance cursor past the current key
+                key = get_key_bytes();
```
swatanabe: The key size should be included (to match `transact::kvNotify`)

packages/system/VirtualServer/service/src/lib.rs
```diff
@@ -123,24 +216,35 @@ mod service {
             return;
         }
 
-        InitRow::init();
+        InitRow::init(get_prealloc());
```
swatanabe: Any writes that happen between `get_prealloc` and `transact::resMonitoring(true)` will not be accounted for.

packages/system/VirtualServer/service/src/tests/chain.rs
swatanabe, line 1: An important test for Disk is to iterate over the entire database and calculate the total usage and verify that it is equal to the prealloc + total resource usage.

A few actions that should be run in this test:
- Add a service with the `isVerify` flag.
- Change the producers

packages/system/VirtualServer/service/src/tests/curve_math.rs
```diff
@@ -0,0 +1,71 @@
+use crate::tables::capacity_limit_pricing::Curve;
+
+const MAX_RESERVES: u64 = 1_000_000;
+const MAX_RESOURCES: u64 = 1_000_000;
+const D: u64 = 10;
+
+fn curve() -> Curve {
+    Curve::new(MAX_RESERVES, MAX_RESOURCES, D)
```
swatanabe: Should also test `max_reserves = u64::MAX`, `D = 1`, `(D + 1) * max_resources = 2^{64} + 1`, and any other boundary values.

packages/system/VirtualServer/service/src/tables.rs
James-Mart: I will rename VirtualBalance.

But not sure I agree with the critique of VirtualServer
image

Do you have any suggestions?

swatanabe: The issue with naming is that the virtual server in the diagram is not the virtual server service. The virtual server service just manages some parts of it. Anyway, the name is not introduced in this PR.

packages/system/VirtualServer/service/src/lib.rs
James-Mart: * Using `KvHandle::new` 
* extracted `kv_greater_equal_value_size` from `kv_greater_equal_bytes`

packages/system/VirtualServer/service/src/lib.rs
James-Mart: Key size now included

packages/system/VirtualServer/service/src/lib.rs
James-Mart: Now scanning the total size of everything in native and service databases. To do so required me to make virtual-server privileged and give privileged services the ability to do unbounded scanning of the service db (outside a service prefix). If you have a better idea for how to accomplish it lmk

packages/system/VirtualServer/service/src/lib.rs
James-Mart: Everything is now accounted for. Where possible I moved writes to either before the scan, or after `resMonitoring(true)`. For the few records that doesn't catch which have to be between `get_prealloc` and `resMonitoring(true)`, I manually add those record (and key) sizes in.

packages/system/VirtualServer/service/src/tests/chain.rs
James-Mart: At one point we decided that we were okay with the total usage drifting from the prealloc + total resource usage by precisely the amounts implied by 

> * Add a service with the isVerify flag.
> * Change the producers

Which is what motivated the whole `skip_billing` design, so `vserver` isn't notified of those particular writes.

If you want this to be a true invariant then `skip_billing` could be changed to `system_write` and we track the write the same way we track other `AccountNumber{0}` writes. But also then we need a solution for the pure native table write to the status row from an `isVerify` service... Should it fire `kvnotify`?

packages/system/VirtualServer/service/src/tables.rs
James-Mart: `VirtualBalance` is now removed.

In a future pr perhaps I'll split `VirtualServer` from `Billing` services

packages/system/VirtualServer/service/src/tables.rs
James-Mart: Done

packages/system/VirtualServer/service/src/tables.rs
James-Mart: Right, removed

packages/system/Transact/src/Transact.cpp
James-Mart: Removed

packages/system/Transact/src/Transact.cpp
James-Mart: I now route all skip/end_skip_billing through `transact` callable only by privileged services. When called it sets a new `systemWrite` flag, which allows `trxSender` to be `None`. If `systemWrite` isn't set and `trxSender` is none, we assert, catching all unexpected db writes.

packages/system/VirtualServer/service/src/tables/rate_limit_pricing.rs
James-Mart: No, restored

packages/system/VirtualServer/service/src/tables/capacity_limit_pricing.rs
James-Mart: Simplified (everywhere)

packages/system/VirtualServer/service/src/tables/capacity_limit_pricing.rs
James-Mart: Actually it can be. Removed this constraint

packages/system/VirtualServer/service/src/tables/capacity_limit_pricing.rs
James-Mart: Relay sub for capacity-limited resources is now:
`<service>.<resource>:relay` (e.g. `vserver.disk:relay`)

packages/system/VirtualServer/service/src/tests/curve_math.rs
James-Mart: I now test a bunch of boundary curves, see if you think there's any important remaining gaps.

packages/user/Tokens/service/src/lib.rs
James-Mart: This is the one thing I haven't addressed yet.

Planning a rework that accumulates writes/frees/and fee-receiver transfers and then reconciles at the end of the tx. Do you want me to add that to this PR or followup?

packages/system/VirtualServer/service/src/lib.rs
swatanabe: The check that the key size is >= 8 seems like a vestige of the original native database access controls. I think it should be removed entirely.

packages/user/Tokens/service/src/lib.rs
swatanabe: Do either of the cases I suggested above actually blow up?

rust/psibase/src/trace.rs
```diff
@@ -176,7 +176,8 @@ fn format_string(mut s: &str, indent: usize, f: &mut fmt::Formatter<'_>) -> fmt:
 
 fn format_console(c: &ConsoleTrace, indent: usize, f: &mut fmt::Formatter<'_>) -> fmt::Result {
     write!(f, "{:indent$}console:    ", "")?;
-    format_string(&c.console, indent + 12, f)
+    format_string(c.console.trim_end_matches('\n'), indent + 12, f)?;
```
swatanabe: A single trailing newline will be handled by `format_string`. If there's more than one, I don't think the extras should be removed.

rust/psibase/src/trace.rs
```diff
@@ -279,6 +280,187 @@ fn format_transaction_trace(
...
+            } else if is_kv_notify {
+                collect_direct_disk(c, deltas, shown_children);
+            } else if has_disk_activity(c) {
+                shown_children.push(c);
+            }
+        }
+    }
+}
+
+fn format_signed_bytes(amount: i64, f: &mut fmt::Formatter<'_>) -> fmt::Result {
```
swatanabe: The `+` format flag handles this.

rust/psibase/src/tester.rs
```diff
@@ -1113,6 +1200,71 @@ impl<'a> Caller for ChainPusher<'a> {
...
+        self.call_returns_nothing(method, args)
+    }
+}
+
+/// Extension trait for service wrappers that enables tests to propose actions
+/// through [`staged_tx`](crate::services::staged_tx).
+///
+/// Implement with a one-liner on any service `Wrapper`:
+/// ```ignore
+/// impl ProposeExt for Wrapper {}
```
swatanabe: Can't you define a blanket `impl`?

packages/system/VirtualServer/service/src/tests/curve_math.rs
```diff
@@ -0,0 +1,225 @@
...
+        let pos = c.pos_from_resources(max_resources);
+        let big_y = max_resources as u128 + c.y0 as u128;
+        let x = pos.reserves as u128 + c.x0 as u128;
+        assert!(
+            x * big_y >= c.k,
+            "undercapitalized at empty endpoint: X*Y={} < k={}",
+            x * big_y,
+            c.k,
+        );
+        if pos.reserves > 0 {
```
swatanabe: Can the assert below fail if `pos.reserves == 0`?

packages/system/VirtualServer/service/src/tests/curve_math.rs
```diff
@@ -0,0 +1,225 @@
...
+    for (max_reserves, max_resources, d) in test_curves() {
+        let c = Curve::new(max_reserves, max_resources, d);
+
+        let partial_amount = (max_resources / 2).max(1);
+        let full = c.pos_from_resources(max_resources);
+        let partial = c.pos_from_resources(max_resources - partial_amount);
+        let empty = c.pos_from_resources(0);
+
+        // No-arbitrage: a buy then sell of the same amount never refunds more
+        // than it cost (rounding/clamping can make the refund strictly less).
```
swatanabe: I thought that the cost and refund round-trip exactly (before the fee).

packages/system/VirtualServer/service/src/tests/chain.rs
```diff
@@ -0,0 +1,566 @@
...
+
+fn status_row_bytes(chain: &psibase::Chain) -> u64 {
+    let key = status_key().to_key();
+    chain
+        .kv_get_bytes(DbId::Native, &key)
+        .map(|value| key.len() as u64 + value.len() as u64)
+        .unwrap_or(0)
+}
+
+fn check_disk_invariant(label: &str, chain: &psibase::Chain) -> bool {
```
swatanabe, line 441: Consider running this at the end of every test case in this file.

packages/system/VirtualServer/service/src/tests/chain.rs
```diff
@@ -0,0 +1,566 @@
...
+    tokens::Wrapper::push_from(chain, symbol)
+        .credit(sys, bob, 100_000_0000u64.into(), "".into())
+        .get()?;
+
+    chain.finish_block();
+
+    Ok(())
+}
+
+// shortfall tracking after consumption, capacity increases, and supply reductions
```
swatanabe, line 96: Are there any tests that free enough DISK to dip into the prealloc?

packages/user/Tokens/service/src/lib.rs
James-Mart: > What happens if a user exactly consumes his resource balance, causing the subaccount row to be deleted and triggering a refund to the same subaccount? 

If the fee == 0, then I would expect the refund from deleting the subaccount balance could be exactly sufficient to cover the cost of the creation of the same record when a refund is issued, which will delete the record again. So perhaps I should have left in the fee != 0 logic? That asymmetry would ensure that ping-ponging like this always drains the balance. Thoughts?

> What if the fee receiver has auto-debit enabled, and triggers a billing event on his system token balance?

Are you referring here to the potential churn on the `SharedBalance` rows when auto-debit is enabled? If so, it's handled in this case because `vserver` configures [`KEEP_ZERO_BALANCES`](https://github.com/gofractally/psibase/pull/1841/changes#diff-469ee1bb9907fc56955c27b6ff0672c2d20289ac2a22414100a8ef1a1df9961bR220) which means when any shared balance records it creates hit zero, the records aren't actually deleted.

swatanabe: I'm more concerned about whether reentering the tokens service happens at a point where reentrancy is safe. Token operations write more than a single row. If the `kvNotify` callback changes a row that the outer call is still using, it can result in corruption.

rust/psibase/src/tester.rs
James-Mart: I was trying to avoid making it available on all wrapper outside tests, in particular because I was considering adding an actual `Wrapper::propose(<account>)` to the wrapper for services to more easily use staged tx proposals.

If I do a blanket impl, I'm thinking I should either:
* Rename the function to `tester_propose` or something, or
* Add a psibase "tester' feature and gate the impl behind that feature? Then users would do 
```
[dev-dependencies]
psibase = { workspace = true, features = ["tester"] }
```

packages/system/VirtualServer/service/src/tests/curve_math.rs
James-Mart: Nah, not needed. Removed.

packages/system/VirtualServer/service/src/tests/curve_math.rs
James-Mart: The rounding in `cost_of` and `refund_of` is such that cost is biased up, refund is biased down. Theoretically identical cost/refund roundtrip, but in this implementation refunds can be less than the cost.

rust/psibase/src/trace.rs
James-Mart: Done

rust/psibase/src/trace.rs
James-Mart: Done

rust/psibase/src/tester.rs
swatanabe: Requiring an explicit impl does not really fulfill your goals. If the impl is defined at all for a particular service, it will be available everywhere. It isn't necessary either, because the method will only be available if the trait is in scope. Don't `use ProposeExt` outside tests and there won't be an ambiguity. If you really want it only to be available in tests, make it a method of `TestChain` instead of the wrapper: `t.propose::<Wrapper>().foo(42)`

packages/system/VirtualServer/service/src/tests/chain.rs
James-Mart: Done

rust/psibase/src/tester.rs
James-Mart: Aha, I like `t.propose::<Wrapper>().foo(42)`. Mainly because of 

> I was considering adding an actual Wrapper::propose(<account>) to the wrapper for services to more easily use staged tx proposals

packages/user/Tokens/service/src/lib.rs
James-Mart: I see. `kvNotify` fires before the row is actually updated, so I think the following is possible:
`KvNotify (record deleted) -> transact (delta < 0 ) -> useDiskSys -> Tokens (update record) -> kvNotify (record written) -> transact (delta == 0)

- The delta == 0 would be "wrong" and a side-effect of the outer kvnotify not yet returning.
- Even worse, I think the data that gets written will end up getting corrupted when the stack clears.

> Planning a rework that accumulates writes/frees/and fee-receiver transfers and then reconciles at the end of the tx. Do you want me to add that to this PR or followup?

I am going to do this now.

packages/system/VirtualServer/service/src/tests/chain.rs
James-Mart: Implementing a test for this scenario helped me discover a memory leak:

If we free more bytes than were consumed after vserver-init (dipping into the prealloc), then the frees were not refunded, and the curve stays at `remaining_capacity` == `max_capacity`. That is fine and correct behavior as far as I'm concerned. However, bytes written in this region are billed and cause us to move up the curve like normal. This essentially means any prealloc deficit can never be recovered (you're shrinking your available disk by that deficit amount).

Even though it's an unlikely scenario, this is serious enough in my opinion that I'm now planning to track a `prealloc_deficit` for when we free more than what was consumed, and any writes while there's a prealloc_deficit are unbilled and untracked. So there's symmetry on either side of the prealloc boundary.

I think that also makes it easy to ensure the disk invariant holds:
`bytes written to disk (measured via full scan) == (prealloc - prealloc_deficit) + consumed`

rust/psibase/src/tester.rs
James-Mart: Done

packages/system/VirtualServer/service/src/tests/chain.rs
James-Mart: Done

packages/system/VirtualServer/service/src/tests/curve_math.rs
swatanabe: Wait, I thought the calculation was essentially:

```
cost_of(amount) = f(amount + current) - f(current);
refund_of(amount) = f(current) - f(current - amount);
f(total_resources) = total required reserves
```

All the rounding happens in `f`, and it should be identical. Getting this to round-trip exactly is the whole point of handling the curve this way.

James-Mart: You're right - I believed my incorrect doc comments.
Fixed the comments and tightened the unit test to use equality.

packages/user/Tokens/service/src/lib.rs
swatanabe: > kvNotify fires before the row is actually updated

Do you think that triggering `kvNotify` *after* updating the row would be safer?

James-Mart: Sounds safer to me, but I'd delegate that decision to you

packages/system/VirtualServer/service/cpp/include/services/system/VirtualServer.hpp
```diff
@@ -61,16 +94,77 @@ namespace SystemService
...
+   /// An undercollateralized curve is recollateralized over time using the fees that would otherwise
+   /// be sent to the fee receiver.
+   ///
+   /// Example disk recollateralization flow diagram (simplified for clarity) for a transaction that
+   /// frees bytes:
+   ///
+   /// ```text
+   ///             .---------------.
+   ///            (   free bytes    )
+   ///             '-------+-------'
```
swatanabe: This looks like the old diagram

rust/psibase/src/trace.rs
```diff
@@ -281,6 +282,178 @@ fn format_transaction_trace(
...
+fn format_transaction_disk_usage_trace(
+    schemas: &SchemaMap,
+    ttrace: &TransactionTrace,
+    show_all_ops: bool,
+    f: &mut fmt::Formatter<'_>,
+) -> fmt::Result {
+    let total: i64 = ttrace.action_traces.iter().map(subtree_disk_net).sum();
+    write!(f, "Total disk usage: ")?;
+    write!(f, "{:+}", total)?;
+    writeln!(f, " bytes")?;
```
swatanabe: These can be a single `writeln`.

rust/psibase/src/trace.rs
```diff
@@ -281,6 +282,178 @@ fn format_transaction_trace(
...
+    shown_children: &mut Vec<&'a ActionTrace>,
+) {
+    use crate as psibase;
+    for inner in &atrace.inner_traces {
+        if let InnerTraceEnum::ActionTrace(c) = &inner.inner {
+            let is_kv_notify =
+                c.action.service == transact::SERVICE && c.action.method == method!("kvNotify");
+            if let Some(amount) = use_disk_sys_amount(&c.action) {
+                deltas.push(amount);
+                collect_direct_disk(c, deltas, shown_children);
```
swatanabe: This shouldn't find anything, right? Even if it did, would it count as "direct" disk?

rust/psibase/src/trace.rs
```diff
@@ -281,6 +282,178 @@ fn format_transaction_trace(
     Ok(())
 }
 
+fn use_disk_sys_amount(action: &Action) -> Option<i64> {
+    use crate as psibase;
+    if action.service == virtual_server::SERVICE && action.method == method!("useDiskSys") {
+        let (_, _, amount) = <(AccountNumber, DbId, i64)>::unpacked(&action.rawData).ok()?;
```
swatanabe: `viritual_server::action_structs::useDiskSys` would be better than a hard-coded tuple

rust/psibase/src/services/virtual_server.rs
```diff
@@ -70,16 +100,77 @@ pub struct BufferConfig {
...
+/// frees bytes:
+///
+/// ```text
+///             .---------------.
+///            (   free bytes    )
+///             '-------+-------'
+///                     |
+///                     v
+///          +--------------------------+
+///          | Reduces curve shortfall  |
```
swatanabe: This is also the old diagram

packages/system/Db/src/Db.cpp
```diff
@@ -71,13 +71,16 @@ UniqueKvHandle Db::open(DbId db, psio::view<const std::vector<char>> prefixView,
       }
       else
       {
-         auto expected = psio::convert_to_key(sender);
-         // Subaccounts of the same primary account, all have
-         // shared read access
-         expected.pop_back();
-         if (!std::ranges::starts_with(prefix, expected))
+         if (!isPrivileged(sender))
```
swatanabe: I didn't add this, because even privileged services should prefer not to use this capability. When they really need it (virtual-server should only use it for calculating the prealloc, right?), they can open the database directly and bypass the `db` service.

swatanabe (dismissed):

packages/system/VirtualServer/service/cpp/include/services/system/VirtualServer.hpp
James-Mart: Fixed

rust/psibase/src/trace.rs
James-Mart: Done

rust/psibase/src/trace.rs
James-Mart: Good points, dropped

rust/psibase/src/trace.rs
James-Mart: Done

rust/psibase/src/services/virtual_server.rs
James-Mart: Fixed

packages/system/Db/src/Db.cpp
James-Mart: I see. Updated vserver so it opens the db directly (and removed the `isPrivileged` carve-out in `db` service)

packages/system/VirtualServer/service/src/lib.rs
```diff
@@ -93,28 +463,48 @@ mod service {
...
-    // Resource type IDs
-    pub mod resources {
-        pub const CPU: u8 = 0;
-        pub const NET: u8 = 1;
-        pub const _STORAGE: u8 = 2;
+    use crate::resource_type::ResourceType;
+    use ResourceType::{Cpu, Disk, Net};
+
+    /// Open a DB handle via native `kvOpen` to bypass the `db` service prefix checks.
+    fn open_direct(db: DbId) -> KvHandle {
```
swatanabe: I would put this in the psibase library as a member of `KvHandle`

James-Mart: Done

swatanabe (approved):

PR #1843, opened 2026-04-28
Modernize burst transfer

James-Mart:
The burst-transfer tool was broken.

This restores it and adds a few improvements.

It's just a development tool, so it doesn't strike me as problematic to add a dependency on FaucetTok being installed on the chain you're targeting. This allows it to generate and fund its own accounts automatically rather than needing to set them up manually beforehand. 

I also enabled usage of the psibase config (same as psibase and cargo-psibase CLI) so you can use `-a [dev1|dev2|etc]` to specify the chain.


rust/psibase/src/lib.rs
```diff
@@ -4,6 +4,8 @@ mod actions;
 mod auth;
 mod block;
 mod boot;
+#[cfg(not(target_family = "wasm"))]
+pub mod cli;
```
swatanabe, line 8: Should we also change `psibase/src/main.rs` to `use psibase::cli`?

rust/burst-transfer/src/main.rs
```diff
@@ -52,135 +80,129 @@ struct Args {
...
+    for account in new_accounts {
+        actions.push(faucet_tok::Wrapper::pack_from(sender).dispense(*account));
     }
+
+    let trx = build_transaction(actions, ref_block);
+    let client = reqwest::Client::new();
+    let afmt = ActionFormatter::new(HttpSchemaFetcher {
+        client: &client,
+        base_url: &args.api,
+    });
```
swatanabe: This should be hoisted, so there's only one `ActionFormatter`. The `ActionFormatter` caches schemas. We only want to fetch them once.

rust/burst-transfer/src/main.rs
```diff
@@ -229,33 +265,88 @@ async fn main() -> Result<(), anyhow::Error> {
...
-                        &args,
-                        tok.id,
-                        &counter,
-                        &ref_block,
-                        args.delay == 0,
-                    ));
+                let mut created = 0;
+                let mut created_names = String::new();
+                let new_accounts = if accounts.len() < args.account_count {
+                    let n = std::cmp::min(ACCOUNTS_PER_SETUP, args.account_count - accounts.len());
```
swatanabe: This setup process seems overly complicated. Why is account creation interleaved with transfers?

rust/burst-transfer/src/main.rs
```diff
@@ -229,33 +265,88 @@ async fn main() -> Result<(), anyhow::Error> {
...
                 }
-                futures::future::join_all(transfers).await;
             }
         }
         if args.delay > 0 {
-            println!("sleep...");
-            tokio::time::sleep(tokio::time::Duration::from_millis(args.delay.into())).await;
+            let delay = tokio::time::Duration::from_millis(args.delay.into());
+            if let Some(remaining) = delay.checked_sub(batch_start.elapsed()) {
+                tokio::time::sleep(remaining).await;
```
swatanabe, line 368: The documentation of `delay` says that it's the delay between bursts, which I interpret as the time from the end of one burst to the start of the next. If you want to change this behavior to be interval of starting bursts, then the documentation should be updated to clarify.

rust/burst-transfer/src/main.rs
James-Mart: I did that so I test a stupidly huge number of accounts and it wouldn't wait for all the accounts to be created before starting the transfers. Essentially to continuously transfer and create accounts. Maybe that makes the tool too strange? I could change it to just create them first, I suppose.

rust/psibase/src/lib.rs
James-Mart: Done

rust/burst-transfer/src/main.rs
James-Mart: Hoisted

rust/burst-transfer/src/main.rs
James-Mart: Updated, it just does a create phase now followed by transfer phase

rust/burst-transfer/src/main.rs
James-Mart: Updated the docs

rust/burst-transfer/src/main.rs
```diff
@@ -273,73 +275,61 @@ async fn main() -> Result<(), anyhow::Error> {
...
-                        ));
-                    }
-                    if repeat {
+                let n = std::cmp::min(ACCOUNTS_PER_SETUP, args.account_count - accounts.len());
+                let new_accounts: Vec<_> = (0..n).map(|_| random_account_name()).collect();
+                match push_setup(&args, &new_accounts, &ref_block, &client, &afmt)
+                    .await
+                    .context("Failed to push setup transaction")
+                {
+                    Err(e) => println!("Setup error | {:?}", e),
```
swatanabe: There should probably be some limit to errors, so we don't continue spamming requests that will never succeed forever.

James-Mart: Now bails after 10 total tx errors.

swatanabe (approved):

PR #1849, opened 2026-04-30
Improvement to Quantity

cool-ant:
Quantity was based on JS's Number, which isn't big enough to handle the full spectrum of chain-side u64 amounts. With this switch to using BigInt internatlly, all quantities can now round-trip to/from the chain successfully.
along the way, I've tried to tighten up the naming of the functions, shortened the comparator functions, and synced up with the chain's paradigms as much as possible without blowing what I think are typical frontend expectations.

I'm sure we will debate naming. Let's start with the idea that Quantity and Decimal are *terribly* generic terms to be used for what will always be an asset amount. The abstractions also work slightly differently on the frontend, so I didn't faithfully replicate the backend concepts verbatim. Note also that there's the wit decimal-to-u64 naming to take into account. So there wasn't 100% 1:1 mapping that I thought made sense. But I've put a lengthy comment at the top of the quantity.ts file to walk through the definitions and reference to the chain and wit ideas.

### DIFF
--- a/packages/shared-ui/components/form/field-token-amount.tsx
+++ b/packages/shared-ui/components/form/field-token-amount.tsx
@@ -168,6 +168,6 @@ const zTokenAmount = ({
             // Skip when the input doesn't parse cleanly; the precision
             // refinement above already produced a more specific error.
             if (input === null) return true;
-            return !balance.isLessThan(input); // input <= balance
+            return !balance.lt(input); // input <= balance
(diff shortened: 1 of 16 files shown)


James-Mart:
> Decimal [is a] terribly generic term to be used for what will always be an asset amount.

Decimal is a generic wrapper for holding decimal numeric values. Unrelated to assets.

PR #1851, opened 2026-05-01
Improve user prompt implementation

### LINKED ISSUE #1846: User prompts depend on returning 'Result<Err>'
We currently have a design problem related to plugin user-prompts. 

Plugins that need user input (via `user-prompt`), depend on returning `Result<Err>`. This is an anti-pattern. 

Standard advice for using Results vs panics would imply:
* “this might reasonably fail” --> Result
* “this must not fail unless there’s a bug / invariant violation” --> panic is fine

But in our case, it's possible that some downstream plugin is trying to get user input about something, and since that is currently expressed as returning a `Result<Err>`, we violate standard error handling expectations. 

A user prompt in some platforms is synchronous. In the web, it should cause the immediate teardown of the plugin callchain rather than propagating an error back through the callstack, which forces a Result<> chain to cascade through all the plugin APIs and requires that consumers of the Results never unwrap them.

James-Mart:
This closes https://github.com/gofractally/psibase/issues/1846

If a user prompt was requested from a plugin, the web host (supervisor) was throwing a RecoverableError payload (`host:types/types::Error`). This could be caught by any upstream plugins that used the very common `-> Result<_, Error>` return type from their plugin function. This is odd and asymmetric with how it would work on other platforms with synchronous user prompts.

The supervisor now instantly tears down the whole plugin callstack (essentially a panic), instead of returning an error variant down the whole callstack.


packages/user/Supervisor/ui/src/supervisor.ts
```diff
@@ -426,22 +437,26 @@ export class Supervisor implements AppInterface {
...
-                    newError = new PluginErrorObject(err.producer, err.message);
-                }
-                this.replyToParent(id, newError);
+            let result: any;
+            if (this.promptSignal === "prompt") {
+                result = new RedirectErrorObject(
+                    { service: "host", plugin: "prompt" },
+                    "user_prompt_request",
+                );
+                this.promptSignal = "none";
```
swatanabe: It would be better to reset this in the finally clause. However, note that a prompt from the plugin calls in `preload` is likely to behave very oddly.

packages/user/Supervisor/ui/src/supervisor.ts
```diff
@@ -276,6 +277,16 @@ export class Supervisor implements AppInterface {
...
     }
 
+    requestPrompt(): void {
+        if (isEmbedded) {
+            this.promptSignal = "embedded-error";
+        } else {
+            this.promptSignal = "prompt";
+        }
+        this.plugins.disposeAll();
+        throw null;
```
swatanabe: Is `throw null` idiomatic in js? I would throw some kind of `NonRecoverableError` that includes the `promptSignal`.

packages/user/Supervisor/ui/src/supervisor.ts
```diff
@@ -276,6 +277,16 @@ export class Supervisor implements AppInterface {
         );
     }
 
+    requestPrompt(): void {
+        if (isEmbedded) {
+            this.promptSignal = "embedded-error";
+        } else {
+            this.promptSignal = "prompt";
+        }
+        this.plugins.disposeAll();
```
swatanabe: Is this needed? The top level entry point should always call `disposeAll`, right?

packages/user/Supervisor/ui/src/supervisor.ts
James-Mart: It's valid syntax. The reason I chose `null` was specifically because I don't believe there's any valid wit representation of the null value, and therefore I can be certain that no upstream plugin's transpiled js bindings could intercept in their catch() handlers that are generated for functions that return a wit `Result`.

packages/user/Supervisor/ui/src/supervisor.ts
James-Mart: > It would be better to reset this in the finally clause.

No longer applicable.

> note that a prompt from the plugin calls in preload is likely to behave very oddly.

Improved. Prompts during preload are explicitly rejected.

packages/user/Supervisor/ui/src/supervisor.ts
James-Mart: Okay I reworked this. I learned that JCO rethrows an Error as long as it doesn't have an own `payload` property, so I now throw a custom `Error` object instead of null.

packages/user/Supervisor/ui/src/supervisor.ts
James-Mart: Not needed. Cleaned up

swatanabe (approved):

PR #1853, opened 2026-05-01
Config UI staged tx improvement

James-Mart:
Adds a button that spawns a modal that allows for simpler review of staged tx in config.

image

image


brandonfancher (approved):
UI code LGTM

packages/user/Packages/plugin/src/lib.rs
```diff
@@ -178,6 +178,48 @@ impl TryFrom<types::PackageOpFull> for psibase::PackageOpFull<Cursor<Vec<u8>>> {
...
+        return Ok("null".to_string());
+    }
+
+    let raw = raw_data_hex
+        .parse::<Hex<Vec<u8>>>()
+        .map_err(|_| ErrorType::UnpackActionError("invalid hex for raw data".to_string()))?
+        .0;
+
+    let schema_json =
+        Server::get_json(&format!("/schema?service={}", service_account)).map_err(|e| {
```
swatanabe: Does this fetch the schema separately for each action? I don't recall setting up cache control headers for the schema. This is more likely to be a problem for package installation than for any other staged transactions, due to the large number of actions.

James-Mart: Added session cache for the schema

swatanabe: I don't know that the session cache is the right place. The schemas can change. Installing packages is precisely what will cause them to change. This also brings up another problem, which is that we may need schemas that are not yet on the chain, because they are part of the packages being installed.

James-Mart: I propose 
* I revert the session cache and deal with the performance hit. Better caching can be added later

> we may need schemas that are not yet on the chain, because they are part of the packages being installed.

This really just refers to the post-install script for a new package, right?

swatanabe: Yes, it should only be the postinstall script. Other actions that are used by install are always present and can't easily be changed without breaking everything. The postinstall script can call actions from other services, though.

swatanabe: An example is installing `TokenUsers`, which also installs `FaucetTok` at the same time.

James-Mart: Maybe I'll just attempt to fetch the associated schema, and if the schema can't be fetched I'll just show an error? This is an edge case I don't feel like fully solving now unless you have an easy solution.

swatanabe: It's much easier to solve by processing the entire transaction at once, rather than an action at a time.

PR #1855, opened 2026-05-04
Add AccountNumber asscociated constants MIN & MAX

Velua:
Adds to `AccountNumber`
```
impl AccountNumber {
    pub const MIN:: Self = AccountNumber { value: 0 };
    pub const MAX: Self = AccountNumber { value: u64::MAX };
}
```


rust/psibase/src/account_number.rs
```diff
@@ -136,66 +139,51 @@ mod tests {
...
-            AccountNumber::new(0)
-        );
-        assert_eq!(
-            AccountNumber::from_str("1234").unwrap(),
-            AccountNumber::new(0)
-        );
+        assert_eq!(AccountNumber::from_str("9").unwrap(), AccountNumber::MIN);
+        assert_eq!(AccountNumber::from_str("3").unwrap(), AccountNumber::MIN);
+        assert_eq!(AccountNumber::from_str("1abc").unwrap(), AccountNumber::MIN);
+        assert_eq!(AccountNumber::from_str("1234").unwrap(), AccountNumber::MIN);
```
swatanabe: `MIN` isn't really correct here. These tests produce the "invalid" value, which happens to be the same as `MIN`.

rust/psibase/src/service.rs
```diff
@@ -5,10 +5,10 @@ use serde::de::DeserializeOwned;
 use serde::Serialize;
 
 #[cfg(target_family = "wasm")]
-static mut SERVICE: AccountNumber = AccountNumber::new(0);
+static mut SERVICE: AccountNumber = AccountNumber::MIN;
```
swatanabe: The same applies here. `0` expresses the meaning better than `MIN`.

swatanabe (approved):

James-Mart (approved):

PR #1860, opened 2026-05-08
add a function in profiles to check if the caller has read permissions

James-Mart:
Adds a `has_auth` function to the permissions plugin. Returns true if the authority has already been granted, or false if the authority has not been granted (and any attempt to use a function that requires such authority may result in a user prompt).


brandonfancher (approved):
🎉

Velua (approved):

PR #1862, opened 2026-05-10
Manual debiting by default

Velua:
Currently all services / accounts need to enable manual debiting to stop auto debiting behaviour, which is constantly set in the `init` actions of most of our services. 

This inverts behaviour, replacing the `manual_debit` flag to `auto_debit` in the `Tokens` and `NFT` service.
 
Services now no longer need to reach out to the tokens and NFT services to enable manual debiting as auto debit is off by default.


rust/psibase/src/services/tokens/mod.rs
```diff
@@ -228,7 +228,7 @@ mod service {
     ///                See `Configurations` for details
     ///
     /// Configurations:
-    /// * 0: manual_debit       - If enabled, any credits of this token must be manually debited by
+    /// * 0: manual_debit       - If enabled, any credits of this token will be automatically debited by
```
swatanabe: The name `manual_debit` needs to be changed (3 other places below, as well).

packages/user/Tokens/service/src/lib.rs
```diff
@@ -345,7 +340,7 @@ pub mod service {
     /// Credit tokens to a debitor (recipient).
     ///
     /// On credit, tokens are typically automatically debited by the debitor. However,
-    /// if the debitor has enabled `manual_debit`, then the tokens will be placed in an intermediate
+    /// if the debitor has enabled `auto_debit`, then the tokens will be placed in an intermediate
     /// "shared balance".
```
swatanabe: This is now backwards.

swatanabe (commented):
Is this the default that we want for regular user accounts? I don't think services should dictate the default behavior for this. Also, if a service depends on manual debit, it's still a good idea to set it in init, even if it is the default.

Velua:
> Is this the default that we want for regular user accounts? I don't think services should dictate the default behavior for this. Also, if a service depends on manual debit, it's still a good idea to set it in init, even if it is the default.

Yes, if services shouldn't influence default behavior, should users? 
It seems weird that services feel a need to reach out to other services like `Nft` and `Tokens` because they present themselves a liability by default.

James-Mart:
> Is this the default that we want for regular user accounts?

I think it should be. I've gone back and forth on it but I increasingly think it should be the default, and am even open to deleting auto debit entirely, but that's probably more controversial.

swatanabe:
> > Is this the default that we want for regular user accounts? I don't think services should dictate the default behavior for this. Also, if a service depends on manual debit, it's still a good idea to set it in init, even if it is the default.
> 
> Yes, if services shouldn't influence default behavior, should users?

That was what I was trying to get at. The default should be set for the needs of ordinary users. I don't really have an opinion on what the correct default is, but I think the justification for this change in the PR description is completely wrong.

> It seems weird that services feel a need to reach out to other services like `Nft` and `Tokens` because they present themselves a liability by default.

It would only be weird for services that don't otherwise use `Nft` or `Tokens`. Services that do use manual debit (or, in general, rely on any configurable behavior of another service) should set that configuration at install time regardless of what the default is.

James-Mart:
> Services that do use manual debit (or, in general, rely on any configurable behavior of another service) should set that configuration at install time regardless of what the default is.

 Tokens can have its config eliminated, but afaik this implies all the other services whose init actions were modified should be reverted, and should still explicitly turn off auto-debit, even though that's the default.
 
On the defensive side, but probably wise.

CMakeLists.txt
```diff
@@ -1040,7 +1040,8 @@ psibase_package(
     NAME TokenUsers
     VERSION ${PSIBASE_VERSION}.0
     DESCRIPTION "Creates accounts alice & bob, and gives each some system tokens. Not for production use."
-    PACKAGE_DEPENDS "AccountsService(^${PSIBASE_VERSION}.0)" "FaucetTok(^${PSIBASE_VERSION}.0)" 
+    PACKAGE_DEPENDS "AccountsService(^${PSIBASE_VERSION}.0)" "FaucetTok(^${PSIBASE_VERSION}.0)" "Tokens(^${PSIBASE_VERSION}.0)" 
+    ACCOUNTS alice bob
```
swatanabe: This sets the auth for `alice` and `bob` to `root`. They were created in the postinstall to ensure that they used `auth-any`.

packages/user/Tokens/src/TokenUsers.json
```diff
@@ -1,22 +1,20 @@
 [
     {
-        "sender": "accounts",
-        "service": "accounts",
-        "method": "newAccount",
+        "sender": "alice",
+        "service": "tokens",
+        "method": "setUserConf",
```
swatanabe: If we've decided that manual debit is the right default for user accounts, I don't think `TokenUsers` should change it.

programs/psinode/tests/test_txqueue.py
```diff
@@ -8,6 +8,13 @@
 from psinode import Action, Transaction, TransactionError, PrivateKey
 from services import Accounts, AuthSig, Tokens, Transact
 
+FAUCET_DEBIT = {"token_id": 1, "creditor": "faucet-tok", "amount": {"value": 1000000000}, "memo": ""}
+
+def accept_faucet_tokens(node, *accounts, keys=[], wait_node=None):
+    for account in accounts:
+        node.push_action(account, 'tokens', 'debit', FAUCET_DEBIT, keys=keys)
+    (wait_node or getattr(node, 'api', node)).wait(new_block())
```
swatanabe: There's no need for `wait_node`. Just call `accept_faucet_tokens` right after boot, and pass in `a` as `node`.

swatanabe (approved):

PR #1863, opened 2026-05-12
FractalCore: More sensible Profiles/contacts integration

brandonfancher:
Issue #1768

Before this, if you navigated to a screen listing members in FractalCore for the first time, you would automatically get redirected to a permissions prompt requesting access to the Profiles (contacts) app, so that the FractalCore UI could display friendly nicknames from your contacts for members. This was jarring for the user.

This introduces a hook that reads (without a prompt) whether or not the app has been granted Profiles permissions. If it has not, we now display a button in strategic places in the app (as a `CardAction` at the top of members/applicants table cards and the evaluation deliberation screen) to allow the user to manually enable contacts/Profiles integration.

It also ensures that the user is redirected to the page they came from after the prompt is resolved.

image

image

### DIFF
--- a/packages/user/FractalCore/ui/src/lib/paths.ts
+++ b/packages/user/FractalCore/ui/src/lib/paths.ts
@@ -5,5 +5,13 @@ export const paths = {
         evaluations: (guild: Account) => `/guild/${guild}/evaluations/upcoming`,
         evaluationGroup: (guild: Account, groupNumber: number) =>
             `/guild/${guild}/evaluations/group/${groupNumber}`,
+        membership: {
+            applicants: (guild: Account) =>
+                `/guild/${guild}/membership/applicants`,
+            members: (guild: Account) => `/guild/${guild}/membership/members`,
+        },
+    },
+    fractal: {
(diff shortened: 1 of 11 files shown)


brandonfancher:
Closes #1768

Velua (approved):

packages/user/FractalCore/ui/src/lib/paths.ts
```diff
@@ -5,5 +5,13 @@ export const paths = {
...
         evaluationGroup: (guild: Account, groupNumber: number) =>
             `/guild/${guild}/evaluations/group/${groupNumber}`,
+        membership: {
+            applicants: (guild: Account) =>
+                `/guild/${guild}/membership/applicants`,
+            members: (guild: Account) => `/guild/${guild}/membership/members`,
+        },
+    },
+    fractal: {
+        members: () => `/fractal/members`,
```
James-Mart: I think this is the wrong path, should just be `/members`

brandonfancher: Good catch. Thanks!

PR #1868, opened 2026-05-13
add has-auth wrappers into psibase-plugin trust lib

James-Mart:
Fixes a dead code warning due to the `has_auth` integration into `psibase-plugin` being incomplete

### DIFF
--- a/rust/psibase_plugin/src/wasm/trust.rs
+++ b/rust/psibase_plugin/src/wasm/trust.rs
@@ -133,14 +133,14 @@ pub trait TrustConfig {
     where
         Self: Sized,
     {
-        has_auth_with_whitelist::<Self>(level, &[])
+        has_auth_with_whitelist(level, &[])
     }
 
     fn has_auth_with_whitelist(level: TrustLevel, whitelist: &[&str]) -> bool
     where
         Self: Sized,
     {
-        has_auth_with_whitelist::<Self>(level, whitelist)
+        has_auth_with_whitelist(level, whitelist)
     }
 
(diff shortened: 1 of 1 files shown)


rust/psibase_plugin/src/wasm/trust.rs
```diff
@@ -138,6 +152,15 @@ pub fn authorized<T: TrustConfig>(level: TrustLevel, fn_name: &str) -> Result<bo
     authorized_with_whitelist::<T>(level, fn_name, &[])
 }
 
+pub fn has_auth<T: TrustConfig>(level: TrustLevel) -> bool {
+    has_auth_with_whitelist::<T>(level, &[])
+}
+
+pub fn has_auth_with_whitelist<T: TrustConfig>(level: TrustLevel, whitelist: &[&str]) -> bool {
```
swatanabe: Is `T` needed here? It's for display to the user, but this will never cause a prompt. If it's for symmetry with `authorized`, then why not `fn_name`, too?

rust/psibase_plugin/src/wasm/trust.rs
```diff
@@ -129,6 +129,20 @@ pub trait TrustConfig {
         authorized_with_whitelist::<Self>(level, fn_name, whitelist)
     }
 
+    fn has_auth(level: TrustLevel) -> bool
+    where
+        Self: Sized,
```
swatanabe: I think it would be better to put a `?Sized` bound on the module scope function.

swatanabe (approved):

PR #1878, opened 2026-05-18
Add 'create-premium' to Accounts plugin

cool-ant:
for upcoming integration of PremAccounts into Homepage as well as allowing creation of premium accounts during Account Creation, e.g., during invite acceptance.

`create-premium` wraps `buy`, `claim`, and auth config of a premium account

Tracking ~requirements at
https://github.com/gofractally/psibase/issues/1892


packages/system/Accounts/plugin/accounts/src/errors.rs
```diff
@@ -13,5 +13,7 @@ plugin_error! {
     InvalidPrefix() => "Prefix must be a-z, 1 - 9 chars in length",
     AccountNotFound(account: String) => "Account not found: {account}",
     UnsupportedAuthService(service: String) => "Account uses unsupported auth service: {service}",
-    AuthorizationFailed(account: String) => "Key cannot authorize account: {account}"
+    AuthorizationFailed(account: String) => "Key cannot authorize account: {account}",
+    PremiumNameLengthNotOffered() => "Not a premium name",
+    PremiumNameMarketDisabled() => "Premium names of this length are disabled",
```
James-Mart: Identifiers `PremiumNameMarketDisabled` and `PremiumNameLengthNotOffered` seem the same to me. And surprising that `PremiumNameLengthNotOffered` has the message, "Not a premium name"

packages/system/Accounts/plugin/accounts/src/interfaces/prompt.rs
```diff
@@ -92,6 +100,34 @@ impl Prompt for AccountsPlugin {
         Ok(private_key)
     }
 
+    fn create_premium(account_name: String, slippage_pct: u8) -> Result<String, Error> {
+        assert_eq!(Client::get_sender(), Client::get_receiver());
+
+        if account_name.len() >= 10 {
```
James-Mart: Named constant might be preferable

packages/system/Accounts/src/Accounts.cpp
```diff
@@ -83,7 +83,6 @@ namespace SystemService
       {
          check(!strName.starts_with("x-"),
                "The 'x-' account prefix is reserved for infrastructure providers");
-         check(strName.length() >= 10, "account name must be at least 10 characters: " + strName);
```
James-Mart: This shouldn't be removed... just check if it's >= the new lower bound for free names (8 char)

packages/system/Accounts/plugin/accounts/src/interfaces/prompt.rs
```diff
@@ -22,6 +30,23 @@ impl Prompt for AccountsPlugin {
...
+
+        // assert(systemToken exists)
+        // assert(PremAccounts installed)
+
+        if let Some(can_create_account) = Invites::get_active_invite() {
+            return can_create_account;
+        }
+
+        false
+    }
```
James-Mart: Can be simpler
```rust
fn can_create_premium_account() -> bool {
    assert_eq!(Client::get_sender(), Client::get_receiver());
    Self::is_logged_in()
}
```

packages/system/Accounts/plugin/accounts/src/interfaces/prompt.rs
```diff
@@ -92,6 +117,34 @@ impl Prompt for AccountsPlugin {
         Ok(private_key)
     }
 
+    fn create_premium(account_name: String, slippage_pct: u8) -> Result<String, Error> {
+        assert_eq!(Client::get_sender(), Client::get_receiver());
+
```
James-Mart, line 105: Could return a nice error here if `!can_create_premium_account()`

packages/system/Accounts/plugin/accounts/src/interfaces/prompt.rs
```diff
@@ -92,6 +117,34 @@ impl Prompt for AccountsPlugin {
...
+        if account_name.len() >= 10 {
+            return Self::create_account(account_name);
+        }
+
+        let ask = premium_market_ask(&account_name)?;
```
brandonfancher: If I'm thinking about this correctly, we shouldn't get the fresh ask price here; it should be passed in along with slippage. Fetching a fresh price here negates the purpose of a UI-drive slippage percent.

When the user creates the name, the UI is displaying to them the latest cost it fetched (and informs them of the 5% slippage). If the price goes up 20% between the time they sees that number and the time this function kicks off, that represents not a 5% slippage, but a _20-25% actual_ slippage.

Instead, the UI should supply both the price and the slippage percent here. That empowers the UI to ensure that the user isn't charged more than the price it displays to them.

brandonfancher (changes requested):

packages/system/Accounts/plugin/accounts/src/interfaces/prompt.rs
brandonfancher: Alternatively, you can do what `PremAccounts::buy` does and take a `max_cost` instead of a `slippage_pct`.

packages/system/Accounts/plugin/accounts/src/interfaces/prompt.rs
```diff
@@ -22,6 +67,40 @@ impl Prompt for AccountsPlugin {
         false
     }
 
+    fn can_create_premium_account() -> bool {
+        assert_eq!(Client::get_sender(), Client::get_receiver());
+
+        if Self::is_logged_in() {
+            return true;
```
brandonfancher: Does this early return mean we don't get the at-least-one-market-is-enabled check?

cool-ant: oh! Great catch. That's a bug.

cool-ant: fix pushed

packages/system/Accounts/plugin/accounts/src/interfaces/prompt.rs
cool-ant: Updating this now-obsolete comment: I could combine some of these if statement together if you like, but I think this reads nice and cleanly.

packages/system/Accounts/plugin/accounts/src/interfaces/prompt.rs
brandonfancher: Working now. Thanks!

brandonfancher:
@cool-ant `Accounts::can_create_premium_account` is working. Thank you.

Suggestion:

`can_create_premium_account` should live in the `PremAccounts` plugin. Then `Accounts::can_create_premium_account` would just call it under the hood, similar to how `Accounts::create_premium` calls `PremAccounts::buy` and `PremAccounts::claim`.

In that way, `can_create_premium_account` becomes something _any_ app can call. (And it would make sense that `PremAccounts` is the authority on whether someone can create a premium account, _not_ `Accounts`.)

My use case: I'd also like to call it from the `Homepage` UI for deciding whether to render the Buy screen. Right now, it's still querying premium prices and checking that at least one is enabled...but as things mature, `PremAccounts` may add other prerequisites, and in that event, `Homepage` won't be doing the right checks anymore!

brandonfancher (approved):
LGTM!

packages/shared-ui/constants.ts
```diff
@@ -0,0 +1,2 @@
+export const MIN_ACCOUNT_NAME_LENGTH = 1;
+export const MAX_ACCOUNT_NAME_LENGTH = 18;
```
Velua: We got a shared zAccount file don't we? Does that export min and max account lengths? Otherwise we're doubling up.

packages/system/Accounts/src/Accounts.cpp
```diff
@@ -83,7 +83,7 @@ namespace SystemService
...
+         check(strName.length() >= 8, "account name must be at least 10 characters: " + strName);
```
Velua: Should the message say 8 instead of 10?

packages/shared-ui/constants.ts
brandonfancher: I have a follow-on PR that centralizes/cleans this up. It's based on this branch.

packages/user/PremAccounts/service/Cargo.toml
```diff
@@ -12,6 +12,24 @@ data = [{ src = "../ui/dist", dst = "/" }]
...
+    { sender = "prem-accounts", service = "prem-accounts", method = "create", data = { length = 5, initial_price = { value = 1000 }, target = 10, floor_price = { value = 100 }, increase_ppm = 50000, decrease_ppm = 50000 } },
+    { sender = "prem-accounts", service = "prem-accounts", method = "create", data = { length = 6, initial_price = { value = 1000 }, target = 10, floor_price = { value = 100 }, increase_ppm = 50000, decrease_ppm = 50000 } },
+    { sender = "prem-accounts", service = "prem-accounts", method = "create", data = { length = 7, initial_price = { value = 1000 }, target = 10, floor_price = { value = 100 }, increase_ppm = 50000, decrease_ppm = 50000 } },
+    { sender = "prem-accounts", service = "prem-accounts", method = "create", data = { length = 8, initial_price = { value = 1000 }, target = 10, floor_price = { value = 100 }, increase_ppm = 50000, decrease_ppm = 50000 } },
+    { sender = "prem-accounts", service = "prem-accounts", method = "create", data = { length = 9, initial_price = { value = 1000 }, target = 10, floor_price = { value = 100 }, increase_ppm = 50000, decrease_ppm = 50000 } },
```
Velua: Aren't these accounts of over 9 free? Why are we creating markets for them?

packages/user/PremAccounts/service/Cargo.toml
```diff
@@ -12,6 +12,24 @@ data = [{ src = "../ui/dist", dst = "/" }]
...
+    { sender = "prem-accounts", service = "prem-accounts", method = "create", data = { length = 4, initial_price = { value = 1000 }, target = 10, floor_price = { value = 100 }, increase_ppm = 50000, decrease_ppm = 50000 } },
+    { sender = "prem-accounts", service = "prem-accounts", method = "create", data = { length = 5, initial_price = { value = 1000 }, target = 10, floor_price = { value = 100 }, increase_ppm = 50000, decrease_ppm = 50000 } },
+    { sender = "prem-accounts", service = "prem-accounts", method = "create", data = { length = 6, initial_price = { value = 1000 }, target = 10, floor_price = { value = 100 }, increase_ppm = 50000, decrease_ppm = 50000 } },
+    { sender = "prem-accounts", service = "prem-accounts", method = "create", data = { length = 7, initial_price = { value = 1000 }, target = 10, floor_price = { value = 100 }, increase_ppm = 50000, decrease_ppm = 50000 } },
+    { sender = "prem-accounts", service = "prem-accounts", method = "create", data = { length = 8, initial_price = { value = 1000 }, target = 10, floor_price = { value = 100 }, increase_ppm = 50000, decrease_ppm = 50000 } },
```
Velua: Is this meant for production use? Ideally we don't hard code prices as we've no idea what `1000` is without context and instead let the user initialize this from a UI which will just do a for each on 1 - 8 at prices the user is happy with.

packages/system/Accounts/plugin/accounts/src/interfaces/prompt.rs
```diff
@@ -92,6 +99,36 @@ impl Prompt for AccountsPlugin {
...
+        Transact::set_propose_latch(Some(&account_name))?;
+        AuthSig::actions::set_key(&keypair.public_key)?;
+        Transact::add_action_to_transaction(
+            AccountsService::action_structs::setAuthServ::ACTION_NAME,
+            &AccountsService::action_structs::setAuthServ {
+                authService: auth_sig::Wrapper::SERVICE,
+            }
+            .packed(),
+        )?;
+        Transact::set_propose_latch(None)?;
```
Velua, line 128: Is setting the propose latch back to `None` significant here? Maybe worth a comment

packages/system/Accounts/plugin/accounts/src/helpers.rs
```diff
@@ -83,3 +86,83 @@ pub fn generate_account(prefix: Option<String>) -> Result<String, CommonTypes::E
...
+        .iter()
+        .find(|s| s.length == length);
+
+    let Some(st) = status else {
+        return Err(NameLengthUnavailable().into());
+    };
+
+    if !st.enabled {
+        return Err(NameLengthUnavailable().into());
+    };
```
Velua:
```suggestion
    response
        .data
        .market_params
        .iter()
        .find(|s| s.length == length)
        .filter(|s| s.enabled)
        .ok_or(NameLengthUnavailable())?;
```

packages/system/Accounts/plugin/accounts/src/helpers.rs
```diff
@@ -83,3 +86,83 @@ pub fn generate_account(prefix: Option<String>) -> Result<String, CommonTypes::E
...
+        .iter()
+        .find(|p| p.length == length);
+
+    let Some(row) = row else {
+        return Err(
+            QueryError("missing price row for configured premium name market".into()).into(),
+        );
+    };
+
+    Ok(row.price.to_string())
```
Velua:
```suggestion
    response
        .data
        .current_prices
        .iter()
        .find(|p| p.length == length)
        .map(|row| row.price.to_string())
        .ok_or(QueryError("missing price row for configured premium name market".into()).into())
```

packages/system/Accounts/plugin/accounts/src/helpers.rs
```diff
@@ -83,3 +86,83 @@ pub fn generate_account(prefix: Option<String>) -> Result<String, CommonTypes::E
...
+    let ask_u64 = ask_dec.quantity.value;
+    let factor = 100u64 + slippage_pct as u64;
+    let max_u64 = ask_u64
+        .checked_mul(factor)
+        // ensure rounding always goes *up* so calculated amount is always *at least* as much as if calculated as a float.
+        .and_then(|v| v.checked_add(99))
+        .map(|v| v / 100)
+        .ok_or_else(|| QueryError("max cost overflow".into()))?;
+    Ok(Decimal::new(max_u64.into(), ask_dec.precision).to_string())
+}
```
Velua: It's a little awkward reviewing this while it's without further context but I'm generally against mixing the maths with this decimal to `u64` behavior, feel like it should just be concerned with receiving a `Quantity` if this is an internal helper. 

Also I think we should be expanding to a `u128`, otherwise I think `checked_mul` in `u64` will reject inputs where the final result would still fit.

Velua (dismissed):

packages/user/PremAccounts/service/Cargo.toml
cool-ant: decided to default enable prem-accounts for all potential markets

packages/user/PremAccounts/service/Cargo.toml
cool-ant: it is meant for production; new chain startup would involve the chain team setting these values

cool-ant: Doing this config in text form like this will be a *lot* easier for chain startup than 1) modiying code and/or 2) having to set config param for many different markets via a UI

packages/system/Accounts/plugin/accounts/src/interfaces/prompt.rs
cool-ant: Just an explicit flush of the actions. following the pattern from elsewhere in the codebase

packages/system/Accounts/plugin/accounts/src/helpers.rs
cool-ant: this was dead code. removed.

packages/system/Accounts/plugin/accounts/src/interfaces/prompt.rs
James-Mart: Yeah in this case it's not strictly necessary, but doesn't hurt.

packages/system/Accounts/plugin/accounts/src/interfaces/prompt.rs
James-Mart: > Updating this now-obsolete comment

What comment are you talking about?

> I could combine some of these if statement together

What if statements are you talking about?



I guess this link got broken because the fn was moved between files:
See `can_create_premium_account` in `packages/user/PremAccounts/plugin/src/lib.rs`
https://github.com/gofractally/psibase/pull/1878/changes#diff-ac54066d1e94a914ad9cc9b8c9cf8371e20565502d4fa8cced08a75f235e3654R113

rust/psibase/src/services/accounts.rs
```diff
@@ -5,6 +5,9 @@ use crate::{Pack, ToSchema, Unpack};
 use serde::{Deserialize, Serialize};
 use serde_aux::field_attributes::deserialize_number_from_string;
 
+pub const MIN_ACCOUNT_NAME_LENGTH: u8 = 1;
+pub const MAX_ACCOUNT_NAME_LENGTH: u8 = 18;
+pub const MIN_FREE_ACCOUNT_NAME_LENGTH: u8 = 8;
```
James-Mart: 1. These `MIN/MAX_ACCOUNT_NAME_LENGTH` constants are referring to the min/max allowable by the encoding rules, right? If so, then such constants (if needed) probably ought to live alongside the encoding algo, not in the accounts service.
2. The name `MIN_FREE_ACCOUNT_NAME_LENGTH` assumes that account lengths less than 8 are not free, but that is not up to the accounts service.

packages/user/PremAccounts/service/src/lib.rs
```diff
@@ -207,11 +205,15 @@ pub mod service {
...
+
         let auctions_table = AuctionsTable::new();
-        check_none(
-            auctions_table.get_index_pk().get(&length),
-            "market already exists",
-        );
+        if auctions_table.get_index_pk().get(&length).is_some() {
+            // market already exists; return (success)
+            // idempotent for postinstall
+            return;
```
James-Mart: This should only be idempotent if the args match the existing market config, otherwise I'd expect an error

packages/user/PremAccounts/service/src/lib.rs
```diff
@@ -207,11 +205,15 @@ pub mod service {
 
         check_market_length(length);
 
+        // check(systemToken exists (inline action on Tokens service))
```
James-Mart: Did you forget to add the actual inline action call here?

packages/system/Accounts/src/Accounts.cpp
```diff
@@ -83,7 +83,7 @@ namespace SystemService
       {
          check(!strName.starts_with("x-"),
                "The 'x-' account prefix is reserved for infrastructure providers");
-         check(strName.length() >= 10, "account name must be at least 10 characters: " + strName);
+         check(strName.length() >= 8, "account name must be at least 8 characters: " + strName);
```
James-Mart: If you add the MIN constant to C++ as well, then you could use it here

packages/user/PremAccounts/service/Cargo.toml
James-Mart: I disagree. The config app is filled with things that the infra providers (or fractal council) need to go through and configure. In production, these have to be set individually and explicitly by the prods, with prices relative to their system token supply

For development, a couple weeks ago we discussed having a separate package (similar to TokenUsers) that could configure these markets automatically, so that a development boot could have them out-of-the-box, but that would be a separate package postinstall, not the PremAccounts package itself

packages/user/PremAccounts/plugin/src/lib.rs
```diff
@@ -89,7 +90,38 @@ impl MarketAdmin for PremAccountsPlugin {
...
+        if overview.market_params.is_empty() && overview.current_prices.is_empty() {
+            return false;
+        }
+
+        overview.market_params.iter().any(|market| {
+            market.enabled
+                && overview
+                    .current_prices
+                    .iter()
+                    .any(|price| price.length == market.length)
```
James-Mart: Is it possible for this 'any' to return false?

packages/user/PremAccounts/service/src/lib.rs
cool-ant: yup. added.

packages/user/PremAccounts/plugin/src/lib.rs
cool-ant: only in highly distorted scenarios. I simplified it, given the reliable relationship between .enabled and .price.length

packages/user/PremAccounts/service/Cargo.toml
cool-ant: Ok, as discussed, I've moved market config out to a DevDefault-included package I named PremMarkets.

brandonfancher (dismissed):
- We still need the `marketParams` query to return initial price
- I'm unable to buy names; getting errors. This is due to various instances herein of `prem-accounts` that should be updated to `prem-accts`. (Changes were required for me in the `Tokens`, `AuthSig` and `Accounts` plugins and in `Accounts.cpp` to get it working, but those changes should be made in this PR, not mine.)
- `DevNameMarkets.json` needs attention too (account name? and removal of 11-18 lengths)

packages/system/Accounts/plugin/accounts/wit/prompt.wit
```diff
@@ -42,6 +42,11 @@ interface prompt {
     ///   is met, then the attempt to create an account will fail.
     create-account: func(account-name: string) -> result<string, error>;
 
+    /// Purchases a premium account name, claims it, configures auth-sig, and returns the new account's private key.
+    ///
+    /// The account is not imported onto this device.
+    create-premium: func(account-name: string, max-cost: string) -> result<string, error>;
```
Velua, line 48: Are we still going with "premium"?

packages/user/Config/plugin/wit/world.wit
```diff
@@ -16,13 +16,15 @@ interface packaging {
...
-interface prem-accounts {
+interface name-market {
     use host:types/types.{error};
+    use name-market:plugin/market-admin.{market-config};
 
-    create: func(length: u8, initial-price: string, target: u32, floor-price: string, increase-ppm: u32, decrease-ppm: u32) -> result<_, error>;
-    configure: func(length: u8, window-seconds: u32, target: u32, floor-price: string, increase-ppm: u32, decrease-ppm: u32) -> result<_, error>;
-    enable: func(length: u8) -> result<_, error>;
-    disable: func(length: u8) -> result<_, error>;
+    /// Batch create or update premium name markets.
```
Velua:
```suggestion
    /// Batch create or update name markets.
```

packages/user/NameMarket/service/src/lib.rs
```diff
@@ -20,6 +20,7 @@ pub mod tables {
         pub length: u8,
         pub nft_id: u32,
         pub enabled: bool,
+        pub initial_price: u64,
```
Velua: What's the point of storing this? And why `u64` and not `Quantity`?

packages/user/NameMarket/plugin/src/lib.rs
```diff
@@ -0,0 +1,364 @@
...
+            }
+        }
+
+        Ok(())
+    }
+}
+
+impl Api for NameMarketPlugin {
+    #[psibase_plugin::authorized(Medium, whitelist = ["accounts", "homepage"])]
+    fn can_create_premium_account() -> bool {
```
Velua:
```suggestion
    fn can_create_account() -> bool {
```

packages/user/NameMarket/service/src/tests.rs
```diff
@@ -216,16 +213,28 @@ mod tests {
...
             )
+            .get()?;
+
+        // --- duplicate create(7) with different params fails ---
+        let err = NameMarket::push_from(&chain, alice)
+            .create(
+                7,
+                Quantity::from(2000u64),
+                10,
+                Quantity::from(100u64),
```
Velua, line 226: I find into easier to read but that's a nit.
```suggestion
                100.into(),
```

packages/system/Accounts/plugin/accounts/wit/prompt.wit
James-Mart: In this particular context, yes, because this is genuinely referring to premium names.

packages/user/NameMarket/service/src/lib.rs
cool-ant: displayed in the frontend and so avaialble by query (and unavailable in DiffAdjust via query).
you're right; swapped

packages/user/NameMarket/query-service/src/lib.rs
```diff
@@ -1,28 +1,36 @@
...
+    use name_market::Wrapper as NameMarketService;
     use psibase::services::diff_adjust::{RateLimitTable, Wrapper as DiffAdjust};
     use psibase::services::tokens::{Decimal, Quantity, Wrapper as TokensWrapper};
     use psibase::*;
     use serde::Deserialize;
     use serde_aux::field_attributes::deserialize_number_from_string;
 
+    const PPM_PER_PCT: u32 = 10_000;
+
+    fn ppm_to_pct(ppm: u32) -> u8 {
```
James-Mart: _Non-pr-blocking_
Sometimes when you have a need for a utility like this in in both queries and plugins, it makes sense to upstream it into the rust bindings file and link to it as needed, so it's defined once.

packages/user/NameMarket/service/src/tests.rs
```diff
@@ -90,20 +88,38 @@ mod tests {
         Ok(())
     }
 
-    /// Full PremAccounts integration on one chain (serial steps; avoids parallel `psitest` races).
-    #[psibase::test_case(packages("Tokens", "Nft", "PremAccounts"))]
-    fn prem_accounts_service_integration_serial(
-        chain: psibase::Chain,
-    ) -> Result<(), psibase::Error> {
+    #[test]
+    fn create_tuple_unpacks_as_action_struct() {
```
James-Mart: This seems to just retest fracpack?

packages/user/NameMarket/query-service/src/lib.rs
```diff
@@ -134,14 +132,15 @@ mod service {
...
+                        initial_price: Decimal::new(auction.initial_price, precision).to_string(),
                         target: rate_limit.target_min, // target_min == target_max in our usage
                         floor_price: Decimal::new(
                             Quantity::from(rate_limit.floor_difficulty),
                             precision,
                         ),
                         window_seconds: rate_limit.window_seconds,
-                        increase_ppm: rate_limit.increase_ppm,
-                        decrease_ppm: rate_limit.decrease_ppm,
+                        increase_pct: ppm_to_pct(rate_limit.increase_ppm),
```
James-Mart: _Non-pr-blocking_
You should learn the `async_graphql` `#[ComplexObject]` pattern we use to do type conversions for query purposes. It will simplify your query services significantly. Examples in `tokens`, `token-stream`, `virtual-server`, `guilds`, etc

packages/user/NameMarket/plugin/src/lib.rs
```diff
@@ -0,0 +1,368 @@
...
+    };
+
+    let price_str = row.price.to_string();
+    TokensHelpers::decimal_to_u64(sys_token_id, &price_str)
+        .map_err(|_| ErrorType::QueryResponseParseError("invalid price".into()).into())
+}
+
+impl Authorized for NameMarketPlugin {
+    #[psibase_plugin::authorized(None)]
+    fn graphql(query: String) -> Result<String, Error> {
```
James-Mart, line 352: Why change this from `Medium` trust to `None`?
That means if alice is logged into mallory's third-party app, it can silently see what premium names alice has bought and read other "infra-protected" user data.

packages/user/PremAccounts/query-service/src/lib.rs
```diff
@@ -170,22 +169,21 @@ mod service {
             .await
         }
 
-        /// Events: premium **name** history for `owner`
+        /// Events: account **name** history for `owner`
         async fn name_events(
             &self,
-            owner: AccountNumber,
```
James-Mart, line 176: Not blocking the PR over it, but this is worse.

There are valid cases where we may want to query the names owned by an account not actively logged in:
* `config` app may want to show account names owned by `root`
* `workshop` app may want to show account names owned by the app account
* etc

packages/user/Config/plugin/wit/impl.wit
```diff
@@ -7,16 +7,17 @@ world impl {
...
     include transact:plugin/imports;
     include packages:plugin/imports;
-    include prem-accounts:plugin/imports;
+    include name-market:plugin/imports;
     include virtual-server:plugin/imports;
     include sites:plugin/imports;
 
     import accounts:plugin/admin;
-    import prem-accounts:plugin/market-admin;
+    import name-market:plugin/authorized;
```
James-Mart: Is this import used?

packages/user/Config/plugin/Cargo.toml
```diff
@@ -7,6 +7,8 @@ publish.workspace = true
 
 [dependencies]
 wit-bindgen-rt.workspace = true
+serde.workspace = true
+serde_json.workspace = true
```
James-Mart: These needed?

packages/user/NameMarket/plugin/src/lib.rs
```diff
@@ -0,0 +1,368 @@
...
+        }
+
+        Ok(())
+    }
+}
+
+impl Api for NameMarketPlugin {
+    #[psibase_plugin::authorized(None)]
+    fn can_create_account() -> bool {
+        if !AccountsApi::is_logged_in() {
```
James-Mart: `<NameMarket plugin>::can_create_account` should answer the question about whether the name market service is sufficiently initialized and there are markets created.

If the accounts-prompt context needs to limit creating accounts through the marketplace to when an account is actively logged-in, then *that* is where the check belongs.

Technically the dependency circularity (`Accounts` -> `NameMarket` -> `Accounts`) is currently allowed. But I am trying to eliminate circular plugin deps, so I'd like to avoid adding new ones

packages/system/Accounts/plugin/accounts/Cargo.toml
```diff
@@ -31,6 +31,9 @@ world = "impl"
 "permissions:plugin" = { path = "../../../../user/Permissions/plugin/wit/world.wit" }
 "invite:plugin" = { path = "../../../../user/Invite/plugin/wit/redemption.wit" }
 "auth-sig:plugin" = { path = "../../../AuthSig/plugin/wit/world.wit" }
+"name-market:plugin" = { path = "../../../../user/NameMarket/plugin/wit/world.wit" }
+"tokens:plugin" = { path = "../../../../user/Tokens/plugin/wit/world.wit" }
```
James-Mart: I think this isn't used?

packages/user/Homepage/plugin/Cargo.toml
```diff
@@ -0,0 +1,27 @@
...
+[package.metadata.component.target]
+world = "impl"
+
+[package.metadata.component.target.dependencies]
+"host:types" = { path = "../../Host/plugin/types/wit/world.wit" }
+"host:crypto" = { path = "../../Host/plugin/crypto/wit/world.wit" }
+"transact:plugin" = { path = "../../../system/Transact/plugin/wit/world.wit" }
+"auth-sig:plugin" = { path = "../../../system/AuthSig/plugin/wit/world.wit" }
+"name-market:plugin" = { path = "../../NameMarket/plugin/wit/world.wit" }
+"accounts:plugin" = { path = "wit/deps/accounts-homepage.wit" }
```
James-Mart: Why can't you just link to the normal accounts plugin `world.wit`?

packages/user/Config/plugin/Cargo.toml
cool-ant: nope

packages/user/Config/plugin/wit/impl.wit
cool-ant: nope

packages/user/PremAccounts/query-service/src/lib.rs
cool-ant: ok, reverted.

packages/user/NameMarket/plugin/src/lib.rs
cool-ant: clear. removed.

packages/user/PremAccounts/query-service/src/lib.rs
cool-ant: and config/workshop-enabled auth added

James-Mart (approved):

PR #1888, opened 2026-05-29
New name encoding

swatanabe:


packages/system/Accounts/src/Accounts.cpp
```diff
@@ -79,11 +90,20 @@ namespace SystemService
...
+         {
+            check(getSender() == name.base(),
+                  "Subaccounts can only be created by the primary account");
+         }
+         else if (getSender() != service)
+         {
+            check(!strName.starts_with("x-"),
+                  "The 'x-' account prefix is reserved for infrastructure providers");
+            check(strName.length() >= 10,
+                  "account name must be at least 10 characters: " + strName);
```
Velua, line 105: I thought the max is now 10 chars? This suggests it can be 10 or more

packages/user/PremAccounts/query-service/src/lib.rs
Velua: Should this be `http` instead of `http-server` now ?

libraries/psibase/common/include/psibase/name.hpp
```diff
@@ -1,509 +1,93 @@
...
-
-      static constexpr const int  PRECISION = std::numeric_limits<code_value>::digits;
-      static constexpr code_value MAX_CODE  = (code_value(1) << code_value_bits) - 1;
-      //  static constexpr code_value MAX_FREQ      = (code_value(1) << FREQUENCY_BITS) - 1;
-      static constexpr code_value ONE_FOURTH = code_value(1) << (code_value_bits - 2);
-      ;
-      static constexpr code_value ONE_HALF      = 2 * ONE_FOURTH;
-      static constexpr code_value THREE_FOURTHS = 3 * ONE_FOURTH;
-      static constexpr int        model_width   = 38;
+      inline constexpr std::string_view subaccountSeparator = "☺";
```
Velua: Is that what I think it is? ☺ U+263A ? I assumed the separator would have been a `#` or something.

swatanabe: If someone tries to use a subaccount as a subdomain name, it should not cause the URL to be interpreted strangely, which means `@`, `?`, `:`, `/`, and `#` are off the table.

packages/system/Accounts/src/Accounts.cpp
swatanabe: That's a different check. This is the check for whether the account is one that anyone can create, which I didn't change (It's only in the diff because it's indented).

libraries/psibase/common/include/psibase/name.hpp
swatanabe: I thought you would be happy to have subaccounts

James-Mart: Hahaha!

libraries/psibase/common/include/psibase/NameEncoder.hpp
```diff
@@ -0,0 +1,146 @@
...
+
+      constexpr T encode(auto&& rng, T sepIdx) const
+      {
+         T           result    = 0;
+         std::size_t maxlen    = N;
+         bool        allowsSep = false;
+         bool        allowsEnd = true;
+         for (T idx : rng)
+         {
+            if (maxlen == 0 || idx > numRegularChars + 1 || (idx == sepIdx && !allowsSep))
```
James-Mart: Wait, the alphabet has cardinality `numRegularChars + 1`, right? So then shouldn't this check be `idx > numRegularChars`?

libraries/psibase/common/include/psibase/name.hpp
```diff
@@ -1,509 +1,93 @@
...
 
-      struct prob
+      // removes and encodes the trailing subaccount
+      constexpr std::uint8_t subaccount_from_str(std::string_view& s, std::string_view sep)
```
James-Mart, line 17: I find it a bit odd that returned "0" from `subaccount_from_str` doesn't distinguish between:
  * no separator found
  * invalid subaccount
  * valid subaccount 0
They all just return 0.

libraries/psibase/common/include/psibase/name.hpp
```diff
@@ -1,509 +1,93 @@
...
-      unsigned char m_Mask     = 0x80;
-      uint64_t      m_output   = 0;
-
-      auto put_bit = [&](bool b)
-      {
-         if (b)
-            m_NextByte |= m_Mask;
-         m_Mask >>= 1;
-         if (!m_Mask)
+         auto pos = s.find(sep);
```
James-Mart, line 19: Does this behave as intended if sep is `""`?

libraries/psibase/common/include/psibase/name.hpp
```diff
@@ -1,509 +1,93 @@
...
-            high <<= 1;
-            high++;
-            low <<= 1;
-            high &= MODEL::MAX_CODE;
-            low &= MODEL::MAX_CODE;
+            std::uint32_t digit =
+                static_cast<unsigned char>(*start) - static_cast<unsigned char>('0');
+            if (digit >= 10)
+               return 0;
+            sub = sub * 10 + digit;
```
James-Mart, line 33: I guess leading zeroes in the subaccount string is okay? Means there's not a single canonical string representation of a subaccount, but I guess that's fine

packages/user/Symbol/test/src/Symbol-test.cpp
```diff
@@ -82,7 +82,10 @@ SCENARIO("Buying a symbol")
       {
          alice.to<Tokens>().credit(sysToken, Symbol::service, quantity, memo);
          auto symbolId = SID{"aBc"};
-         CHECK(a.create(symbolId).failed(notForSale));
+         if (symbolId.str() != "abc")
+         {
+            CHECK(a.create(symbolId).failed(notForSale));
```
James-Mart, line 87: This test could probably just be removed if capital letters aren't persisted in the AccountNumber

libraries/psibase/common/include/psibase/name.hpp
swatanabe: No subaccount and valid subaccount 0 are considered equivalent here. We're not trying to parse subaccounts separately from the whole account. An invalid subaccount can't be parsed successfully as part of the base account, so it will be caught then.

libraries/psibase/common/include/psibase/name.hpp
swatanabe: That isn't allowed. The separator also can't start with a character in the alphabet.

libraries/psibase/common/include/psibase/name.hpp
swatanabe: I don't see an immediate reason to enforce canonical string representation, everywhere. If we want this, we also need to disallow parsing upper case. Note that the existing codebase already enforces canonical account names in some places by checking that they round-trip.

packages/user/Symbol/test/src/Symbol-test.cpp
swatanabe: I left the test in case we ever tinker with `AccountNumber` more or implement a separate encoding for SID.

libraries/psibase/common/include/psibase/NameEncoder.hpp
swatanabe: Right, though it won't actually break anything, due to the fact that the input is sanitized in `name_to_number`. The only value outside the alphabet that might be seen is `UINT64_MAX`.

James-Mart (approved):

PR #1889, opened 2026-06-05
Handle flags in local packages

swatanabe:
Previously, `PUT` to `/services/*` couldn't control the code flags, so it just hard-coded `isPrivileged`. This change allows deploying non-privileged local services, and also allows setting the `isReplacement` flag.


James-Mart (approved):
I finished reviewing this before realizing I wasn't a reviewer

PR #1896, opened 2026-06-11
Fix blockstart query scan

James-Mart:
## Problem

The block-context subquery generated by rust/cpp gql lib (`WHERE ROWID <= r.rowId ORDER BY ROWID DESC LIMIT 1` against `"history.transact.blockStart"`) forces SQLite to scan every blockStart event from genesis, once per returned row. The events virtual table in REvents only iterates forward (`kvGreaterEqual`) and never sets `orderByConsumed`, so SQLite cannot stop after one row.

## Solution

`event_best_index` now tells SQLite when a scan already satisfies `ORDER BY ROWID` (setting `orderByConsumed` and passing a direction flag to `event_filter`), and the cursor supports descending iteration. With the order consumed, SQLite stops after `LIMIT 1`, so the blockStart lookup becomes a single `kvLessThan` host call per returned row instead of a scan over every block since genesis.

## Additional fix

`event_filter` never reset the upper scan bound (`c->end`), so restarting a scan on a reused cursor could retain a stale, narrower bound.


packages/user/Events/src/REvents.cpp
```diff
@@ -609,13 +619,20 @@ int event_best_index(sqlite3_vtab* base_vtab, sqlite3_index_info* info)
       }
    }
    int  best = std::ranges::min_element(constraints) - constraints.begin() - 1;
-   auto buf  = static_cast<char*>(alloca(info->nConstraint + 1));
+   bool desc = false;
+   if (best == -1 && info->nOrderBy == 1 && info->aOrderBy[0].iColumn == -1)
```
swatanabe: This only handles `ORDER BY ROWID`. A more general test is `best == info->aOrderBy[0].iColumn`. Also, `IndexConstraints` can be adjusted to reduce the `estimatedCost()` of the `ORDER BY` index, based on the `LIMIT`. `SQLITE_INDEX_CONSTRAINT_LIMIT` indicates the `LIMIT` if there is one.

packages/user/Events/src/REvents.cpp
```diff
@@ -141,8 +141,10 @@ struct EventCursor : sqlite3_vtab_cursor
 {
    std::vector<char> key;
    std::vector<char> end;
+   std::vector<char> begin;
```
swatanabe: `begin` is not set for forward traversal. For `descending`, `end` is set but not used. I think it would be better to use only two variables. Either use `[begin, end)` and choose one as the key: `descending? end : begin`, or always use `[key, end)` and set `end` to the appropriate boundary based on direction.

packages/user/Events/src/REvents.cpp
```diff
@@ -153,94 +155,95 @@ struct EventCursor : sqlite3_vtab_cursor
...
    bool          eof() { return key.size() < prefixLen; }
    std::uint64_t eventId() const { return keyToEventId(key, prefixLen); }
-   void          seek()
+   void          load(std::uint32_t sz)
    {
-      auto sz = psibase::raw::kvGreaterEqual(handle, key.data(), key.size(), prefixLen);
       if (sz == 0xffffffffu)
       {
          setEof();
+         return;
```
swatanabe: Was this really necessary? Reindenting for no reason makes review harder, because github doesn't show `diff -b`.

James-Mart: Fyi you can hide whitespace changes in the diff settings in GitHub

packages/user/Events/src/REvents.cpp
James-Mart: Done

packages/user/Events/src/REvents.cpp
James-Mart: Undid the early return

packages/user/Events/src/REvents.cpp
James-Mart: Done

swatanabe (approved):

PR #1897, opened 2026-06-15
adopt getDelegations

Velua:
Implements #1799


packages/system/AuthAny/include/services/system/AuthAny.hpp
```diff
@@ -17,17 +17,23 @@ namespace SystemService
 
       void canAuthUserSys(psibase::AccountNumber user);
 
+      std::vector<psibase::AccountNumber> getDelegations(psibase::AccountNumber      sender,
```
swatanabe: This needs to use the `Sys` suffix.

packages/system/AuthAny/include/services/system/AuthAny.hpp
```diff
@@ -17,17 +17,23 @@ namespace SystemService
...
+                                                         std::optional<ServiceMethod> method);
+
       bool isAuthSys(psibase::AccountNumber              sender,
-                     std::vector<psibase::AccountNumber> authorizers);
+                     std::vector<psibase::AccountNumber> authorizers,
+                     std::optional<ServiceMethod>        method);
```
swatanabe: The signature doesn't need to be changed. Unused trailing optional parameters can be omitted.

packages/system/AuthDelegate/include/services/system/AuthDelegate.hpp
```diff
@@ -57,35 +57,35 @@ namespace SystemService
...
+      /// Get the accounts this auth service delegates authority to for a sender.
+      std::vector<psibase::AccountNumber> getDelegations(psibase::AccountNumber      sender,
+                                                         std::optional<ServiceMethod> method);
+
       /// Check whether a specified set of authorizer accounts are sufficient to authorize sending a
       /// transaction from a specified sender.
       ///
       /// * `sender`: The sender account for the transaction potentially being authorized.
-      /// * `authorizers`: The set of accounts that have already authorized the execution of the transaction.
+      /// * `authorizers`: Accounts that have already been authorized through their delegations.
```
swatanabe: This sentence is confusing

packages/system/StagedTx/service/src/policy.rs
```diff
@@ -9,43 +9,100 @@ pub fn get_auth_service(sender: AccountNumber) -> Option<AccountNumber> {
...
+}
+
+fn check_auth_recursive(
+    user: AccountNumber,
+    method: Option<ServiceMethod>,
+    responders: &[AccountNumber],
+    check_fn: fn(&ServiceCaller, AccountNumber, Vec<AccountNumber>, Option<ServiceMethod>) -> bool,
+    path: &[AccountNumber],
+) -> bool {
+    if path.contains(&user) {
```
swatanabe: This is replicating the old algorithm, which was designed to work across services. It can be improved significantly to reduce duplicate calls, which was one of the goals of this task.

packages/system/Producers/src/Producers.cpp
```diff
@@ -273,43 +241,30 @@ namespace SystemService
...
-      auto authSet = authSet_opt ? std::move(*authSet_opt) : std::vector<AccountNumber>{};
-
-      // Base case to prevent infinite recursion
-      if (std::ranges::contains(authSet, sender))
-         return false;
-
-      authSet.push_back(sender);
+      return ::getProducers()                          //
+             | std::views::transform(&Producer::name)  //
+             | std::ranges::to<std::vector>();
```
James-Mart: This duplicates `std::vector<psibase::AccountNumber> Producers::getProducers()`

packages/system/StagedTx/service/src/policy.rs
Velua: Reapproached this, hope I'm on the right track...

swatanabe: It's on the right track, but there's one nasty case that the old code handles that this doesn't.

If A delegates to B and B delegates to A, and A can be successfully authorized without B but B cannot be successfully authorized without A, then the result depends on the order in which A and B are reached. The correct result in this case is success for both.

packages/system/StagedTx/service/src/policy.rs
```diff
@@ -9,43 +9,183 @@ pub fn get_auth_service(sender: AccountNumber) -> Option<AccountNumber> {
...
+                    .copied()
+                    .filter(|delegate| {
+                        responders.contains(delegate) || authorized.contains(delegate)
+                    })
+                    .collect()
+            };
+
+            // Only the originating account is checked against the requested
+            // method; delegated checks have the method wiped by staged-tx.
+            let method = if delegation.account == user {
```
swatanabe: This isn't quite right. `user` can appear both with and without `ServiceMethod` and these should be distinct.

packages/system/StagedTx/service/src/policy.rs
```diff
@@ -9,43 +9,183 @@ pub fn get_auth_service(sender: AccountNumber) -> Option<AccountNumber> {
...
+
+/// Determines whether `user` is authorized given the accounts that actually
+/// responded (`responders`).
+///
+/// A leaf account (one that delegates to no one) is authorized if it is among
+/// the responders. Otherwise authority is derived from the delegated accounts
+/// that are themselves authorized.
+
+fn check_delegation_auth(
+    user: AccountNumber,
```
swatanabe, line 230: In `reject` and `accept` + `autoexec` (but not in `execute_impl`), delegated account auth can be reused across actions.

packages/system/StagedTx/service/src/policy.rs
```diff
@@ -9,43 +9,183 @@ pub fn get_auth_service(sender: AccountNumber) -> Option<AccountNumber> {
...
+    responders: &[AccountNumber],
+    check_fn: CheckFn,
+) -> bool {
+    let graph = build_delegation_graph(user, Some(method));
+    let mut authorized: Vec<AccountNumber> = Vec::new();
+
+    loop {
+        let mut changed = false;
+
+        for delegation in &graph {
```
swatanabe, line 187: A few optimizations are possible:
- If authorization fails when all the delegates are known to succeed or fail, then it can be marked as known to fail.
- If authorization succeeds before adding delegates, then we can short-circuit and skip adding delegates.
- Working bottom up reduces the number of passes that need to be made. The current algorithm actually hits the worst case for a simple chain `A -> B -> C -> ...`

packages/system/Producers/include/services/system/Producers.hpp
```diff
@@ -79,37 +79,33 @@ namespace SystemService
...
-      ///              the sender is already in this set, then the function should return false.
       ///
       /// Returns:
-      /// * `true`: If the total authorizations from `authorizers` or their auth services meets sender's threshold
-      /// * `false`: If not returning true, or on recursive checks for the same sender
-      bool isAuthSys(psibase::AccountNumber                             sender,
-                     std::vector<psibase::AccountNumber>                authorizers,
-                     std::optional<ServiceMethod>                       method,
-                     std::optional<std::vector<psibase::AccountNumber>> authSet);
+      /// * `true`: If enough authorizers meet the sender's threshold
```
swatanabe: This sentence doesn't quite mean what you intend.

packages/system/Transact/include/services/system/Transact.hpp
```diff
@@ -138,43 +138,49 @@ namespace SystemService
...
       ///
       /// Returns:
       /// * `true`: The authorizers are sufficient to authorize a transaction from the sender.
       /// * `false`: The authorizers are not sufficient to authorize a transaction from the sender.
-      bool isAuthSys(psibase::AccountNumber                             sender,
-                     std::vector<psibase::AccountNumber>                authorizers,
-                     std::optional<ServiceMethod>                       method,
-                     std::optional<std::vector<psibase::AccountNumber>> authSet);
+      bool isAuthSys(psibase::AccountNumber              sender,
+                     std::vector<psibase::AccountNumber> authorizers);
```
swatanabe: `method` shouldn't be removed here.

packages/system/StagedTx/service/src/policy.rs
```diff
@@ -9,43 +9,235 @@ pub fn get_auth_service(sender: AccountNumber) -> Option<AccountNumber> {
...
+        Exit(Delegation),
+    }
+
+    let mut stack: Vec<Frame> = vec![Frame::Enter(user, Some(method))];
+    let mut graph: Vec<Delegation> = Vec::new();
+    let mut seen: Vec<AccountNumber> = Vec::new();
+    let mut authorized: Vec<AccountNumber> = Vec::new();
+    let mut failed: Vec<AccountNumber> = Vec::new();
+
+    // Top-down expansion. Records `graph` in post-order and resolves every
```
swatanabe, line 127: pre-order with a final reverse would be a little simpler.

packages/system/StagedTx/service/src/policy.rs
```diff
@@ -9,43 +9,235 @@ pub fn get_auth_service(sender: AccountNumber) -> Option<AccountNumber> {
...
+        for delegate in get_delegations(&caller, account, account_method) {
+            if !delegates.contains(&delegate) {
+                delegates.push(delegate);
+            }
+        }
+
+        // Authorizers available without resolving any delegate: the responders
+        // themselves for a leaf, or the delegates that responded directly.
+        let direct_authorizers: Vec<AccountNumber> = if delegates.is_empty() {
+            responders.to_vec()
```
swatanabe: This seems inconsistent. I think the set of authorizers passed to `is_auth` should be calculated in the same way regardless of whether the account is a leaf.

packages/system/StagedTx/service/src/policy.rs
```diff
@@ -9,43 +9,235 @@ pub fn get_auth_service(sender: AccountNumber) -> Option<AccountNumber> {
...
+                .collect()
+        };
+
+        if check_fn(&caller, account, direct_authorizers, account_method) {
+            // Authorized already; its delegates can't change that, so skip them.
+            authorized.push(account);
+            continue;
+        }
+
+        if delegates.is_empty() {
```
swatanabe, line 163: The general condition is `delegates ⊆ direct_authorizers`, which can be simplified to `delegates.len() <= direct_authorizers.len()`, because `direct_authorizers` was already filtered.

packages/system/StagedTx/service/src/policy.rs
```diff
@@ -9,43 +9,235 @@ pub fn get_auth_service(sender: AccountNumber) -> Option<AccountNumber> {
...
+                .collect();
+
+            let caller = auth_caller(delegation.auth_service);
+            if check_fn(&caller, delegation.account, authorizers, delegation.method) {
+                authorized.push(delegation.account);
+                changed = true;
+            } else if delegation
+                .delegates
+                .iter()
+                .all(|delegate| authorized.contains(delegate) || failed.contains(delegate))
```
swatanabe: Should this check `responders` too? Actually, is there any difference between `responders` and `authorized` that requires them to be separate sets?

packages/system/StagedTx/service/src/policy.rs
```diff
@@ -9,43 +9,235 @@ pub fn get_auth_service(sender: AccountNumber) -> Option<AccountNumber> {
...
+        seen.push(account);
+
+        let Some(auth_service) = get_auth_service(account) else {
+            // An account with no auth service can never become authorized.
+            failed.push(account);
+            continue;
+        };
+
+        let caller = auth_caller(auth_service);
+        let mut delegates: Vec<AccountNumber> = Vec::new();
```
swatanabe: `HashSet` automatically deduplicates and has efficient set intersection/is_subset. If the order is significant, there's `indexmap::IndexSet`.

packages/system/StagedTx/service/src/policy.rs
Velua: Can we get away with that working with a DAG though? If we knew we were working with a tree it would be simpler.

packages/system/StagedTx/service/src/policy.rs
Velua: I've written a response to this but AI still explains it better haha

> Good eye, but I don't think these *can* be unified. `direct_authorizers` is the "can we decide this account up front, from who's already responded, before expanding any delegates?" check.
>
> For a **non-leaf**, the accounts that can authorize it are exactly its declared delegates, so we filter to `delegates ∩ responders` and ask whether those alone satisfy `is_auth` — if so, we short-circuit and skip expanding them.
>
> For a **leaf**, there are no delegates to intersect with, so applying the same formula would always give the empty set and *every* leaf would fail. A leaf's authorizing identity (keys/claims) lives inside its auth service and isn't visible to us as accounts, so we hand `is_auth` the full responder set and let it decide.
>
> So the asymmetry is deliberate: for a non-leaf we *know* the exact accounts that matter (the delegates) and narrow to them; for a leaf we can't, so we defer to the auth service with everything. Unifying to `delegates ∩ responders` breaks leaves; unifying the other way (all responders for non-leaves too) would feed an auth service accounts that aren't its delegates, quietly widening trust.

packages/system/StagedTx/service/src/policy.rs
Velua: Responders are fixed input whereas authorized is derived, I don't think it worth merging them.

packages/system/StagedTx/service/src/policy.rs
Velua: Adopted hashsets

packages/system/StagedTx/service/src/policy.rs
swatanabe: As usual, the AI is wrong. It's describing the code without any real understanding of the underlying reason.

- Why does a leaf depend on something which is not in the delegates?
- Does this extra dependency have any logical relationship to being a leaf? Is there any reason a non-leaf couldn't have the same extra dependency?

Also the claim that passing extra accounts to `is_auth` widens trust is just silly.

packages/system/StagedTx/service/src/policy.rs
swatanabe: Whether it's fixed or derived doesn't imply that there is any difference between how they're used.

packages/system/StagedTx/service/src/policy.rs
swatanabe: It's isn't a DAG, though. I suppose that the order you have is probably better on average even with cycles, although cycles can still create anomalous orderings. If we really want to be optimal, however, we would need to use Tarjan's algorithm to group strongly connected components.

packages/system/StagedTx/service/src/policy.rs
Velua: Yeah hang on, under what circumstances would a leaf account return `true` on `is_auth` ...? 
If the auth service can't mention any accounts from `getDlgs` then why would it start returning `true` from either `is_auth` or `is_reject` ? 

If it's a leaf account can we safely assume it's not interested in being queried `is_auth` or `is_reject` ? Given it's offering no help in discovery of accounts it would be satisfied with when asked `is_auth` or `is_reject` ?

swatanabe: The reason for this shows up in AuthSig and AuthAny:

```c++
bool AuthSig::isAuthSys(psibase::AccountNumber              sender,
                        std::vector<psibase::AccountNumber> authorizers)
{
   return std::ranges::contains(authorizers, sender);
}
```

Your current code relies on this, but only at the top level.

packages/system/StagedTx/service/src/policy.rs
Velua: Okay, I've now combined responders and authorized into backed.

swatanabe (approved):

PR #1902, opened 2026-06-18
DiffAdjust: Support target=0, unclamp ppm changes, batched=individual increments

cool-ant:
target=0 allows for *every* event trigger a price change.
inc/dec changes were clamped to 100%; now they're opened to the full u32, which effectivelly allows >400,000% adjustments


packages/user/DiffAdjust/service/src/lib.rs
```diff
@@ -123,6 +117,41 @@ pub mod tables {
             self.decrease_ppm as f64 / ONE_MILLION as f64
         }
 
+        fn f64_to_u64(value: f64) -> u64 {
+            if !value.is_finite() || value <= 0.0 {
```
Velua: `!value.is_finite()` would mean, it's infinity right? So why do we return `0` and not `u64::max` for that case?

Velua (changes requested):

packages/user/DiffAdjust/service/src/lib.rs
cool-ant: We're checking against *negative* infinity (only for a decrease); we're not concerned with positive infinite (we ensure that in the parent fn). Your question exposes some logic that shouldn't be grouped together like this, namely that line (and the next) are decrease-related but in a fn that is equally applied to increase and decrease. It works, but it's hard to reason about. I've reworked it and the surrounding functions to make things more clear.

swatanabe: If you're checking for negative infinity, it's redundant, because `-f64::INFINITY < 0`

cool-ant: well, true. is_finite() will also return false for NaN which was actually what initially kicked off the change. For some reason, the tests passed locally, but in CI wasm produced NaN, and I needed a check that included that. This was a broad enough check to catch that additional case.

packages/user/DiffAdjust/service/src/lib.rs
```diff
@@ -133,13 +171,13 @@ pub mod tables {
                 self.counter = 0;
                 self.last_update = TimePointSec::from(now.seconds - seconds_remainder as i64);
                 if below_target {
-                    let mut new_difficulty = self.active_difficulty;
                     let factor = 1.0 - self.ratio_decrease();
-                    for _ in 0..windows_elapsed {
-                        let temp = new_difficulty as f64 * factor;
-                        new_difficulty = (temp as u64).max(self.floor_difficulty);
```
swatanabe, line 140: $\texttt{factor}^{\texttt{windows\\_elapsed}}$ is calculated this way for a reason. It ensures that the rounding of the final result does not depend on the number of windows that are applied at once.

packages/user/DiffAdjust/service/src/lib.rs
```diff
@@ -123,6 +117,50 @@ pub mod tables {
...
+            } else {
+                Self::f64_to_u64_saturating(value)
+            }
+        }
+
+        fn apply_increase(difficulty: u64, factor: f64, times: u32) -> u64 {
+            if times == 0 || difficulty == u64::MAX || factor <= 1.0 {
+                return difficulty;
+            }
+            let powered = factor.powf(times as f64);
```
swatanabe: `powi` is more efficient and doesn't require a cast.

packages/user/DiffAdjust/service/src/lib.rs
cool-ant: ok, didn't realize that. emulating individual window adjustments

packages/user/DiffAdjust/service/src/lib.rs
cool-ant: This required some other checks.

Velua (approved):

packages/user/DiffAdjust/service/src/lib.rs
```diff
@@ -148,15 +204,15 @@ pub mod tables {
         fn check_difficulty_increase(&mut self, clamp_increase: bool) -> u64 {
             if self.counter > self.target_max {
                 let factor = 1.0 + self.ratio_increase();
-                let mut times_over_target = self.counter / self.target_max;
+                let mut times_over_target = match self.target_max {
+                    0 => self.counter,
+                    _ => self.counter / self.target_max,
```
swatanabe: This math seems a bit wonky:

- 0 -> increase difficulty on every increment
- 1 -> increase difficulty by **2** every other increment
- 2 -> increase difficulty by 1 every third increment
- k -> increase difficulty by 1 every k+1 increments

Also, if we call `increment(C)` instead of `increment(1)`, we increase the difficulty $\lfloor C/k \rfloor$ times. i.e. `increment(C)` is noticeably difference from `increment(1)` repeated `C` times.

packages/user/DiffAdjust/service/src/lib.rs
```diff
@@ -123,6 +117,68 @@ pub mod tables {
...
+            }
+        }
+
+        fn apply_increase(difficulty: u64, factor: f64, times: u32) -> u64 {
+            if times == 0 || difficulty == u64::MAX || factor <= 1.0 {
+                return difficulty;
+            }
+            // ensure no cast safety of `times` from u32 to i32
+            let powered = if times > i32::MAX as u32 {
+                f64::INFINITY
```
swatanabe, line 129: This is correct, but its correctness depends on the fact that we are using PPM. $1.000001^k$ doesn't pass `f64::MAX` until close to `i32::MAX` (`0x2A4E6E1C`). It overflows `u64` at `0x02a4e6e2`.

cool-ant: ok, since we're using PPM, are you comfortable with the state of this? Or do you think we need to change how this works?

swatanabe: It's not obvious that this is something to be careful about. I think the precondition needs to be called out or even asserted, so we don't forget.

packages/user/DiffAdjust/service/src/lib.rs
```diff
@@ -108,7 +102,7 @@ pub mod tables {
                 floor_difficulty,
                 last_updated,
                 increase_ppm,
-                decrease_ppm,
+                Self::clamp_decrease_ppm(decrease_ppm),
```
James-Mart: All the other fields use `check` with explicit errors if misconfigured, but you're just silently clamping `decrease_ppm`.

packages/user/DiffAdjust/service/src/lib.rs
James-Mart: So... `value.is_nan() || value <= 0.0`?

packages/user/DiffAdjust/service/src/lib.rs
```diff
@@ -306,8 +372,12 @@ pub mod service {
...
     /// Increment RateLimit instance, potentially increasing the difficulty.
     ///
-    /// The difficulty may increase multiple times if the counter exceeds `target_max`
-    ///   by more than one multiple of `target_max`.
+    /// `target_max` events accumulate with no adjustment, and the next event triggers one,
+    ///   so the number of adjustments for `events` accumulated in the window is
+    ///   `floor(events / (target_max + 1))`. Adjustments therefore land every
+    ///   `target_max + 1` events (and `target_max == 0` means every event adjusts). A single
+    ///   large increment may apply multiple adjustments at once, matching the equivalent
+    ///   sequence of single increments (the remainder is carried across calls).
```
James-Mart: I find your update to the comment overly verbose to the point of confusion.

Wouldn't the whole concept pretty much be captured by:
```
A difficulty increase is applied once for every `target_max + 1` the counter has accumulated.
```

packages/user/DiffAdjust/service/src/lib.rs
```diff
@@ -211,10 +279,8 @@ pub mod tables {
         pub fn set_ppm(&mut self, increase_ppm: u32, decrease_ppm: u32) {
             self.check_sender_has_nft();
             self.check_difficulty_decrease();
-            Self::check_ppm_change(increase_ppm);
-            Self::check_ppm_change(decrease_ppm);
             self.increase_ppm = increase_ppm;
-            self.decrease_ppm = decrease_ppm;
+            self.decrease_ppm = Self::clamp_decrease_ppm(decrease_ppm);
```
James-Mart: Same comment as 
https://github.com/gofractally/psibase/pull/1902/changes#r3507986795

packages/user/DiffAdjust/service/src/lib.rs
```diff
@@ -148,16 +207,25 @@ pub mod tables {
...
                 let factor = 1.0 + self.ratio_increase();
-                let mut times_over_target = self.counter / self.target_max;
+                // `target_max` events accumulate with no adjustment; the next event triggers
+                //   one. So one adjustment is applied per `target_max + 1` events, and the
+                //   cumulative adjustments for `events` in a window is
+                //   floor(events / (target_max + 1)).
+                // The remainder is carried in `counter` so a batched
+                //   increment produces the same result as the equivalent single increments.
+                // (u64 math avoids overflow when target_max == u32::MAX.)
+                let events_per_adjustment = self.target_max as u64 + 1;
```
James-Mart: I'm kind of against adding this new term, "event", which shows up here and in other places in this PR.

`DiffAdjust` uses a target difficulty and a counter for how much difficulty has accumulated.  This could just be called  `difficulty_per_adjustment` to keep it aligned with existing terminology.

packages/user/DiffAdjust/service/src/lib.rs
James-Mart: Could we just use `let powered = factor.powf(times as f64);`?

packages/user/DiffAdjust/service/src/lib.rs
```diff
@@ -123,6 +117,71 @@ pub mod tables {
...
+            let powered = if times > i32::MAX as u32 {
+                f64::INFINITY
+            } else {
+                factor.powi(times as i32)
+            };
+            // Large exponents can overflow to inf or NaN on wasm; increases saturate, never collapse to zero.
+            if !powered.is_finite() || powered >= u64::MAX as f64 {
+                return u64::MAX;
+            }
+            Self::increase_f64_to_u64(difficulty as f64 * powered)
```
James-Mart: I think this can just be 
`(difficulty as f64 * powered) as u64`

> Casting from a float to an integer will round the float towards zero
> 
> * `NaN` will return `0`
> * Values larger than the maximum integer value, including `INFINITY`, will saturate to the maximum value of the integer type.
> * Values smaller than the minimum integer value, including `NEG_INFINITY`, will saturate to the minimum value of the integer type.

[Source](https://doc.rust-lang.org/stable/reference/expressions/operator-expr.html#numeric-cast)

packages/user/DiffAdjust/service/src/lib.rs
```diff
@@ -123,6 +117,71 @@ pub mod tables {
...
+            let factor = factor.max(0.0);
+            if factor == 0.0 {
+                return floor;
+            }
+            if factor == 1.0 {
+                return difficulty;
+            }
+            // Per-window truncate + floor so batching N windows matches N single-window steps.
+            for _ in 0..times {
+                difficulty = Self::decay_f64_to_u64(difficulty as f64 * factor).max(floor);
```
James-Mart: Similar to [this comment](https://github.com/gofractally/psibase/pull/1902/changes#r3508907697)

I think this can be simplified to just
`difficulty = ((difficulty as f64 * factor) as u64).max(floor);`

packages/user/DiffAdjust/service/src/lib.rs
```diff
@@ -123,6 +117,71 @@ pub mod tables {
...
+        }
+
+        /// Increase compound math may overflow to inf or NaN; saturate rather than collapse to zero.
+        fn increase_f64_to_u64(value: f64) -> u64 {
+            if !value.is_finite() {
+                u64::MAX
+            } else {
+                Self::f64_to_u64_saturating(value)
+            }
+        }
```
James-Mart: If I'm correct about [this](https://github.com/gofractally/psibase/pull/1902/changes#r3508907697) and [this](https://github.com/gofractally/psibase/pull/1902/changes#r3508930877)

All three of these math helpers can be deleted.

packages/user/DiffAdjust/service/src/lib.rs
James-Mart: You could consider updating/improving the high level `//!` comment at the top of the DiffAdjust bindings file (`diff_adjust.rs`) to more clearly document the concept if you feel inspired to improve the docs.

You could also copy that comment into this `DiffAdjust/service/src/lib.rs` file to increase its visibility.

packages/user/DiffAdjust/service/src/lib.rs
cool-ant: This used to be `powf`, which had tradeoffs, leading us to `powi`. There's tradeoffs with both; there are limits that need checking with both. So how about we say what our intentions are with the code and commentary so we can align on a direction.
@James-Mart @swatanabe

packages/user/DiffAdjust/service/src/lib.rs
cool-ant: so how I have it written works. Are we trying to tighten performance? Or communicate more explicitly what we expect to go wrong?

packages/user/DiffAdjust/service/src/lib.rs
cool-ant: ok, wouldn't it be `counts_per_adjustment`? `difficulty` is what's being applied, right? Not what's triggering the adjustment.

James-Mart: Hmm, looking again at the docs, `activity` is technically the term used, so my comment was wrong

> `DiffAdjust` uses a target difficulty and a counter for how much difficulty has accumulated. 

Should be 

> `DiffAdjust` uses a target difficulty and a counter for how much activity has accumulated. 

I'd push either for standardizing on the doc (I guess `activity_per_adjustment`?) or doing a unified terminology cleanup across the whole service.

packages/user/DiffAdjust/service/src/lib.rs
James-Mart: I was just hoping there was some way to avoid the issue altogether using something like `powf`. If there's not, then it would def be better to assert the precondition than to have this "// PRECONDITION" comment.

That said, I don't feel very strongly here, was mostly chiming in to ask if `powi` over `powf` was intentional -sounds like it was.

packages/user/DiffAdjust/service/src/lib.rs
cool-ant: +infinity returns u64::MAX instead of 0
Given the rest of the code, +infinity will next be received by this function.
The current implementation is just a little more resilient again changes to the rest of the code so the parts of this service aren't as-fragile puzzle pieces.

packages/user/DiffAdjust/service/src/lib.rs
cool-ant: that would make a NaN input (result of the multiply) return 0 instead of u64::MAX, which is opposite my approach, and I think less true to intent.
As with the decrease scenario, this is also protecting against scenarios with less assumption about checks that happened before calling this function.

swatanabe: `pow` should only return NaN for $0^0$ or if the base is negative, neither of which should be possible. I don't know why `powf` was returning `NaN` before, but I hope changing to `powi` fixed it.

packages/user/DiffAdjust/service/src/lib.rs
cool-ant: separated this rename/refactor into a separate PR for easier reviewing:
https://github.com/gofractally/psibase/pull/1932

packages/user/DiffAdjust/service/src/lib.rs
cool-ant: In looking over the high-level doc as well as the inline comments, I think everything that's needed is said.

packages/user/DiffAdjust/service/src/lib.rs
cool-ant: Ok, 2 test added, one for the precondition scenario and one for the overflow check that happens here.

packages/user/DiffAdjust/service/src/lib.rs
cool-ant: added a check

swatanabe (approved):

James-Mart (approved):

Issue #1864, opened 2026-05-12
Plugin instantiate/dispose race


swatanabe:
I see the following in my browser console on the home page:

```
Call to accounts:plugin/api->getCurrentUser failed common-lib.js:5218:15
Plugin either invalid or called before ready.
```

This appears to be because multiple calls to supervisor.entry are active at the same time and they share the set of instantiated plugins. The awaits in preload allow for context switching that causes problems.

The execution order is something like this:
```
entry 1: instantiate
entry 2: instantiate (does nothing)
entry 1 dispose
entry 2: run (fails)
```

Velua:
Can we close this now?

PR #1905, opened 2026-06-23
Fixing Supervisor method call race

cool-ant:
Problem (https://github.com/gofractally/psibase/issues/1864)
Method calls to the Supervisor client could race. The big issue was that each method call ends with `disposeAll()` (wasms). If 2 method calls were made in rapid succession, the first call's disposeAll() could rug pull the wams the 2nd method call planned to use. More technically, for `Promise.all[supervisor.preloadPlugins(...), supervisor.entry(...))` the 2nd method call would `await this.preload(...)`, as soon as 1st method call resolved the preload promise, 2nd method call would continue, but the 1st method call wasn't quite done. In its `finally { ... }`, it would call `disposeAll()` potentially at quite the inopportune time for the 2nd method call.

I've included tests that repro the issue (and currently pass with the fix), but if you want to see the issue, I've added logging, which slows things just enough to show the issue. branch: `mm/exposing-supervisor-entry-race`

The Fix
While perhaps not the most elegant, is the simplest effective fix that also mirrors how this was being done already with the old `this.preloadLock` which has effectively had its scope expanded to the full scope of each method call isntead of its former, smaller scope.

Now all Supervisor method calls are guaranteed to complete before the next Supervisor method call is entertained, ensuring there's no `disposeAll()` stomping of other Supervisor method calls.

### DIFF
--- a/packages/.pnp.cjs
+++ b/packages/.pnp.cjs
@@ -2913,10 +2913,13 @@ const RAW_RUNTIME_STATE =
           ["@types/react", "npm:19.2.8"],\
           ["@types/react-dom", "virtual:dc3fc578bfa5e06182a4d2be39ede0bc5b74940b1ffe0d70c26892ab140a4699787750fba175dc306292e80b4aa2c8c5f68c2a821e69b2c37e360c0dff36ff66#npm:19.2.3"],\
           ["eslint", "virtual:dc3fc578bfa5e06182a4d2be39ede0bc5b74940b1ffe0d70c26892ab140a4699787750fba175dc306292e80b4aa2c8c5f68c2a821e69b2c37e360c0dff36ff66#npm:9.39.2"],\
+          ["happy-dom", "npm:20.10.6"],\
           ["react", "npm:19.2.3"],\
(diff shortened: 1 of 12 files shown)


cool-ant:
@James-Mart Interested to form a test strategy with you. we don't have much in the way of tests for packages, and I'd liek to start including them. We should plan that a little. I'll apply whatever we come up with to these tests.

James-Mart:
Some of my feedback on this PR is coming in the form of a proposed PR into your PR:
https://github.com/gofractally/psibase/pull/1927

James-Mart:
So... the vitest thing is of questionable value, in my opinion. Since vitest isn't a full browser like playright, you're just embedding the supervisor lib and calling its exports directly, which isn't exercising the actual supervisor dispatcher (`main.ts`), which is what we really should be doing to validate its behavior. However, I think it has some value right now, because there's no other test harness for front-end stuff *at all*, and at least it's something. For that reason, I'm okay with merging this as is - I just don't think it's worth investing much more time into.

Ultimately my high level strategy for future testing of the infra is:
* Playwright (or similar) for a limited set of front-end tests, in order to validate our browser apps aren't broken, and which also exercise the web platform plugin host (supervisor). 
* `cargo test` for the full plugin integration test harness, which is dependent on getting the psibase CLI platform up and running, which is dependent on various improvements to the plugin infra.

James-Mart (approved):

PR #1906, opened 2026-06-23
VS Code Psibase extension (Package Entities support)

cool-ant:


cool-ant:
this was put in repo `psibase-dev-tools`

PR #1908, opened 2026-06-23
add bash fix

Velua:


.vscode/scripts/build.sh
```diff
@@ -1,4 +1,4 @@
-#bash
+#!/usr/bin/env bash
```
cool-ant, line 1: This *was* (before the nix changes)  #!/bin/bash, and I think that's what it should get back to. Doesn't this approach put a burden on the user (not present in the psibase-contributor container) for have /usr/bin/env in place? I'd hope we could get a solution that returns this to #!/bin/bash.

Velua: I believe generally `#!/usr/bin/env bash` is [more compatible](https://stackoverflow.com/a/10383546) than the original brittle ` #!/bin/bash`. This is documented here on [Wikipedia](https://en.wikipedia.org/wiki/Shebang_(Unix)#Portability) and while we're motivated for [Nix](https://discourse.nixos.org/t/shebang-locations/28992/2) apparently this is also something more compatible for other systems such as [BSD](https://github.com/actions/runner/issues/565#issuecomment-667651399)

nix/README.md
```diff
@@ -72,18 +72,10 @@ nix develop
...
-Inside the `nix develop` shell:
-
-```bash
-# Installs cargo-component, cargo-generate, and cargo-edit (pinned for Rust 1.86.0) into `$CARGO_INSTALL_ROOT`.
-./nix/scripts/setup_prereqs.sh
-```
-
-#### 2.2 For VS Code and forks: run .vscode/scripts/env-setup.sh
+#### 2.1 For VS Code and forks: run .vscode/scripts/env-setup.sh
 This will ensure your build and dev envs can find everything they need
```
cool-ant, line 76: what will? you should mention setting up how to launch Cursor in a `nix develop` shell. I just set up this script
```
#!/usr/bin/env bash
set -euo pipefail

SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
PSIBASE_REPO="$HOME/repos/fractally/psibase"
WORKSPACE="$SCRIPT_DIR/psibase-nix-dev.code-workspace"

exec nix develop "$PSIBASE_REPO" -c env HOST_IP=127.0.0.1 bash -c "cd '$PSIBASE_REPO' && exec cursor '$WORKSPACE'"
```

nix/README.md
```diff
@@ -72,18 +72,10 @@ nix develop
...
 
-#### 2.3 Initializing/Configuring the Environment (using direnv,
+#### 2.2 Initializing/Configuring the Environment (using direnv,
```
cool-ant, line 78: this is completely optional. a preference

nix/README.md
```diff
@@ -102,7 +94,7 @@ EOF
 direnv allow
 ```
 
-#### 2.4 extensions
+#### 2.3 extensions
```
cool-ant: this has been deleted in my main branch

cool-ant:
image
You should also handle this /bin/bash ref as part of this PR

.vscode/scripts/build.sh
cool-ant: huh, never seen it before, but I'm fine with it. That path is the same on Mac as well as Ubuntu (the docker container)?

Velua:
> image You should also handle this /bin/bash ref as part of this PR

Think this bin bash change is going to warrant it's own PR...

Velua:
https://github.com/gofractally/psibase/pull/1912

PR #1910, opened 2026-06-23
use check

Velua:


James-Mart:
Nope, was intentional. I don't want to construct the string unless the condition fails.

PR #1912, opened 2026-06-23
use /usr/bin/env to find bash in scripts

Velua:
This change uses the /usr/bin/env utility to find the location of bash in scripts.
Right now, the scripts are assuming the location of bash which might not be true on some systems such as NixOS and others.

For more information, please see https://en.wikipedia.org/wiki/Shebang_(Unix)#Portability , https://stackoverflow.com/a/10383546

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1274,7 +1274,7 @@ function(write_package_index target dir)
     add_custom_command(
         OUTPUT ${dir}/index.json
         DEPENDS ${deps}
-        COMMAND /bin/bash ${CMAKE_CURRENT_SOURCE_DIR}/make_package_index.sh ${CMAKE_COMMAND} ${dir} ${ARGN} >${dir}/index.json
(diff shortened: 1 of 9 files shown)


CMakeLists.txt
```diff
@@ -1274,7 +1274,7 @@ function(write_package_index target dir)
     add_custom_command(
         OUTPUT ${dir}/index.json
         DEPENDS ${deps}
-        COMMAND /bin/bash ${CMAKE_CURRENT_SOURCE_DIR}/make_package_index.sh ${CMAKE_COMMAND} ${dir} ${ARGN} >${dir}/index.json
+        COMMAND bash ${CMAKE_CURRENT_SOURCE_DIR}/make_package_index.sh ${CMAKE_COMMAND} ${dir} ${ARGN} >${dir}/index.json
```
cool-ant, line 1277: This works? Why have this differ from the shebang path?

swatanabe (approved):

PR #1914, opened 2026-06-24
Accounts Marketplace UI in Homepage, Config and Accounts packages

brandonfancher:
Currently based on the branch from #1878.

If this isn't merged by the time #1902 is merged, there will be a further adjustment to this PR to support target=0.


packages/shared-ui/components/account-key/save-key-step.tsx
```diff
@@ -0,0 +1,152 @@
...
+                                    value={keyB64}
+                                    readOnly
+                                    className="pr-24"
+                                    onFocus={(e) => e.target.select()}
+                                />
+                                <div className="absolute right-2 top-1/2 flex -translate-y-1/2 gap-1">
+                                    <PasswordVisibilityButton
+                                        showPassword={showPassword}
+                                        onToggle={() =>
+                                            setShowPassword(!showPassword)
```
Velua, line 99: I think technically we should be doing something like...
```suggestion
                                            setShowPassword(showPassword => !showPassword)
```

as there's a potential of using stale state if this thing is toggled fast enough or laggy, don't forget we have `useBoolean` around too..

```
  const { showPassword,  toggle: toggleShowPassword } = useBoolean(false)
   <PasswordVisibilityButton
                                        showPassword={showPassword}
                                        onToggle={() => { toggleShowPassword() }
```

packages/shared-ui/components/live-price.tsx
```diff
@@ -0,0 +1,106 @@
...
+): PriceFlash | null {
+    const prevQty = new Quantity(previous, precision, 0);
+    const nextQty = new Quantity(next, precision, 0);
+    if (prevQty.equals(nextQty)) {
+        return null;
+    }
+    if (nextQty.isGreaterThan(prevQty)) {
+        return "up";
+    }
+    return "down";
```
Velua, line 23: Nit.
```suggestion
    return nextQty.isGreaterThan(prevQty) ? "up" : "down";
```

packages/shared-ui/lib/token-decimal-to-bigint.ts
```diff
@@ -0,0 +1,13 @@
+import { tokens } from "./plugins";
+import { callPluginFunction } from "./plugins/index";
+
+export async function tokenDecimalToBigInt(
+    tokenId: number,
+    decimalAmount: string,
+): Promise<bigint> {
```
Velua: Where is this function being used?

packages/user/Config/ui/src/hooks/name-markets/use-configured-markets.ts
Velua: I think we're flirting with a target of `0` aren't we? the `min(1)` might not be what we want.

packages/user/Config/ui/src/pages/account-marketplace/index.tsx
```diff
@@ -0,0 +1,423 @@
...
+                return;
+            }
+
+            await saveMarkets({ dirtyRows, systemToken });
+            const { data: freshRows } = await refetch();
+            formApi.reset(buildNameMarketsFormValues(freshRows));
+        },
+    });
+
+    const isDirty = useStore(form.store, (state) => state.isDirty);
```
Velua: This dirty state is based off Tanstacks idea of dirty which if it's touched but returned to original state it's still considered dirty, which is kinda weird but that's what the library does. 
I think your idea of dirty is different from the default value, which is probably reflected with...
```suggestion
    const isDirty = useStore(form.store, (state) => !state.isDefaultValue);
```

Without this change, I can change a fields floor price, change it back to the original and settle the Unsaved changes pill however the Save button is no longer disabled because Tanstack considers the form dirty.

packages/shared-ui/components/account-key/save-key-step.tsx
brandonfancher: That's an optimization that not going to matter practically here. There are definitely cases where we need to use the `setState` callback, but I can't imagine it a necessity here. This is a common react pattern. Maybe the `setState` callback is _more_ better...but lots of things can be _more_ better in lots of places at lots of times.

packages/shared-ui/lib/token-decimal-to-bigint.ts
brandonfancher: Stale. Removed. Good catch!

packages/user/Config/ui/src/hooks/name-markets/use-configured-markets.ts
brandonfancher: We are, but that's not yet a reality in `main`. I have updating this value on my radar once it is.

packages/user/Config/ui/src/pages/account-marketplace/index.tsx
brandonfancher: Fixed by the PR you opened against this branch, which I merged Thursday. Thanks!

packages/user/Config/ui/src/hooks/name-markets/use-configured-markets.ts
brandonfancher: FWIW, `increasePct` and `decreasePct` are also constrained by current reality. Those are going to change too, and when they do, I'll update this behavior.

Velua (approved):

PR #1915, opened 2026-06-24
Move events to be fully implemented in wasm

swatanabe:


packages/user/Events/include/services/user/Events.hpp
```diff
@@ -48,23 +48,61 @@ namespace UserService
...
+
+      /// Returns the current event merkle root
+      ///
+      /// This must be called by transact in every transaction that
+      /// sends merkle events. After it is called, no more merkle events
+      /// can be sent. transact is responsible for storing the merkle
+      /// root in the block header.
+      psibase::Checksum256 saveMerkle();
+
+      EventTable openEvents(EventDb db, psibase::KvMode mode)
```
James-Mart, line 78: Does this work to open the event service tables outside the tester?

CMakeLists.txt
```diff
@@ -877,9 +877,24 @@ psibase_package(
...
+    DESCRIPTION "Indexes events and provides querying"
+    POSTINSTALL ${CMAKE_CURRENT_SOURCE_DIR}/packages/user/Events/src/postinstall.json
+    DEPENDS wasm
+    # FIXME: The existence of this service forces anything that depends
+    # on Events to be installed after it. Figure out a way to represent
+    # this constraint more directly.
+    SERVICE events+3
+        WASM ${CMAKE_CURRENT_BINARY_DIR}/Nop.wasm
+        SCHEMA ${CMAKE_CURRENT_BINARY_DIR}/Nop-schema.json
+    PACKAGE_DEPENDS "EventsService" "Transact(^${PSIBASE_VERSION}.0)" "TransactService(^${PSIBASE_VERSION}.0)" "HttpServer(^${PSIBASE_VERSION}.0)"
```
James-Mart: Why no version for `EventsService`?

CMakeLists.txt
```diff
@@ -877,9 +877,24 @@ psibase_package(
...
+    VERSION ${PSIBASE_VERSION}.0
+    DESCRIPTION "Indexes events and provides querying"
+    POSTINSTALL ${CMAKE_CURRENT_SOURCE_DIR}/packages/user/Events/src/postinstall.json
+    DEPENDS wasm
+    # FIXME: The existence of this service forces anything that depends
+    # on Events to be installed after it. Figure out a way to represent
+    # this constraint more directly.
+    SERVICE events+3
+        WASM ${CMAKE_CURRENT_BINARY_DIR}/Nop.wasm
+        SCHEMA ${CMAKE_CURRENT_BINARY_DIR}/Nop-schema.json
```
James-Mart, line 896: Is using `nop` here related to the `FIXME`? I don't really understand how this workaround works.

CMakeLists.txt
```diff
@@ -877,9 +877,24 @@ psibase_package(
...
+    OUTPUT ${SERVICE_DIR}/Events.psi
+    NAME Events
+    VERSION ${PSIBASE_VERSION}.0
+    DESCRIPTION "Indexes events and provides querying"
+    POSTINSTALL ${CMAKE_CURRENT_SOURCE_DIR}/packages/user/Events/src/postinstall.json
+    DEPENDS wasm
+    # FIXME: The existence of this service forces anything that depends
+    # on Events to be installed after it. Figure out a way to represent
+    # this constraint more directly.
+    SERVICE events+3
```
James-Mart, line 894: Why +3?

James-Mart (approved):
Great cleanup - remove two intrinsics (simplifying native), remove net 1k lines (much of which is from service libs), remove unused "ui" events
👍 👍

CMakeLists.txt
swatanabe: +1 and +2 are used

CMakeLists.txt
swatanabe: Yes. This is related to the details of the package manager's rules for determining install order.

Specifically the rule right now is that a package is installed after any packages that contain services that might be called by the postinstall script. A service is assumed to call only services provided by the direct dependencies of the package containing it.

Adding the service causes this rule to apply to any package that depends on `Events` and has an `init` action.

packages/user/Events/include/services/user/Events.hpp
swatanabe: Why wouldn't it? It's used in the service implementation, so it had better work there.

CMakeLists.txt
swatanabe: I don't know. Added.

CMakeLists.txt
James-Mart: Makes sense. Consider adding those details into that `FIXME` comment

PR #1919, opened 2026-06-26
Improve dirty tracking

Velua:
From what I can tell we can depend on `meta.isDefaultValue` for tracking dirty state rather than doing our own equality checks like this
```
    return (
        a.enabled === b.enabled &&
        a.initialPrice === b.initialPrice &&
        a.floorPrice === b.floorPrice &&
        a.windowAmount === b.windowAmount &&
        a.windowUnit === b.windowUnit &&
        a.target === b.target &&
        a.increasePct === b.increasePct &&
        a.decreasePct === b.decreasePct
    );

 vs

MARKET_EDITABLE_FIELDS.some((field) => {
    const meta = fieldMeta[`markets[${index}].${field}`];
    return meta != null && meta.isDefaultValue === false;
});
```

defaultValues is built by `buildNameMarketsFormValues` so we can iterate on `state.values.markets` directly rather than doing our own equality checks manually between defaultValues and form state.

### DIFF
--- a/packages/shared-ui/components/form/internal/field-errors.tsx
+++ b/packages/shared-ui/components/form/internal/field-errors.tsx
@@ -7,7 +7,7 @@ type FieldErrorsProps = {
 };
 
 export const FieldErrors = ({ meta }: FieldErrorsProps) => {
-    if (!meta?.errors.length) return null;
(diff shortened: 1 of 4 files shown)


brandonfancher (approved):
Nice

PR #1920, opened 2026-06-26
accrue mid-tx billing transfers until settlement at end-of-tx

James-Mart:
Instead of doing token transfers in real time as billing charges occur, we lazily cache a user's resource balance the first time it's needed, and then accumulate billing charges/refunds throughout the tx (aborting if there would be any overdraft). Then we do a one-time settlement at the end of the tx.

This also applies to:
* The fees that would be sent to the fee receiver.
* The transfers in/out of the disk-billing relay

### DIFF
--- a/packages/system/Transact/src/Transact.cpp
+++ b/packages/system/Transact/src/Transact.cpp
@@ -32,19 +32,56 @@ namespace SystemService
          return config && config->enabled;
       }
 
-      // When set, kv writes are attributed to the system account (`{}`) instead of `trxSender`
-      bool systemWrite = false;
+      // While enabled, kv writes are attributed to the system account (`{}`) instead of
+      // `trxSender`, up to a budget of `systemWriteBudget` net bytes written.
+      bool     systemWriteEnabled = false;
+      uint64_t systemWriteBudget  = 0;
 
       // RAII guard that marks the enclosing scope as performing authorized system
-      // writes. Restores the previous value on destruction so nesting is safe.
+      // writes. All writes are attributed to the system account until the guard is
+      // destroyed.
       struct SystemWriteGuard
       {
-         bool prev;
-         SystemWriteGuard() : prev(systemWrite) { systemWrite = true; }
-         ~SystemWriteGuard() { systemWrite = prev; }
+         SystemWriteGuard()
+         {
+            check(!systemWriteEnabled, "system write already active");
+            systemWriteEnabled = true;
+            // Use the guard when the limit is specified by the scope of the guard,
+            // rather than a byte limit.
+            systemWriteBudget = static_cast<uint64_t>(-1);
+         }
+         ~SystemWriteGuard()
+         {
+            systemWriteEnabled = false;
+            systemWriteBudget  = 0;
+         }
          SystemWriteGuard(const SystemWriteGuard&)            = delete;
          SystemWriteGuard& operator=(const SystemWriteGuard&) = delete;
       };
+
+      // Returns true if a write of net size `delta` should be attributed to the
+      // system account
+      bool chargeSystemWrite(int64_t delta)
+      {
+         if (!systemWriteEnabled)
+            return false;
+         if (delta > 0)
+         {
+            auto d = static_cast<uint64_t>(delta);
+            if (d > systemWriteBudget)
+            {
+               systemWriteBudget  = 0;
+               systemWriteEnabled = false;
+               // If the budget is exceeded, attribute the whole write to
+               // the normal sender.
+               return false;
+            }
+            systemWriteBudget -= d;
+            if (systemWriteBudget == 0)
+               systemWriteEnabled = false;
+         }
+         return true;
+      }
    }  // namespace
 
    void Transact::startBoot(psio::view<const std::vector<Checksum256>> bootTransactions)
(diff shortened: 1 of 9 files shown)


packages/system/Transact/src/Transact.cpp
```diff
@@ -32,19 +32,56 @@ namespace SystemService
          return config && config->enabled;
       }
 
-      // When set, kv writes are attributed to the system account (`{}`) instead of `trxSender`
-      bool systemWrite = false;
+      // While enabled, kv writes are attributed to the system account (`{}`) instead of
+      // `trxSender`, up to a budget of `systemWriteBudget` net bytes written.
+      bool     systemWriteEnabled = false;
```
swatanabe: Does this need to be a separate variable?

packages/system/VirtualServer/service/src/lib.rs
```diff
@@ -417,9 +641,11 @@ mod service {
         };
 
         if balance.value != 0u64 {
-            let sub_account = UserSettings::to_sub_account_key(sender, Some(sub_account));
-            Tokens::call().fromSub(sys, sub_account, balance);
+            let sub_key = UserSettings::to_sub_account_key(sender, Some(sub_account.clone()));
+            Tokens::call().fromSub(sys, sub_key, balance);
```
swatanabe, line 644: I don't think this does the right thing if the subaccount has been billed at all in this transaction.

- If the subaccount expects to pay, `update_balance` will make `available` go negative.
- If the subaccount expects a refund, the balance will be recreated at settlement, which, I think will fail because billing is locked.

packages/system/VirtualServer/service/src/billing.rs
```diff
@@ -0,0 +1,108 @@
...
+        CapacityPricing::settle_relay(resource, relay_delta);
+
+        // Settle drift with relay's own generated fee, leftovers go to the fee receiver
+        to_fee_receiver += CapacityPricing::settle_drift(resource, fee);
+    }
+
+    // 4. Process payer resource refunds
+    for ((account, sub), net) in &payer_accruals {
+        if *net > 0 {
+            Tokens::call().toSub(sys, key(account, sub), Quantity::new(*net as u64));
```
swatanabe, line 71: Is it possible to bill a subaccount that doesn't yet exist, if it's only used to process a refund? I don't see anything in `bill_to_sub` that requires the subaccount to exist.

packages/system/VirtualServer/service/src/lib.rs
```diff
@@ -5,12 +5,25 @@ pub mod rpc;
...
+mod billing;
 mod math_utils;
 
 mod tx_cache {
     use psibase::{abort_message, AccountNumber};
     use std::cell::{Cell, RefCell};
     use std::collections::HashMap;
+    use std::thread::LocalKey;
+
+    pub fn check_lock(lock: &'static LocalKey<Cell<bool>>) {
```
swatanabe: I think a better way to handle this locking would be to use `Option`. Access can call `unwrap` or `expect`. `std::mem::take` will leave it as `None` causing further accesses to fail. This makes it impossible to forget `check_lock`.

packages/system/VirtualServer/service/src/lib.rs
```diff
@@ -720,6 +942,8 @@ mod service {
                 .history()
                 .consumed(disk_user, Disk.as_id(), disk_amount, disk_cost);
         }
+
+        billing::reconcile();
```
swatanabe, line 951: Should we rename `useCpuSys`, since it isn't just handling CPU?

packages/system/VirtualServer/service/src/lib.rs
```diff
@@ -82,6 +95,212 @@ mod tx_cache {
                 .unwrap_or_default()
         })
     }
+
+    /// Cache of account resource balances.
```
swatanabe, line 86: Is every place that should use this accounted for? In this file I see `res_balance`, `res_balance_sub`, and `get_cpu_limit` still using `UserSetting::get_resource_balance`.

packages/system/VirtualServer/service/src/lib.rs
James-Mart: Yeah this one is already on my radar. Was planning to cover in a separate PRs that will be strictly dedicated to renaming / refactoring without changing behavior (not only affecting `useCpuSys`)

packages/system/Transact/src/Transact.cpp
James-Mart: No, simplified

packages/system/VirtualServer/service/src/lib.rs
James-Mart: Good call, action not really intended to be used during the same tx that the subaccount is being used. Added a check.

packages/system/VirtualServer/service/src/billing.rs
James-Mart: bill_to_sub now requires the subaccount to exist

packages/system/VirtualServer/service/src/lib.rs
James-Mart: Nice idea, implemented

packages/system/VirtualServer/service/src/lib.rs
James-Mart: Thank you, fixed!

swatanabe (approved):

PR #1927, opened 2026-06-30
proposed restructuring supervisor message queuing 

James-Mart:
Extracts the message serialization up into the supervisor ingress layer instead of adding it inside the supervisor lib itself.

Also, I propose renaming away from "session lock" because "session" in this context can be confused with a browser session and this is not a browser-session-based lock.

### DIFF
--- a/packages/user/Supervisor/ui/src/serialize.ts
+++ b/packages/user/Supervisor/ui/src/serialize.ts
@@ -3,7 +3,7 @@
 let queue: Promise<unknown> = Promise.resolve();
 
 export function serialize<T>(fn: () => Promise<T>): Promise<T> {
-    const res = queue.then(() => fn());
+    const res = queue.catch(() => {}).then(() => fn());
     queue = res.catch(() => {});
     return res;
 }
--- a/packages/user/Supervisor/ui/src/serialize.ts
+++ b/packages/user/Supervisor/ui/src/serialize.ts
@@ -1,0 +1,9 @@
+// Async serializer pattern:
+// https://advancedweb.hu/how-to-serialize-calls-to-an-async-function/
+let queue: Promise<unknown> = Promise.resolve();
+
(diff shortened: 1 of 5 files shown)


packages/user/Supervisor/ui/src/serialize.ts
```diff
@@ -0,0 +1,9 @@
+// Async serializer pattern:
+// https://advancedweb.hu/how-to-serialize-calls-to-an-async-function/
+let queue: Promise<unknown> = Promise.resolve();
+
+export function serialize<T>(fn: () => Promise<T>): Promise<T> {
+    const res = queue.then(() => fn());
+    queue = res.catch(() => {});
+    return res;
+}
```
cool-ant, line 12: Pls look closely at / consider modeling this implementation on my PR's promise management.
My impl handles things in a subtly-different, slightly more robust way:

In `serialize`, error recovery happens only when updating `queue` (L7), not when consuming it. If `queue` is in a rejected state at the start of a call, `fn()` won't get executed (the promise chain will skip the `then()`), and the caller receives the rejection from the earlier state. `runExclusive` prevents that by doing a `catch` on the lock before invoking `fn`, so any unexpected error state is cleared, thereby ensuring `fn` gets run regardless of prior failures.

The ways such a rejected state can happen are admittedly few and very corner casey, but it is a slightly more robust impl.

My impl also has another subtlty that show up in typing... above, `queue` holds whatever result fn() returned, i.e., Promise<number>, where mine restores it always to a reliable Promise<void> state.

cool-ant (changes requested):

packages/user/Supervisor/ui/src/serialize.ts
James-Mart: Good points - fixed

cool-ant (approved):

PR #1930, opened 2026-07-06
Vserver resource balance caching and db-writes during reconciliation

James-Mart:
This PR clarifies how resource token balances are cached in `vserver`. It consolidates the warming of the cache for all accounts whose resource balances might be touched during a transaction into a `prestartTx` hook called by transact (symmetric with `finishTx`). The accounts whose resource balances might be touched include both user accounts as the sender of any action in the tx, as well as the relay subaccount for any capacitylimited resources (which may be "billed" when the user frees any corresponding resource and is therefore refunded from the relay).

Furthermore, this PR clarifies the new rule used in `vserver::reconcile` (the settlement of the payments accrued during tx processing): Any records written during reconciliation either need to have been pre-created (particularly if they are user-controlled records), or they must be attributed to the system account. This avoids the reconciliation process itself incurring subsequent disk-write billing, which would violate an invariant.

### DIFF
--- a/packages/system/Transact/src/Transact.cpp
+++ b/packages/system/Transact/src/Transact.cpp
@@ -633,9 +633,8 @@ namespace SystemService
       check(trx.actions().size() > 0, "transaction has no actions");
       auto _           = recurse();
       bool enforceAuth = checkTapos(id, trx.tapos(), true, false);
-      if (enforceAuth) {
+      if (enforceAuth)
          checkAuth(trx.actions().front(), trx.claims(), true, true);
-      }
       return enforceAuth;
    }
 
@@ -705,9 +704,8 @@ namespace SystemService
       if (enforceAuth && isResMonitoring())
       {
          auto actors = std::vector<AccountNumber>();
-         actors.reserve(trx.actions().size());
-         for (auto act : trx.actions())
-            actors.push_back(act.sender().unpack());
+         actors.reserve(1);
+         actors.push_back(trxSender.value());
          to<VirtualServer>().prestartTx(actors);
       }
 
--- a/packages/system/Transact/src/Transact.cpp
+++ b/packages/system/Transact/src/Transact.cpp
@@ -633,8 +633,15 @@ namespace SystemService
       check(trx.actions().size() > 0, "transaction has no actions");
       auto _           = recurse();
       bool enforceAuth = checkTapos(id, trx.tapos(), true, false);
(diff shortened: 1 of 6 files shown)


packages/system/VirtualServer/service/src/tables/capacity_limit_pricing.rs
```diff
@@ -459,11 +459,21 @@ impl CapacityPricing {
     pub(crate) fn settle_relay(resource: ResourceType, delta: i128) {
         let sys = BillingConfig::get_assert().sys;
         let sub = relay_sub(resource);
+        if delta == 0 {
+            return;
+        }
+
+        // [DB WRITE] Possible allocation - Therefore must be attributed to the system account.
+        // Consider finding a way to guarantee avoiding this allocation, which would avoid the need to
```
swatanabe, line 467: Isn't this what `createSub` is for?

packages/system/Transact/src/Transact.cpp
```diff
@@ -701,6 +702,15 @@ namespace SystemService
       Actor<CpuLimit> cpuLimit(Transact::service, CpuLimit::service, CallFlags::runModeRpc);
       Actor<Accounts> accounts(Transact::service, Accounts::service);
 
+      if (enforceAuth && isResMonitoring())
+      {
+         auto actors = std::vector<AccountNumber>();
+         actors.reserve(trx.actions().size());
+         for (auto act : trx.actions())
+            actors.push_back(act.sender().unpack());
+         to<VirtualServer>().prestartTx(actors);
```
swatanabe: Why load all senders when only the first sender is billed?

packages/system/VirtualServer/service/src/tables/capacity_limit_pricing.rs
James-Mart: Unfortunately `createSub` is not sufficient. It is already called in `init`, and it creates the `SubAccountTable` record, but it doesn't create the `SubAccountBalanceTable` record. 

I would have to add some way to create a zero-balance record for the system token (after the system token is initialized).

packages/system/Transact/src/Transact.cpp
James-Mart: Updated, the interface now only takes a single actor (and transact only passes sender of the first action)

swatanabe (approved):

PR #1931, opened 2026-07-06
recollateralize relay shortfall from enable-billing over time

James-Mart:
Before this PR, any disk curve shortfall that accumulated due to system billing was recollateralized over time by taking the fees that would otherwise be sent to the fee receiver, with one exception:

When billing was first enabled, there would be a large undercollateralization due to the initial byte allocations that happened before billing was enabled. Historically this has been handled by requiring the caller of enable-billing to pay for the cost of settling this shortfall atomically with the enabling of billing.

This PR unifies this exception with the normal recollateralization strategy.

To do so, we must do the following:
* After measuring the preallocated bytes, we immediately update the curve to acknowledge the consumption.
* When billing is enabled, we "do nothing" (remove the code that required a payment and instantaneously settled). 

With these changes, the consumed capacity on the curve already represents the full tally of bytes that have been written. The curve reports a large shortfall, and will slowly recollateralize over time from fee-receiver fees.

### DIFF
--- a/packages/system/VirtualServer/service/src/lib.rs
+++ b/packages/system/VirtualServer/service/src/lib.rs
@@ -380,11 +380,9 @@ mod tx_cache {
 /// 2 - Call `init_billing`: Initializes the billing system. To call this, the system token
 ///     must have already been set in the `Tokens` service. The specified `fee_receiver` will
 ///     receive all of the system token fees paid for resources by users.
-/// 3 - Call `enable_billing(true, Some(payer))`: When ready, calling this action enables the
-///     billing system. Two requirements: (a) the caller must have already filled their resource
-///     buffer because this action is itself billed, and (b) `payer` must have credited sufficient
-///     system tokens to this service to settle any net disk consumption that accumulated while
-///     billing was disabled (see the `enableBillingCost` GraphQL query for the required amount).
+/// 3 - Call `enable_billing(true)`: When ready, calling this action enables the
+///     billing system. The caller must have already filled their resource buffer because
+///     this action is itself billed.
 ///
 /// > Note: typically, step 1 should be called at system boot, and therefore the network should
 /// >       always at least have some server specs and derived network specs.
@@ -563,18 +561,9 @@ mod service {
     /// If billing is disabled, resource consumption will still be tracked, but the resources will
     /// not be automatically metered by the network. This is insecure and allows users to abuse
     /// the network by consuming all of the network's resources.
-    ///
-    /// `payer` is required only when `enabled = true`: that account must have previously credited
-    /// sufficient system tokens to this service to cover the settlement cost of any net disk
-    /// consumption that occurred while billing was disabled.
     #[action]
-    fn enable_billing(enabled: bool, payer: Option<AccountNumber>) {
+    fn enable_billing(enabled: bool) {
         check(get_sender() == get_service(), "Unauthorized");
-
-        if enabled {
-            let payer = check_some(payer, "payer is required when enabling billing");
-            billing::settle_relay_balance(Disk, payer);
-        }
 
         BillingConfig::enable(enabled);
     }
--- a/packages/system/VirtualServer/service/src/lib.rs
+++ b/packages/system/VirtualServer/service/src/lib.rs
@@ -1130,6 +1130,8 @@ mod service {
 
         tx_cache::set_billable_account(account);
 
+        warm_billable_account_caches();
+
         CpuLimit::rpc().setCpuLimit(get_cpu_limit(account, None));
     }
 
--- a/packages/system/VirtualServer/service/src/lib.rs
+++ b/packages/system/VirtualServer/service/src/lib.rs
@@ -240,6 +240,7 @@ mod tx_cache {
             use super::is_system_user;
             use crate::resource_type::ResourceType;
             use crate::tables::capacity_limit_pricing::relay_sub_account;
+            use crate::tables::tables::CapacityPricing;
             use crate::tx_cache::{balance_cache, drain_cache, with_cache};
             use crate::Wrapper;
             use psibase::AccountNumber;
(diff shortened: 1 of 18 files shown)


packages/system/VirtualServer/service/src/lib.rs
```diff
@@ -287,18 +287,30 @@ mod tx_cache {
...
-                    return;
+            ) -> u64 {
+                let total_out = user_refund + fee;
+                if is_system_user(user) || total_out == 0 {
+                    return 0;
                 }
+                let collateral = get_collateral(resource);
+
+                let (gross, clamped_fee, refund) = if total_out > collateral {
+                    // Clamp the fee and the refund (keeping fee_ppm ratio)
```
swatanabe: Is this the right formula? I expected it to prioritize the refund. Reducing the fee and refund proportionally means that freeing memory across several transactions can get a higher refund than freeing it all at once.

James-Mart: > Is this the right formula? 

I'm okay with either. My justification for the rule as written is that I find it easier to understand:

* The refund for the sale of bytes is the byte value in the relay clamped by available collateral, then divided up by the configured fee %.

> Reducing the fee and refund proportionally means that freeing memory across several transactions can get a higher refund than freeing it all at once.

Worth noting that it's still speculative, though. You may also get a lower refund than freeing it all at once if other bytes are net freed between those transactions.

swatanabe: I think it makes more sense to divide the refund first and then clamp the amount sent to the client. Tokens are only sent to the fee receiver when the relay is fully capitalized which is a stronger restriction than having any tokens in the relay.

I don't understand what you mean by "speculative". The way to optimize the refund can be calculated exactly. If we clamp first, each successive transaction can get refunds of 1/2, 1/4, 1/8, etc of the original available collateral (for refund = 50%). Of course, other transactions can always affect refunds and fees, but that's a separate issue.

swatanabe: Also, if the relay is undercapitalized, the total released is split between the client's refund and reducing the shortfall. The holds even when we reduce the refund because there are insufficient tokens. If there are 4 SYS in the relay and the curve position is reduced by 8 SYS and the client receives 2 SYS, the shortfall is reduced by 6 SYS. If we refund the user 4 SYS, the shortfall is reduced by 4 SYS.

James-Mart: > I think it makes more sense to divide the refund first and then clamp the amount sent to the client. Tokens are only sent to the fee receiver when the relay is fully capitalized which is a stronger restriction than having any tokens in the relay.

Done

packages/system/VirtualServer/service/src/lib.rs
```diff
@@ -368,18 +374,81 @@ mod tx_cache {
...
+///
+/// ```text
+///            .---------------.
+///           (   free bytes    )
+///            '-------+-------'
+///                    |
+///                    v
+///      +--------------------------+
+///      | Reduce curve shortfall   |
+///      +------------+-------------+
```
swatanabe: I thought reducing the curve shortfall happens later.

packages/system/VirtualServer/service/src/lib.rs
```diff
@@ -368,18 +374,81 @@ mod tx_cache {
...
+///                        v
+///               .-----------------.
+///              /                   \
+///      yes    /  Is curve under-    \     no
+///    .-------(   collateralized?     )-------.
+///    |        \                     /        |
+///    |         \                   /         |
+///    v          '-----------------'          v
+/// +----------------------+     +----------------------+
+/// | Fee covers shortfall |     | Send full fee to fee |
```
swatanabe: The amounts by which the fee and the refund were reduced also help cover the shortfall, even though they aren't explicitly represented.

packages/system/VirtualServer/service/src/lib.rs
```diff
@@ -368,18 +374,81 @@ mod tx_cache {
...
+/// of reserve tokens required to be held as collateral for the current utilization levels of the
+/// resource.
+///
+/// This service attempts to balance the required collateral as reported by the state of the curve
+/// with the actual collateral as measured by the relay subaccount token balance. The curve for a
+/// capacity-limited resource can be undercollateralized in certain well-defined scenarios (e.g.
+/// consumption before billing is enabled).
+///
+/// An undercollateralized curve is recollateralized over time using the fees that would otherwise
+/// be sent to the fee receiver.
```
swatanabe: This isn't quite accurate and I think some of the confusion in the edge case of running out of collateral is a result of this inaccuracy. The undercollateralization is not fixed by increasing the actual collateral. It's fixed by reducing the required collateral.

packages/system/VirtualServer/service/src/lib.rs
```diff
@@ -287,18 +287,24 @@ mod tx_cache {
...
-                if is_system_user(user) || gross == 0 {
-                    return;
+            ) -> u64 {
+                let total_out = user_refund + fee;
+                if is_system_user(user) || total_out == 0 {
+                    return 0;
                 }
+                let collateral = get_collateral(resource);
+                let refund = user_refund.min(collateral);
+                let clamped_fee = fee.min(collateral.saturating_sub(refund));
```
swatanabe: I think this still isn't quite right. If the curve is undercollateralized, the entire collateral should be available for refunds, instead of reserving some amount as fees, only to put it back at the end of the transaction.

packages/system/VirtualServer/service/src/lib.rs
James-Mart: I was referring to the instantaneous drop in required collateral, but it's confusing. Removed.

packages/system/VirtualServer/service/src/lib.rs
James-Mart: Aha, fixed.

packages/system/VirtualServer/service/src/lib.rs
James-Mart: Reworked

packages/system/VirtualServer/service/src/lib.rs
James-Mart: Reworded

swatanabe (approved):

Issue #977, opened 2025-01-06
CSP Policy review


James-Mart:
Consider:
https://infosec.mozilla.org/guidelines/web_security.html#cookies

Lots of important security information here. CSP headers on supervisor domain should be more strict, probably other actionables generated from reviewing this.

James-Mart:
Brandon to define policy, someone else can add to postinstall script to set defaults

PR #1935, opened 2026-07-08
CSP - Policy defaults and app-level overrides

brandonfancher:
Addresses #977. Leverages new `{root}` keyword in #1936.

This starts with a very restrictive default. The idea is that developers should be forced to explicitly loosen defaults; not tighten them.

See [csp-recommendations.md](https://github.com/gofractally/psibase/blob/749379d0f57ab71490359b9fca07098235b12077/csp-recommendations.md) in this branch for a deep dive into this work and the thinking behind it. This file will be deleted before merge.

We should be very close. If after merging we get CSP warnings in the browser with accompanying errors, we can further adjust.


packages/user/FractalCore/postinstall.json
```diff
@@ -4,5 +4,14 @@
...
     "method": "enablespa",
     "data": { "enable": true }
+  },
+  {
+    "sender": "fractal-cr",
+    "service": "sites",
+    "method": "setcsp",
+    "data": {
+      "path": "*",
+      "csp": "default-src 'self';script-src 'self';style-src 'self' 'unsafe-inline';img-src 'self' data: profiles.{{root}}/avatar/ branding.{{root}};font-src 'self';connect-src 'self' guilds.{{root}} invite.{{root}} staged-tx.{{root}} profiles.{{root}};frame-src supervisor.{{root}};frame-ancestors 'self';base-uri 'none';form-action 'self';object-src 'none';"
```
Velua, line 14: Do we not need `fractals` or `evaluations` here then?

packages/user/Explorer/postinstall.json
```diff
@@ -0,0 +1,11 @@
+[
+  {
+    "sender": "explorer",
+    "service": "sites",
+    "method": "setcsp",
+    "data": {
+      "path": "*",
+      "csp": "default-src 'self';script-src 'self' 'unsafe-inline';style-src 'self' 'unsafe-inline';img-src 'self' data: profiles.{{root}}/avatar/ branding.{{root}};font-src 'self';connect-src 'self' tokens.{{root}};frame-src supervisor.{{root}};frame-ancestors 'self';base-uri 'none';form-action 'self';object-src 'none';"
```
Velua, line 8: Think `tokens` might be good but I suppose no one uses Explorer anyway

packages/user/TokenStream/service/Cargo.toml
```diff
@@ -9,6 +9,7 @@ publish.workspace = true
 postinstall = [
     { sender = "tok-stream", service = "sites", method = "enablespa", data = { enable = true } },
     { sender = "tok-stream", service = "tokens", method = "setUserConf", data = { index = 0, enabled = false } },
+    { sender = "tok-stream", service = "sites", method = "setcsp", data = { path = "*", csp = "default-src 'self';script-src 'self';style-src 'self' 'unsafe-inline';img-src 'self' data: profiles.{{root}}/avatar/ branding.{{root}};font-src 'self';connect-src 'self' token-stream.{{root}} nft.{{root}} tokens.{{root}} profiles.{{root}};frame-src supervisor.{{root}};frame-ancestors 'self';base-uri 'none';form-action 'self';object-src 'none';" } },
```
Velua, line 12: `token-stream.{{root}}` does self cover that already? Also, I think it's `tok-stream` given the new max length limitations, the service was renamed.

Velua (dismissed):

doc/src/default-apps/sites.md
```diff
@@ -10,17 +10,31 @@ The `sites` service is used to host content for an account and serve it over HTT
...
+Notable consequences of the strict baseline:
+
+- Inline `<script>` tags, `eval`, and CDN-hosted scripts are blocked by default. Apps that need them must opt in via `setCsp`.
+- Pages may only be embedded same-origin by default (`frame-ancestors 'self'`). Pages designed to be embedded by the supervisor (e.g. plugin prompt pages) must widen `frame-ancestors` to include the supervisor's origin.
+- The only frameable origin is the supervisor (every app embeds the hidden supervisor iframe).
+- `fetch` and WebSocket connections are limited to the app's own origin plus the `profiles` and `tokens` services. `'self'` covers an app's own `/graphql` and RPC endpoints and the `/common/*` endpoints (served same-origin on every subdomain). `profiles.{{root}}` and `tokens.{{root}}` are granted by default because the shared UI hooks (`useProfile`, `useSystemToken`) fetch them from nearly every app; both expose only public data, and writes require signed transactions. Apps that fetch any other service's subdomain directly must opt in via `setCsp` with an explicit host list (e.g. `connect-src 'self' profiles.{{root}} tokens.{{root}} invite.{{root}}`); no subdomain wildcard is granted by default.
```
James-Mart: > `fetch` and WebSocket connections are limited to the app's own origin plus the `profiles` and `tokens` services

If we made all queries (with text/json response type) through an authenticated interface (through a plugin call), we  could remove `profiles` and `tokens` from `connect-src`. 

In fact, `connect-src` only being `self` really basically forces everyone to use authenticated queries by default, which is great. Means the path of least resistance for making queries is authenticated & billed.

doc/src/default-apps/sites.md
```diff
@@ -10,17 +10,31 @@ The `sites` service is used to host content for an account and serve it over HTT
...
+- Cross-subdomain image loads are limited to user avatars (served by the `profiles` service at `profiles.{{root}}/avatar/<account>`) and the network logo (served from `branding.{{root}}`). Apps that display images from other origins must opt in via `setCsp`.
```
James-Mart: A variant of "[if your app has more requirements you] must opt in via `setCsp`." is repeated in most of these lines. I think a general comment about loosening restrictions via `setCsp` could just be stated once (or not at all, since the next section is about loosening CSP), and these bullets could then be more succinct.

doc/src/default-apps/sites.md
```diff
@@ -10,17 +10,31 @@ The `sites` service is used to host content for an account and serve it over HTT
...
+| `frame-ancestors` | `'self'`                                                     |
+| `base-uri`        | `'none'`                                                     |
+| `form-action`     | `'self'`                                                     |
+| `object-src`      | `'none'`                                                     |
+
+Notable consequences of the strict baseline:
+
+- Inline `<script>` tags, `eval`, and CDN-hosted scripts are blocked by default. Apps that need them must opt in via `setCsp`.
+- Pages may only be embedded same-origin by default (`frame-ancestors 'self'`). Pages designed to be embedded by the supervisor (e.g. plugin prompt pages) must widen `frame-ancestors` to include the supervisor's origin.
+- The only frameable origin is the supervisor (every app embeds the hidden supervisor iframe).
```
James-Mart: Technically it's not a requirement that "every app" embeds a hidden supervisor iframe. It's the normal path for apps that make plugin calls, but static sites and other custom apps may not need to.

packages/system/Accounts/postinstall-ui.json
```diff
@@ -4,5 +4,14 @@
...
         "method": "enablespa",
         "data": { "enable": true }
+    },
+    {
+        "sender": "accounts",
+        "service": "sites",
+        "method": "setcsp",
+        "data": {
+            "path": "*",
+            "csp": "default-src 'self';script-src 'self';style-src 'self' 'unsafe-inline';img-src 'self' data: profiles.{{root}}/avatar/ branding.{{root}};font-src 'self';connect-src 'self' tokens.{{root}} namemarket.{{root}};frame-src supervisor.{{root}};frame-ancestors 'self' supervisor.{{root}};base-uri 'none';form-action 'self';object-src 'none';"
```
James-Mart, line 14: Don't think `name-market` (or `tokens`) are needed in `connect-src` if you make the queries authenticated

packages/user/Config/postinstall.json
```diff
@@ -4,5 +4,14 @@
...
     "method": "enablespa",
     "data": { "enable": true }
+  },
+  {
+    "sender": "config",
+    "service": "sites",
+    "method": "setcsp",
+    "data": {
+      "path": "*",
+      "csp": "default-src 'self';script-src 'self';style-src 'self' 'unsafe-inline';img-src 'self' data: profiles.{{root}}/avatar/ branding.{{root}};font-src 'self';connect-src 'self' producers.{{root}} sites.{{root}} staged-tx.{{root}} transact.{{root}} vserver.{{root}} tokens.{{root}} profiles.{{root}} x-admin.{{root}} packages.{{root}} namemarket.{{root}} http: https:;frame-src supervisor.{{root}};frame-ancestors 'self';base-uri 'none';form-action 'self';object-src 'none';"
```
James-Mart, line 14: Instead of continuing to comment this on each post-install script, my general feedback is that if there's a bunch of affordances in the connect-src, it's probably a sign that we aren't making authenticated queries where we should be. 

Not *all* custom `connect-src` can be removed, for example this has `x-admin` for which there can be no authenticated query interface, but many of them can/should.

csp-recommendations.md
James-Mart, line 1: Did you intend to commit this? 

If there's helpful pieces you want to preserve from this, I don't think this info should be in a root level file. It should be split between doc comments in the Sites service, and the mdbook in `/doc/src/`.

James-Mart (commented):
Many of my comments are about authenticated queries. 

I think one strategy here would be to deliver this with all the various `connect-src` affordances, but add an issue to come back and tighten the defaults further after converting a bunch of direct fetch requests into authenticated queries.

csp-recommendations.md
brandonfancher: I did, as a way to get us all on the same page quickly. It should be deleted before merge. And yes, with proper documentation added in the right places. Sorry, I should have made that more explicit in the description.

James-Mart:
Related
https://github.com/gofractally/psibase/issues/2015

packages/user/FractalCore/postinstall.json
brandonfancher: Yes, good catch.

packages/user/Explorer/postinstall.json
brandonfancher: Yeah, we need to rewrite Explorer.

packages/user/TokenStream/service/Cargo.toml
brandonfancher: Yes, it does. And good catch on the rename.

doc/src/default-apps/sites.md
brandonfancher: Issue created for follow up

doc/src/default-apps/sites.md
brandonfancher: Improved (removed)

packages/system/Accounts/postinstall-ui.json
brandonfancher: Resolving since we agreed to tackle this as a follow up

packages/user/Config/postinstall.json
brandonfancher: Resolving since we agreed to tackle this as a follow up

doc/src/default-apps/sites.md
brandonfancher: Fixed

doc/src/default-apps/sites.md
brandonfancher: Resolving since we agreed to tackle this as a follow up

csp-recommendations.md
James-Mart: >  It should be deleted before merge

Ok, I won't approve until it gets removed then

James-Mart (commented):
I don't have any more feedback

csp-recommendations.md
brandonfancher: Okay, it is now removed

James-Mart (approved):

PR #1937, opened 2026-07-08
Unpack nested actions in traces

swatanabe:


rust/psibase/src/trace.rs
```diff
@@ -402,26 +424,66 @@ impl<'a, F: SchemaFetcher + 'a> ActionFormatter<'a, F> {
...
+                        rawData: raw_data.into(),
+                    });
+                }
+            });
+            custom.insert("Action".to_string(), &custom_action);
+            let mut cschema = CompiledSchema::new(&schema.types, &custom);
+            let action_name = MethodString(act.method.to_string());
+            if let Some(action_type) = schema.actions.get(&action_name) {
+                cschema.extend(&action_type.params);
+                let _ = cschema.verify(&action_type.params, act.rawData.0);
```
cool-ant, line 460: likely very rare (and perhaps captured in a test somewhere), but if packed structure and schema don't match, the error would not propagate here, right?

rust/psibase/src/trace.rs
```diff
@@ -796,17 +846,99 @@ impl<'a> CustomHandler for CustomResolvedServiceMethod<'a> {
...
+            }
+        }
+        let mut result = serde_json::Map::with_capacity(4);
+        result.insert(sender_name.to_owned(), value.sender.to_string().into());
+        result.insert(service_name.to_owned(), value.service.to_string().into());
+        result.insert(method_name.to_owned(), method.0.into());
+        if let Some(data) = data {
+            result.insert("data".to_string(), data);
+        } else {
+            result.insert(raw_data_name.to_owned(), value.rawData.to_string().into());
```
cool-ant, line 920: minor inconsistency (I think):
This line seems to be an improvement over L608-612 / 623; perhaps we update that to match this label (and output `raw_data_name.to_owned()`)?

rust/psibase/src/trace.rs
```diff
@@ -214,8 +215,10 @@ fn format_action_trace(
     )?;
     if let Some(action_type) = action_type {
         let mut custom = schema_types();
-        let service_method = CustomResolvedServiceMethod::new(schemas);
+        let service_method = CustomResolvedServiceMethod::new(schemas, std::mem::drop);
         custom.insert("ServiceMethod".to_string(), &service_method);
+        let custom_action = CustomActionCollector::new(&*schemas, drop_action);
+        custom.insert("Action".to_string(), &custom_action);
         let mut cschema = CompiledSchema::new(&schema.unwrap().types, &custom);
```
cool-ant, line 222: These lines are repeated a few times. Perhaps a helper macro? like this
```
macro_rules! trace_compiled_schema {
    ($schemas:expr, $types:expr, $on_service_method:expr, $on_action:expr, $cschema:ident $(,)?) => {
        let mut custom = schema_types();
        let service_method = CustomResolvedServiceMethod::new($schemas, $on_service_method);
        custom.insert("ServiceMethod".to_string(), &service_method);
        let custom_action = CustomActionCollector::new($schemas, $on_action);
        custom.insert("Action".to_string(), &custom_action);
        let mut $cschema = CompiledSchema::new($types, &custom);
    };
}
```

cool-ant (approved):
none of my comments are critical. just options

rust/psibase/src/trace.rs
swatanabe: Right. Errors don't really matter here. They just mean that we'll be missing some schemas and won't be able to unpack some data later.

rust/psibase/src/trace.rs
swatanabe: This is to get the names from the schema. `UnpackedAction` is used for the top level of the trace, which isn't dependent on a dynamic schema.

PR #1938, opened 2026-07-10
Fix FractalGen

### LINKED ISSUE #1929: FractalGen failing to boot
There's a few issues with `FractalGen` and `fractals` in general in `main` probably created from the introduction of sub accounts.

On boot when including `FractalGen` there's the error `"The account  required by FractalGen is not defined by any package"` I imagine this is created because `fractal-gen` is 11 chars when our new limit is 10.
Also, services are no longer allowed to read tables of other services directly right? Services like `de-facto` and `guilds` read `fractals` directly. While `guilds` reads `evaluation`

I think we can resolve `de-facto` by just making it's service `fractals+2` however how do we wanna tackle `guilds` and `evaluations`? Should I be going through their use and start offering public actions on the relevant services for the others to read the relevant data? `Guilds` will currently reads `EvaluationTable` of the `evaluation` service, so this information could be exposed through a new action?

Velua:
- Evaluations UI was using `evaluations` still rather than the new `evaluation` name.
- Introduce new actions for `fractals` `evaluation` and `guilds` to use. 
- Rename `fractal-gen` to `frac-gen`
- Update dependencies


CMakeLists.txt
```diff
@@ -545,8 +545,8 @@ cargo_psibase_package(
 )
 
 cargo_psibase_package(
-    OUTPUT ${SERVICE_DIR}/FractalGen.psi
-    PATH packages/user/FractalGen
+    OUTPUT ${SERVICE_DIR}/FracGen.psi
+    PATH packages/user/FracGen
```
swatanabe: I don't think we need to rename directories and packages whenever we change a service name. I'd rather keep the names the don't need to be short unobfuscated.

swatanabe (approved):

PR #1940, opened 2026-07-13
Adopt sub accounts in Fractal services

Velua:
No longer independent accounts as role accounts for Fractals and Guilds but using sub accounts instead.


packages/user/FractalGen/service/src/lib.rs
```diff
@@ -35,7 +35,7 @@ mod service {
     use psibase::*;
 
     const SYS_FRACTAL: AccountNumber = account!("core-fract");
-    const SYS_GUILD: AccountNumber = account!("core-gld-1");
+    const SYS_GUILD: AccountNumber = account!("guild-onee");
```
James-Mart: Should that be `guild-one`?
If you added an extra `e` because of a minimum character limit, you can just go ahead and lower the number of characters that are creatable in the accounts service so that you can use `guild-one`. I think we said >=8 chars are free

packages/user/FractalGen/Cargo.toml
```diff
@@ -13,8 +13,9 @@ package-name = "FractalGen"
 services = ["fractal-gen"]
 
 [package.metadata.psibase.dependencies]
-Sites = "0.23.0"
+DynLd = "0.23.0"
 HttpServer = "0.23.0"
+Sites = "0.23.0"
```
James-Mart: Deps are unioned between top level Cargo.tom and service Cargo.toml, don't need to repeat them (They should live where they logically make the most sense in any particular case)

packages/user/Fractals/plugin/src/lib.rs
```diff
@@ -53,15 +52,7 @@ impl AdminFractal for FractallyPlugin {
 
         let fractal = fractal_account.parse().unwrap();
 
-        Fractals::add_to_tx().create_frac(
-            fractal,
-            gen_rand_account(Some("leg"))?.parse().unwrap(),
```
James-Mart, line 58: Should we just just delete `gen_rand_account` capability from `accounts` plugin?

packages/user/Fractals/service/src/tables/fractal.rs
```diff
@@ -33,6 +33,10 @@ use psibase::{get_sender, RawKey, TableQuery, TimePointSec};
 
 impl Fractal {
     fn new(account: AccountNumber, name: String, mission: String) -> Self {
+        assert!(
+            !account.is_subaccount(),
```
James-Mart, line 37: Interesting. This makes me wonder if checks like this are going to end up everywhere.
I guess the criteria is that a subaccount can't be used anywhere that the account needs to have an associated domain name

rust/psibase/src/services/guilds.rs
```diff
@@ -1,3 +1,25 @@
+use crate::Subaccount;
+
+#[repr(u8)]
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+pub enum GuildSubaccount {
```
James-Mart: I think this should be called `GuildRole`. So you have `GuildRole::to/from_subaccount()` instead of `GuildSubaccount::to/from_subaccount`

rust/psibase/src/services/guilds.rs
```diff
@@ -1,3 +1,25 @@
...
+    Base = 0,
+    Council = 1,
+    Rep = 2,
+}
+
+impl GuildSubaccount {
+    pub fn subaccount(self) -> Subaccount {
```
James-Mart: And for symmetry with `from_subaccount` this should be called `to_subaccount`.

packages/user/Guilds/service/src/tables/guild.rs
```diff
@@ -113,21 +124,35 @@ impl Guild {
...
     }
 
+    pub fn by_council_sender() -> Self {
+        let (guild, sub_account) = get_sender().split();
+        let guild = Self::get_assert(guild);
+        assert_eq!(
+            sub_account,
+            GuildSubaccount::Council.subaccount(),
+            "sender must be council role account of guild",
+        );
```
James-Mart, line 136: Makes me think a helper might be clearer. Whole function might be something like 
```
GuildSubaccount::Council.assert_match(get_sender());
Self::get_assert(get_sender().base())
```

packages/user/Guilds/service/src/lib.rs
```diff
@@ -488,10 +470,14 @@ pub mod service {
...
     fn account_policy(account: AccountNumber) -> Option<auth_dyn::policy::DynamicAuthPolicy> {
-        Guild::get(account)
-            .map(|guild| guild.auth_policy())
-            .or(Guild::get_by_council_role(account).map(|guild| guild.council_role_auth()))
-            .or(Guild::get_by_rep_role(account).map(|guild| guild.rep_role_auth()))
+        let (base, subaccount) = account.split();
+
+        let guild = Guild::get(base)?;
+        GuildSubaccount::from_subaccount(subaccount).map(|subaccount| match subaccount {
+            GuildSubaccount::Base => guild.auth_policy(),
```
James-Mart: I don't think `Base` belongs in `GuildSubaccount`.

The logic should be more like 
```
        let (base, subaccount) = account.split();
        let guild = Guild::get(base)?;

        if !account.is_subaccount() {
            return Some(guild.auth_policy());
        }

        GuildSubaccount::from_subaccount(subaccount).map(|role| match role {
            GuildSubaccount::Council => ...,
            GuildSubaccount::Rep => ...,
        })
```

packages/user/Fractals/service/src/tables/role.rs
```diff
@@ -63,10 +49,17 @@ impl Role {
...
 
-        let new_instance = Self::new(fractal, account, role, occupation);
+        let new_instance = Self::new(fractal, role, occupation);
         new_instance.save();
+        let new_account = new_instance.account();
 
-        create_managed_account(account, || {});
+        accounts::Wrapper::call_as(fractal).newAccount(
+            new_account,
+            auth_any::Wrapper::SERVICE,
```
James-Mart, line 54: We are trying to get away from using `auth-any`. We should use `auth-delegate`, set the owner to `get_service()`. `call_as` should still work.

.gitignore
```diff
@@ -25,6 +26,7 @@ Makefile
 node_modules
 temp
 dist
+.direnv
```
James-Mart, line 30: What uses this, zed? The direnv config in contributor doesn't generate a .direnv file, just curious why its suddenly needed

Velua: Its a nix develop thing, I made a seperate PR for it's introduction.

packages/user/Fractals/plugin/src/lib.rs
Velua: Removed

James-Mart:
afaik this is still waiting for changes. Or do you want me to review again? Can you either convert this to a draft or re-request a review from me so it's clear whose court it's in?

packages/user/Fractals/service/src/helpers/create_managed_account.rs
```diff
@@ -1,11 +1,21 @@
...
+/// Subaccounts must be created by their parent, so those are preapproved
+/// first. The parent must already exist and this service must be able to
+/// `call_as` it (auth-delegate owned by this service).
 pub fn create_managed_account<F>(new_account: AccountNumber, f: F)
 where
     F: Fn(),
 {
     let self_service = get_service();
+    if new_account.is_subaccount() {
+        accounts::Wrapper::call_as(new_account.base()).preapproveAcc(new_account);
```
James-Mart: Nice

James-Mart (approved):
Sorry for the delay, looks good

PR #1941, opened 2026-07-13
Event access control

swatanabe:
Restrict access to events to the service that created them (and subaccounts). Events can be marked as public to make them visible to all service.


James-Mart (approved):
Sorry this review took forever. Would appreciate it if you would add a brief description to the PR to summarize the change.

PR #1946, opened 2026-07-15
Move VSCode scripts to shared dir

Velua:
Scripts like `build.sh` `continue.sh` and `launch.sh` aren't solely relevant to VSCode, moved them into a shared directory, updated the tasks.json.sample in `.vscode` to reflect this, and created a similar `tasks.json.sample` for [zed](https://zed.dev/)

Nix also creates profile stuff after running `nix develop` so `.direnv` is ignored in `.gitignore` as well.


James-Mart:
This make the scripts feel more canonical than they are.
Should the directory be `example_scripts` or something?

swatanabe:
> This make the scripts feel more canonical than they are. Should the directory be `example_scripts` or something?

Rather than `example_scripts` I think we should consider what these scripts are actually for:

- They are only usable in a canonical dev environment (nix or docker)
- They are intended to be linked to IDE buttons. Command line usage doesn't really benefit from using these scripts over running `psinode` directly.

Maybe put them in `.editor-shared/scripts` or something.

Velua:
Made the change @swatanabe @James-Mart

James-Mart (approved):

PR #1948, opened 2026-07-15
Limit block size

swatanabe:
With a release build, blocks could become too large for triedent.

Transaction are only applied when the current block size is below the limit. This allows one transaction over the limit, which will trigger price increases in VirtualServer. Before VirtualServer is installed, RTransact imposes a limit of 8MB.


James-Mart (approved):

James-Mart:
Classified as a dev exp bug because it presents as boot failing. If you think it should be something else, feel free to update the labels

PR #1950, opened 2026-07-17
WebRTC: Add objective Chat service: Spaces, sessions, and call timeline

cool-ant:
https://github.com/gofractally/psibase-pm/issues/23

This is PR1 in a chin of PRs to release the webrtc work. First phase is the Chat objective service, which handles
- Spaces: generalized term for groups of wrtc clients, commonly called "rooms" in a video conf context. Spaces determine who's allow to connect in a partiuclar instance. In the eventual app, this would be a particular DM pair or a group chat and can involve chat, file sends, and occasional video/audio calls.
- Sessions: media-specific structure for what and who are allowed to do what, e.g., a session for chat, that's associated with a Space.
- Call events: history of connection/call details, which allow for metadata in the UI, e.g., "Missed call", "Called ended".

## Architectural Decisions and their Impact
**Added in this PR:** the **Chat** package — objective service `chat` and `r-chat`. Plugin API included; there is no Homepage UI here.

---

**`chat` is an objective (on-chain) service.**  
Spaces, session metadata (`chat-data` / `av-call` / pair), and call-timeline events live in consensus state. Architectural impact:

1. **Membership and session auth are protected and durable** — who may join a Space/session is agreed across the network, not tied to one node’s memory.
2. **Third-party apps can reuse it as shared infra** — they don’t need to roll their own “rooms,” session ids, or join-authorization model; they call Chat (e.g. `authorizeSessionJoin`) and build UX on top.
3. **Message and media bodies stay out of this service** — Chat owns *who’s allowed* and *call/session history metadata*; actual chat bytes and A/V ride peer-to-peer (WebRTC) (later PRs), with subjective signaling (`x-wrtcsig` -- also future PR) brokering live presence.

In short: **Chat (this PR) = objective who’s-allowed / session + call metadata infra; realtime who’s-connected-now comes later as subjective `x-wrtcsig`.**

## Summary
- Adds the objective **Chat** package: DM/group Spaces, WebRTC session metadata (`chat-data` / `av-call` / pair sessions), and call-timeline events.
- Exposes plugin APIs (`ensure-dm`, `ensure-group`, `create-session`, `close-session`, …) and authenticated GraphQL (`mySpaces`, `activeSession`, `sessionJoinAuth`, `callEvents`, …).
- Wires Chat into package/CMake/`ctest` so the on-chain model can be built and tested on its own.

### via `ctest` / manual `cargo-psibase test`
There is **no Homepage UI** in this PR. Validation is via `cargo-psibase test` (primary) and optional GraphiQL on a node with Chat installed.

For convenience, here's the ctest command:
```bash
cargo-psibase test \
  --manifest-path packages/user/Chat/service/Cargo.toml \
  --target-dir /tmp/psibase-chat-target \
  --psitest=build/psitest
```

### on chain:
On a booted node with the Chat package installed:
- Open `https://chat.<host>/graphiql` authenticated as alice.
- Drive (plugin / push_action / GraphQL as available):
- Alice ensureDm(bob) → Expect: Space id returned.
- Alice createSession(spaceId, purpose: "chat-data"|"av-call") → Expect: session id; participants are the Space members.
- Query mySpaces / activeSession / sessionJoinAuth → Expect: alice↔bob membership and join auth match the Space.


packages/user/Chat/plugin/src/lib.rs
```diff
@@ -0,0 +1,150 @@
...
+fn purpose_to_chain(purpose: SessionPurpose) -> String {
+    match purpose {
+        SessionPurpose::ChatData => "chat-data".to_owned(),
+        SessionPurpose::AvCall => "av-call".to_owned(),
+    }
+}
+
+impl Spaces for ChatPlugin {
+    fn ensure_space(members: Vec<String>) -> Result<(), Error> {
+        trust::assert_authorized_with_whitelist(
```
swatanabe: The new way is the `authorized` derive macro

packages/user/Chat/plugin/src/lib.rs
```diff
@@ -0,0 +1,150 @@
...
+            trust::FunctionName::ensure_space,
+            vec!["homepage".into()],
+        )?;
+        let members = parse_members(members)?;
+        let packed = chat::action_structs::ensureSpace { members }.packed();
+        add_action_to_transaction(chat::action_structs::ensureSpace::ACTION_NAME, &packed)
+            .unwrap();
```
swatanabe: `add_to_tx` is less verbose

packages/user/Chat/service/src/lib.rs
```diff
@@ -0,0 +1,369 @@
...
+    };
+    use psibase::services::transact::Wrapper as TransactSvc;
+    use psibase::*;
+
+    fn now_micros() -> i64 {
+        TransactSvc::call()
+            .currentBlock()
+            .time
+            .seconds()
+            .microseconds()
```
swatanabe: Why truncate to seconds?

packages/user/Chat/service/src/lib.rs
```diff
@@ -0,0 +1,369 @@
...
+            .microseconds
+    }
+
+    fn abort_on_space_err(result: Result<(), SpaceError>) {
```
swatanabe: This should be spelled `unwrap`

packages/user/Chat/service/src/sessions/types.rs
```diff
@@ -0,0 +1,86 @@
...
+        match err {
+            SpaceError::UnknownSpace(msg) => SessionError::UnknownSpace(msg),
+            SpaceError::NotMember(msg) => SessionError::NotMember(msg),
+            SpaceError::InvalidMemberSet(msg) => SessionError::InvalidParticipants(msg),
+        }
+    }
+}
+
+impl SessionError {
+    pub fn message(&self) -> String {
```
swatanabe: Why a custom function instead of `Display` (or `thiserror`)

packages/user/Chat/service/src/lib.rs
```diff
@@ -0,0 +1,369 @@
...
+    }
+
+    fn abort_on_session_err<T>(result: Result<T, SessionError>) -> T {
+        match result {
+            Ok(value) => value,
+            Err(err) => abort_message(&err.message()),
+        }
+    }
+
+    fn session_view(session: sessions::Session) -> Session {
```
swatanabe: Why are there two identical `Session` types?

packages/user/Chat/service/src/spaces.rs
```diff
@@ -0,0 +1,251 @@
...
+        SPACE_KIND_GROUP
+    })
+}
+
+pub fn space_uuid_for_members(members: &[AccountNumber]) -> Result<String, SpaceError> {
+    validate_space_members(members)?;
+    let encoded_members: Vec<String> = members.iter().map(ToString::to_string).collect();
+    let bytes = serde_json::to_vec(&encoded_members)
+        .map_err(|err| SpaceError::InvalidMemberSet(err.to_string()))?;
+    Ok(format!("space:{}", sha256(&bytes)))
```
swatanabe, line 49: UUID has a specific meaning. This is not a UUID.

packages/user/Chat/service/src/sessions/purpose.rs
```diff
@@ -0,0 +1,64 @@
...
+        Ok(())
+    } else {
+        Err(SessionError::InvalidPurpose(format!(
+            "unsupported session purpose: {purpose}"
+        )))
+    }
+}
+
+/// For group `chat-data` and `av-call` sessions, objective metadata always includes all
+/// Space members (N-party mesh). DM and other purposes use the explicit participant list.
```
swatanabe: I'm confused.
- There is no other purpose
-  The only way for DM not to include all space members is to have only one participant. Is that even allowed?

packages/user/Chat/service/src/lib.rs
cool-ant: oh! good catch. Thanks

packages/user/Chat/service/src/lib.rs
cool-ant: These should be the same. This is an artificial/inappropriate "api" separation.

packages/user/Chat/service/src/sessions/types.rs
cool-ant: ah, ok. Pattern/standard noted. (and code updated)

packages/user/Chat/service/src/sessions/purpose.rs
cool-ant: hmm, I likely pruned one along the way. Reworded. And DMs now confirm both participatns, so the session memberhsip matches the  space membership

swatanabe: I don't understand why using the explicit participants list for other purposes is safer. If every existing purpose uses the space members, why would you expect a new purpose to be different? Is the explicit participants list needed at all?

packages/user/Chat/service/src/sessions/pair.rs
```diff
@@ -0,0 +1,55 @@
...
+        return None;
+    }
+    Some(vec![lower.parse().ok()?, higher.parse().ok()?])
+}
+
+pub fn is_pair_session_id(session_id: &str) -> bool {
+    parse_pair_session_id(session_id).is_some()
+}
+
+pub(crate) fn authorize_pair_session_join(
```
swatanabe: What is this for? It seems to be intended to be a static session that always exists for every pair of accounts, AFAICT, it's only used by the `session_join_auth` query. Nothing else that deals with sessions handles it, which makes it an odd special case.

rust/psibase/src/services/chat.rs
James-Mart, line 1: I'd like to get doc comments in here. It would help me review so I can get a quick high level understanding of the interface.

James-Mart:
Thoughts reading the pr description, not the code (Will start code review later today):

> Call events: history of connection/call details, which allow for metadata in the UI, e.g., "Missed call", "Called ended".

Interesting, implying making a calling and hanging up require a transaction 

>> yes. worth a discussion

> infra-public

>> Maybe we adopt "protected"? updated; this PR started before we made that decision.

> actual chat bytes and A/V ride peer-to-peer (WebRTC) (later PRs)

Are those future upgrades to the chat service? Or is the chat service only ever "objective who’s-allowed / session + call metadata infra"? If the latter, is `chat` really the right name for this service?

>> no, bytes will travel over the webrtc connection, which will be managed by a local service, not objective data, not changes to the chat service.
Right, only ever objctive who's-allowed / /session + call metadata".
reasonable question. It was the first name I chose at the start of this whole thing, so it can very likely be refined to be quite technical at this point where we know where things landed. let's discuss.

packages/user/Chat/service/src/sessions/purpose.rs
cool-ant: removed. participant list now relies on Spaces' lists

packages/user/Chat/service/src/sessions/pair.rs
cool-ant: It's mostly for a yet-to-come PR (this and every PR in this sequence are extracts from the massive/master branch). I'll just move this to the next PR. it'll make a little more sense there.

rust/psibase/src/services/chat.rs
cool-ant: added

PR #1953, opened 2026-07-21
Add nix module

Velua:
Adds a proper Nix deploy path for psibase:

- `packages.psibase` — pulls the release tarball, patches it for the store (x86_64-linux only, currently 0.25.0-pre)
- `nixosModules.psibase` — hardened systemd unit for psinode, optional SoftHSM for producers, loopback by default
- flake eval / VM checks so the module doesn’t rot
- pulled the devShell out of `flake.nix` into `nix/dev/` while we were in there

Usage sketch and options live under nix/deploy/.


nix/deploy/README.md
```diff
@@ -0,0 +1,108 @@
...
+```
+
+`-a`/`--api` is a subcommand argument, not a global one. It defaults to
+`http://psibase.localhost:8080/` and also reads `PSINODE_URL`, so it can be
+omitted when those match.
+
+## Not handled by this module
+
+TLS, reverse proxy and admin auth are yours. psinode does not authenticate
+`/native/admin/`, so terminate TLS and authenticate admin in a proxy (Caddy,
```
swatanabe: This isn't really true. `psinode` (actually XAdmin) provides several options for authenticating `/native/admin`.
- HTTP Basic auth is directly supported (XBasic)
- Once the chain is booted, on-chain accounts can be designated as admins and authorized by the usual cookie or `Authorization` header.
- IP addresses can be whitelisted (Only loopback by default). This is proxy-aware and will verify the client's address, even through a reverse proxy.
- Forwarded authorization from a trusted reverse proxy is possible, as below.

Requests to `x-admin` that do not meet any of these criteria will fail, so the default behavior of `psinode` is safe regardless of whether it is directly accessible or behind a reverse proxy.

Velua: Addressed claim

nix/deploy/module.nix
```diff
@@ -0,0 +1,358 @@
...
+      description = "Hostname for the service HTTP interface (psinode --host).";
+    };
+
+    listenAddress = lib.mkOption {
+      type = lib.types.str;
+      default = "127.0.0.1";
+      example = "0.0.0.0";
+      description = ''
+        IP address on which psinode listens. The loopback default keeps the
+        unauthenticated admin API behind a local reverse proxy. Set this to
```
swatanabe: The admin API is not unauthenticated

nix/deploy/module.nix
```diff
@@ -0,0 +1,358 @@
...
+      example = 120;
+      description = ''
+        Idle HTTP connection timeout in seconds (psinode --http-timeout).
+        Null uses `<dataDir>/db/config` (editable via x-admin). A value set
+        here overrides that file on every restart.
+      '';
+    };
+
+    pkcs11Module = lib.mkOption {
+      type = lib.types.nullOr lib.types.path;
```
swatanabe: Is a list allowed?

nix/deploy/module.nix
```diff
@@ -0,0 +1,358 @@
...
+        is true, defaults to SoftHSM's libsofthsm2.so unless overridden.
+      '';
+    };
+
+    openFirewall = lib.mkOption {
+      type = lib.types.bool;
+      default = false;
+      description = ''
+        Open the listen port in the firewall. Usually false: psinode does not
+        authenticate `/native/admin/`. Put a reverse proxy in front. To accept
```
swatanabe: Another instance of the false claim about the admin api

nix/deploy/README.md
```diff
@@ -0,0 +1,116 @@
...
+psinode, expose it explicitly:
+
+```nix
+services.psibase = {
+  listenAddress = "0.0.0.0";
+  openFirewall = true;
+};
+```
+
+Optional: pass the authenticated user from a trusted local proxy:
```
swatanabe: While this is useful for integrating `psinode` with an external authentication system, if you aren't constrained to use such a system, I don't recommend this approach. It can be tricky to get right.

nix/deploy/README.md
```diff
@@ -0,0 +1,116 @@
...
+
+`-a`/`--api` is a subcommand argument, not a global one. It defaults to
+`http://psibase.localhost:8080/` and also reads `PSINODE_URL`, so it can be
+omitted when those match.
+
+### Not handled by this module
+
+TLS and reverse proxy are yours. psinode listens on `127.0.0.1` by default;
+XAdmin already authenticates `/native/admin/` (loopback, HTTP Basic, on-chain
+admins, or a trusted proxy header). If a remote proxy connects directly to
```
swatanabe: I don't think a list of XAdmin authentication methods makes much sense here. If remote access to XAdmin is desired, HTTP Basic auth is the most straightforward approach.

nix/deploy/README.md
Velua: Added a comment

nix/deploy/README.md
Velua: removed

nix/deploy/module.nix
```diff
@@ -0,0 +1,354 @@
...
+        [cfg.package]
+        ++ lib.optionals cfg.softHsm.enable [
+          softhsmPkg
+        ];
+
+      networking.firewall.allowedTCPPorts = lib.mkIf cfg.openFirewall [cfg.listen];
+
+      # <dataDir>/db is created by triedent, not tmpfiles.
+      systemd.tmpfiles.rules =
+        ["d ${cfg.dataDir} 0750 psibase psibase -"]
```
swatanabe: The `systemd` documentation recommends using `StateDirectory` in the service configuration, instead of `tmpfiles`. How do we know that the directories are created before they're needed?

Velua:
@swatanabe do you have further feedback?

nix/deploy/module.nix
```diff
@@ -0,0 +1,354 @@
...
+
+        serviceConfig = {
+          Type = "oneshot";
+          RemainAfterExit = true;
+          # Root: read root-owned sops pin; create tokens as psibase.
+          ExecStart = pkgs.writeShellScript "psibase-softhsm-init" ''
+            set -euo pipefail
+            export SOFTHSM2_CONF=${lib.escapeShellArg softhsmConf}
+            mkdir -p ${lib.escapeShellArg cfg.softHsm.tokenDir}
+            chown psibase:psibase ${lib.escapeShellArg cfg.softHsm.tokenDir}
```
swatanabe: Isn't this directory already created by tmpfiles?

nix/deploy/module.nix
```diff
@@ -0,0 +1,354 @@
...
+      systemd.services.psibase = {
+        description = "Psibase node (psinode)";
+        wantedBy = ["multi-user.target"];
+        after = ["network-online.target"] ++ lib.optionals cfg.softHsm.enable ["psibase-softhsm-init.service"];
+        wants = ["network-online.target"];
+        requires = lib.optionals cfg.softHsm.enable ["psibase-softhsm-init.service"];
+
+        environment = cfg.environment // softhsmEnv;
+
+        serviceConfig = {
```
swatanabe, line 321: `Type` defaults to `simple`

> Typically, Type=exec is the better choice -- systemd.service(5)

The difference is that `Type = exec` will wait until `psinode` is running before considering the service to be successfully started.

nix/deploy/module.nix
```diff
@@ -0,0 +1,354 @@
...
+      systemd.services.psibase-softhsm-init = lib.mkIf cfg.softHsm.enable {
+        description = "Initialize SoftHSM token for psibase (once)";
+        after = ["local-fs.target"];
+
+        path = [pkgs.coreutils pkgs.gawk pkgs.util-linux softhsmPkg];
+
+        serviceConfig = {
+          Type = "oneshot";
+          RemainAfterExit = true;
+          # Root: read root-owned sops pin; create tokens as psibase.
```
swatanabe: This may or may not use sops. That's determined by the config.

nix/deploy/module.nix
```diff
@@ -0,0 +1,354 @@
...
+          ExecStart = pkgs.writeShellScript "psibase-softhsm-init" ''
+            set -euo pipefail
+            export SOFTHSM2_CONF=${lib.escapeShellArg softhsmConf}
+            mkdir -p ${lib.escapeShellArg cfg.softHsm.tokenDir}
+            chown psibase:psibase ${lib.escapeShellArg cfg.softHsm.tokenDir}
+
+            # psibase is nologin; script needs a real shell.
```
swatanabe: I don't follow 
- This service runs as `root`, not as `psibase`
- What is a "real shell"?

nix/deploy/module.nix
```diff
@@ -0,0 +1,354 @@
...
+            fi
+
+            pin=$(tr -d '\n' < ${lib.escapeShellArg cfg.softHsm.pinFile})
+            if [ -z "$pin" ]; then
+              echo "SoftHSM PIN file is empty: ${cfg.softHsm.pinFile}" >&2
+              exit 1
+            fi
+
+            # PIN via stdin+pty (not argv). Discard output: failures can leak PIN.
+            if ! printf '%s\n%s\n%s\n%s\n' "$pin" "$pin" "$pin" "$pin" \
```
swatanabe: This assumes that `printf` is a shell builtin. It is in `bash`, but it's also a separate program.

nix/deploy/module.nix
```diff
@@ -0,0 +1,354 @@
...
+          ExecStart = pkgs.writeShellScript "psibase-softhsm-init" ''
+            set -euo pipefail
+            export SOFTHSM2_CONF=${lib.escapeShellArg softhsmConf}
+            mkdir -p ${lib.escapeShellArg cfg.softHsm.tokenDir}
+            chown psibase:psibase ${lib.escapeShellArg cfg.softHsm.tokenDir}
+
+            # psibase is nologin; script needs a real shell.
+            as_psibase() {
+              runuser -u psibase -- \
+                env SOFTHSM2_CONF="$SOFTHSM2_CONF" PATH="$PATH" \
```
swatanabe: AFAICT, `runuser` and `env` both propagate (most of) the environment by default.

nix/deploy/module.nix
swatanabe: I think the proper way to manage the PIN file is to run the service as `psibase` and use `LoadCredential`.

Velua:
@swatanabe I've made changes

nix/deploy/module.nix
swatanabe: I don't think you quite understood the point of my comment. Why does this use stdin+pty instead of argv?

Velua: Oh right, I think it had a use prior but no longer exists, I've removed it

swatanabe: There is a legitimate reason to avoid passing secrets as command line arguments. It can leak the secrets to other users through `/proc/<pid>/cmdline`.

James-Mart:
This feels to me like it belongs in a separate repo more than `nix/dev`. Isn't this basically the nix version of `psibase-node-deployment`?

* It doesn't need to build from source
* It doesn't update when `main` on psibase updates, only when there are new published versions
* Importing `nixosModules.psibase` from this flake causes a node operator to fetch the contributor-shell's extra inputs. 

Any reason not to move this to e.g. `gofractally/psibase-deploy-nix` or similar?

Velua:
[PsiNix](https://github.com/velua/psinix) is my `psibase-node-deployment`, this PR is what felt like the right level of abstraction to me between the two.

Velua:
Closing in favour of [psibase-nix](https://github.com/gofractally/psibase-nix)

PR #1954, opened 2026-07-21
Add x-wrtcsig signaling service and Homepage query-token auth

cool-ant:
This is PR2 in a chain of PRs to release the webrtc work. Second phase is the node-local signaling service (`x-wrtcsig`), which handles
- Authenticated websockets: clients connect with `psibase.realtime.v1` + a bearer token (Homepage can obtain that token via Host).
- Presence: who among your contacts is online on this node (snapshot + deltas).
- Session signaling: join / signal / leave / invite fanout for Chat-authorized sessions (`chat-data`, `av-call`, pair), with welcome ICE defaulting to STUN.

## Architectural Decisions and their Impact
**`x-wrtcsig` is a subjective (node-local) service.**  
Presence, live WS joins, and signal routing live in subjective tables on the node. Architectural impact:

1. **Realtime stays off the consensus path** — fanout and ICE/signaling don’t bloat the chain or require global agreement on who’s “online right now.”
2. **Peers must share a node (or a path to the same signaling instance)** to see each other’s presence and exchange offers/ICE — the node can meter/bill that usage.
3. **Authorization still comes from objective Chat** — join is gated by on-chain `authorizeSessionJoin`, so the subjective socket can’t invent membership; it only brokers live sessions for members the chain already allows.

In short: **Chat (PR1) = objective who’s-allowed / session metadata infra; `x-wrtcsig` (this PR) = subjective who’s-connected-now and how they signal, node-local.**

## Summary
Adds the local XWebRtcSig package (`x-wrtcsig`): authenticated WS handshake, presence, join/signal/leave/invite, and welcome ICE (default STUN; TURN merge path exists, config UI later).
Relaxes Host `get-active-query-token` so Homepage (as the active app) can fetch a bearer for the signaling handshake.
Wires `XWebRtcSig` into CMake / `XDefault` installs, plus a minimal impl that returns `[]` as a list of TURN servers until TURN admin lands so welcome can still fall back to STUN.

**Added in this PR:** one service — `x-wrtcsig` (`XWebRtcSig`, `scope = local`). Host auth is a plugin tweak on existing Host, not a new service. (Chat / Spaces / sessions landed in PR1.)

### via `ctest` / manual `cargo-psibase test`
There is no Homepage UI in this PR. Validation is via cargo-psibase test (primary) and optional Python WS integration against a built node.

For convenience, here's the unit-test command (also `ctest rs-test-XWebRtcSig`):
```
build/rust/release/cargo-psibase test \
  --manifest-path packages/local/XWebRtcSig/service/Cargo.toml \
  --target-dir packages/local/target \
  --psitest=build/psitest
  ```
### Python integration (opt-in; not default ctest)
Meaningful end-to-end for this PR — real WS + Chat `authorizeSessionJoin`. Smoke: auth, presence, one join/signal. Full suite includes slow aiortc P2P (see `packages/local/XWebRtcSig/README.md` for the smoke subset).
```
PYTHONPATH=programs/psinode/tests \
PSIBASE_PACKAGE_DIR=build/share/psibase/packages \
python packages/local/XWebRtcSig/tests/python/test_xwebrtcsig.py \
  --psinode=build/psinode
```
## Expectations
WS without bearer → rejected.
Authed connect → welcome with STUN ICE.
Two users online → presence deltas.
`joinSession` / `signal` / `leaveSession` for chat-data and av-call → Expect: Chat authorizes members; non-members rejected.
Chat message bodies on the signaling socket → rejected.

### CODE REVIEW

File: packages/user/Host/plugin/auth/src/lib.rs [hunk 1]
```diff
@@ -70,9 +70,17 @@ impl Api for HostAuth {
...
             "get-active-query-token@host:auth/api",
         );
+        let caller = get_sender();
+        if caller == "homepage" {
+            assert!(
+                app == "homepage",
+                "[get-active-query-token@host:auth/api] Unauthorized caller: {}",
+                caller
+            );
+        }
```
cool-ant, line 83: @James-Mart I'm pretty sure we discussed this needing some better approach. This is the minimal/most specific implementation I could come up with for now. Happy to discuss future-oriented, better options when you're ready.

packages/local/XWebRtcSig/service/src/trace.rs
```diff
@@ -0,0 +1,14 @@
...
+//! Enable with `cargo build --features rt-trace` (off by default in production builds).
+
+macro_rules! xrtcsig_trace {
+    ($($arg:tt)*) => {{
+        #[cfg(feature = "rt-trace")]
+        {
+            psibase::write_console(&format!($($arg)*));
+        }
+    }};
+}
```
cool-ant: @swatanabe @James-Mart I know this is non-standard in the codebase. Do we have any rule for what gets logged (in a Debug build)? I don't know if there's any precedent for adding infra-type things to the trace or not. Happy to strip it all out if not. Also happy to simply thin out what gets traced if you think some minimal form of tracing might be useful.

swatanabe: If it's something that an infrastructure provider might want to see, use `x-http::log` instead of writing to the console.

cool-ant: ok, I reviewed all call sites and opted to just remove the whole mechanims.

PR #1958, opened 2026-07-21
Adding missing check in Accounts for "x-"-prefixed name creation

### LINKED ISSUE #1956: Users can buy `x-*` names via NameMarket (and shouldn't be able to!)
My work creating a package for reserving names (https://github.com/gofractally/psibase-pm/issues/44) surfaced this as a potential oversight. Fable flagged this, and when I asked it to verify that it is indeed a bug, it hit the safety filter, degraded to Opus, and claims to have verified it. Here's what Fable said:

NameMarket's buy action preapproves arbitrary names with no x- prefix check of its own — it only verifies that an auction exists for that name length and is enabled:

```rs
// /root/psibase/packages/user/NameMarket/service/src/lib.rs
let auction = AuctionsTable::new()
    .get_index_pk()
    .get(&length)
    .expect("auction not found for this length");

assert!(auction.enabled, "market is disabled");

// Alert the DiffAdjust service to the purchase
let current_price = Quantity::from(DiffAdjust::call().increment(auction.nft_id, 1));

Accounts::Wrapper::call().preapproveAcc(account);
AuthDelegate::Wrapper::call().newAccount(account, get_service(), true);
```

So if auctions are enabled for, say, 7-character names, it looks like anyone could buy x-sites through the market. The same applies to MIN_ALLOWED_ACCOUNT_LENGTH (8): short generics like root or auth can't be created through the normal path at all — the name market is precisely how they'd be obtained, which is also why your ReservedNames package matters even for names the plain-creation rules already block.

cool-ant:
Fixes #1956

### DIFF
--- a/packages/system/Accounts/include/services/system/Accounts.hpp
+++ b/packages/system/Accounts/include/services/system/Accounts.hpp
@@ -64,7 +64,8 @@ namespace SystemService
       /// Preapprove an account of any length to be created within this transaction.
       /// If an account is preapproved, it will bypass naming restrictions.
       ///
-      /// The `Accounts` service must itself authorize this action.
+      /// The `Accounts` or `NameMarket` service must itself authorize this action.
+      /// Names with an "x-" prefix must be preapproved by the `Accounts` service specifically.
       ///
       /// The preapproval only lasts for the duration of the transaction context, after
       /// which the preapproval is cleared.
--- a/packages/system/Accounts/include/services/system/Accounts.hpp
+++ b/packages/system/Accounts/include/services/system/Accounts.hpp
@@ -64,7 +64,7 @@ namespace SystemService
       /// Preapprove an account of any length to be created within this transaction.
       /// If an account is preapproved, it will bypass naming restrictions.
       ///
-      /// The `Accounts` or `NameMarket` service must itself authorize this action.
+      /// The `Accounts` or `NameMarket` service must authorize this action.
       /// Names with an "x-" prefix must be preapproved by the `Accounts` service specifically.
       ///
       /// The preapproval only lasts for the duration of the transaction context, after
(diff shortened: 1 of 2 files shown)


packages/system/Accounts/src/Accounts.cpp
```diff
@@ -64,6 +64,11 @@ namespace SystemService
       {
          check(getSender() == getReceiver() || getSender() == AccountNumber("namemarket"),
                "unauthorized");
+         if (name.str().starts_with("x-"))
+         {
+            check(getSender() == getReceiver(),
+                  "The 'x-' account prefix is reserved for infrastructure providers");
+         }
       }
```
James-Mart, line 72: As written, sender is checked if it is receiver twice for `x-` names.
I think this would read more clearly if the `x-` check was top level
```
if (name.subaccount) -> specific rule (must be base)
else if (name starts with "x-") -> specific rule (must be receiver)
else -> generic (can be receiver or namemarket)
```

packages/system/Accounts/include/services/system/Accounts.hpp
```diff
@@ -64,7 +64,8 @@ namespace SystemService
       /// Preapprove an account of any length to be created within this transaction.
       /// If an account is preapproved, it will bypass naming restrictions.
       ///
-      /// The `Accounts` service must itself authorize this action.
+      /// The `Accounts` or `NameMarket` service must itself authorize this action.
```
swatanabe: The reflexive pronoun only makes sense for `Accounts`, not `NameMarket`

swatanabe (approved):

packages/system/Accounts/include/services/system/Accounts.hpp
cool-ant: true. Thanks for giving myself that feedback. :wink:

brandonfancher (approved):

PR #1960, opened 2026-07-21
Add reserved-names package to reserve confusable account names

brandonfancher:
https://github.com/gofractally/psibase-pm/issues/44

## Summary

Adds a new CLI tool, `reserved-names` (in `rust/reserved-names`), that generates a meta-only psibase package reserving account names which could otherwise be used for phishing or impersonation. This protects two categories of names:

1. **Confusable lookalikes of service names.** For every account defined by the given `.psi` packages, the tool generates all homoglyph variants by swapping the easily-confused spellings `{i, l, 1}`, `{o, 0}`, `{rn, m}`, and `{vv, w}` (e.g. `s1tes`, `t0kens`, `adrnin`, `vvallet`), so a malicious user can't register a lookalike of a real app and serve it from a convincing subdomain.
2. **Generic names.** A curated list in `rust/reserved-names/generic-names.txt` (`root`, `admin`, `auth`, `login`, `wallet`, ...) that users could abuse to appear official. Since most of these are shorter than the 8-character minimum for ordinary account creation, they would otherwise be obtainable through the premium name market.

The output is a `.psi` package (a zip containing only `meta.json`) listing all reserved names as non-service accounts, with dependencies on `Accounts` and `AuthDelegate`. Installing it creates each account owned by the installer via `auth-delegate`, making them unclaimable by ordinary users. Because account creation during package installation accepts existing accounts owned by the installer, reserving the real service names does **not** block installing those packages later — and individual names can be released with `setOwner` if needed.

## Usage

```bash
# Write the package
reserved-names --names generic-names.txt --version 0.23.0 \
    -o ReservedNames.psi path/to/packages/*.psi

# Or just list the reserved names on stdout
reserved-names --names generic-names.txt path/to/packages/*.psi
```

## Implementation notes

- Parses input packages with `psibase::PackagedService` and collects accounts via `get_accounts()`.
- Subaccounts (`name+N`) are reduced to their base name; reserving the base is sufficient since subaccount creation requires the base account's approval.
- `x-*` names are excluded: the `x-` prefix is already reserved for infrastructure providers by the Accounts service, and confusable variants of an `x-` name still start with `x-`.
- Variants that exceed the 10-character account name limit (from multi-character substitutions) are dropped; those that don't round-trip through the `AccountNumber` compression are skipped with a warning.

## Next steps

- @swatanabe to integrate it into CMake


brandonfancher:
@James-Mart A couple questions:

1. We originally discussed generating the package with `cargo-psibase`, but apparently it has no way to declare non-service accounts and it crashes on crates without a psibase dependency (which I guess means it doesn't support meta-only packages). Is it okay that I went the `zip` crate route instead?
2. This is currently configured to substitute `{i, l, 1}` and `{o, 0}`. We can do any combination of `{i, l, 1}`, `{o, 0}` and even multi-character confusables (`rn`/`m`, `vv`/`w`). Thoughts? I don't want to go overboard.

James-Mart:
>  Is it okay that I went the zip crate route instead?

Yeah that works

> This is currently configured to substitute {i, l, 1} and {o, 0}. We can do any combination of {i, l, 1}, {o, 0} and even multi-character confusables (rn/m, vv/w). Thoughts? I don't want to go overboard.

That sounds good! On multi-character just be careful it doesn't push the name over the char limit

James-Mart:
CICD failure: Uncommitted lock file `rust/Cargo.lock`

rust/reserved-names/Cargo.toml
```diff
@@ -0,0 +1,18 @@
...
+publish.workspace = true
+description = "Generates a psibase package that reserves confusable account names"
+license.workspace = true
+
+[dependencies]
+anyhow = { workspace = true }
+clap = { workspace = true }
+psibase = { workspace = true }
+serde_json = { workspace = true }
+zip = { version = "7.2", default-features = false, features = ["deflate"] }
```
cool-ant: It'd be nice to centralize this version number in the parent Cargo.toml, as so many others already do (if the feature flags match)

cool-ant:
Thoughts:
- I think this should be look-alike names only. That avoids the issue we saw where this package failed to install. Any *actual* name should be separately reserved via some other mechanims.
- Any reason not to make this more deeply integrated rather than so opt-in?
- perhaps we include mistypes as well? admim, afmin, etc.?
- I think we should consider other confusables, e.g.., cl/d, nn/m, u/v

rust/reserved-names/src/main.rs
```diff
@@ -0,0 +1,361 @@
...
+    } else {
+        for name in &generated.reserved {
+            println!("{}", name);
+        }
+    }
+    Ok(())
+}
+
+#[cfg(test)]
+mod tests {
```
cool-ant: I tend to not like files longer than ~250 lines. Tests would be a great place to split this file without having to think through how to logically extract out some other set of functions.

brandonfancher: Good call.

brandonfancher:
> * I think this should be look-alike names only. That avoids the issue we saw where this package failed to install. Any _actual_ name should be separately reserved via some other mechanims.

We definitely want to reserve _actual_ names in this body of work. In fact, look-alike names were a secondary objective...almost an afterthought...a bonus.

> * Any reason not to make this more deeply integrated rather than so opt-in?

This isn't core and is quite opinionated. Certain types of networks may not have the need for this.

> * perhaps we include mistypes as well? admim, afmin, etc.?

Why?

> * I think we should consider other confusables, e.g.., cl/d, nn/m, u/v

Yeah, we could. But I'm ambivalent beyond what we have. The multi-character confusables already feel a bit arbitrary to me.

@James-Mart feel free to offer your thoughts on these as well.

rust/reserved-names/src/main.rs
```diff
@@ -0,0 +1,259 @@
...
+        // occurrences of "vv" in "vvv").
+        let mut literal = prefix.to_string();
+        literal.push(first);
+        expand(&rest[first.len_utf8()..], &literal, results);
+        // Replace any confusable spelling that starts here with each other
+        // member of its class.
+        for class in CONFUSABLE_CLASSES {
+            for spelling in *class {
+                if let Some(tail) = rest.strip_prefix(spelling) {
+                    for replacement in *class {
```
swatanabe, line 83: This should skip expanding `replacement == spelling`. That case is covered by adding the next character. This is an EXPTIME algorithm, so it's important to reduce the number of recursive calls as much as possible.

rust/reserved-names/src/main.rs
```diff
@@ -0,0 +1,259 @@
...
+        }
+    }
+
+    if let Some(names) = &args.names {
+        seeds.extend(read_names_file(names)?);
+    }
+
+    let generated = generate(&seeds);
+
+    if !generated.skipped.is_empty() {
```
swatanabe: Why bother issuing a warning? The point of this package is to prevent users from creating those accounts. If it isn't a valid `AccountNumber`, they can't create it. In addition, the only way to "fix" the warning is to remove the base name, but that's wrong if the base name or some of its variants *are* valid.

The only case that I think would be worth warning about is if none of the base name or any of its variants are valid.

rust/reserved-names/src/main.rs
```diff
@@ -0,0 +1,258 @@
...
+fn generate(seeds: &BTreeSet<String>) -> Generated {
+    let mut reserved = BTreeSet::new();
+    let mut unused_seeds = Vec::new();
+    for seed in seeds {
+        let mut any_valid = false;
+        for variant in confusable_variants(seed) {
+            // Substitutions like m -> rn can push a variant past the
+            // maximum name length; such variants simply don't exist as
+            // account names, so drop them.
+            if variant.len() > MAX_ACCOUNT_NAME_LENGTH as usize {
```
swatanabe: This is redundant now

rust/reserved-names/Cargo.toml
brandonfancher: Addressed. Thanks!

rust/reserved-names/src/main.rs
brandonfancher: Thanks!

rust/reserved-names/src/main.rs
```diff
@@ -0,0 +1,252 @@
...
+            .collect(),
+        accounts: reserved.iter().map(|name| name.parse().unwrap()).collect(),
+        services: vec![],
+        exports: vec![],
+    };
+
+    let file =
+        File::create(path).with_context(|| format!("Failed to create {}", path.display()))?;
+    let mut out = ZipWriter::new(file);
+    out.start_file("meta.json", SimpleFileOptions::default())?;
```
swatanabe: `cargo-psibase` uses `SimpleFileOption::DEFAULT` to make the output closer to deterministic. For this tool, using `DEFAULT` should be sufficient to make the archive deterministic.

rust/reserved-names/src/tests.rs
```diff
@@ -0,0 +1,116 @@
...
+        assert!(name.len() <= MAX_ACCOUNT_NAME_LENGTH as usize);
+        assert!(AccountNumber::from_exact(name).is_ok());
+    }
+    assert!(generated.unused_seeds.is_empty());
+}
+
+#[test]
+fn warns_only_when_no_variant_is_valid() {
+    // Individual invalid variants of a good seed are dropped silently.
+    let seeds = set(&["billing"]);
```
swatanabe: How does this test dropping invalid variants? `billing` should have 81 variants, all of which are valid, right? With the new name algorithm, the only way to get an invalid name from a valid one is to increase the length, or if the substitutions include invalid characters or `-` (which none do).

rust/reserved-names/src/main.rs
brandonfancher: addressed, thanks!

rust/reserved-names/src/tests.rs
brandonfancher: addressed, thanks!

swatanabe (approved):

PR #1962, opened 2026-07-22
Pack nested actions in postinstall scripts

swatanabe:
This avoids the need to manually pack a hex `rawData` field for `runAs`.


rust/psibase/src/package.rs
```diff
@@ -354,6 +360,240 @@ impl PrettyAction {
...
+        val: &serde_json::Value,
+        dest: &mut Vec<u8>,
+    ) -> Result<(), serde_json::Error> {
+        let (sender, service, method, _raw_data_name) = deserialize_pretty_action(ty, val)?;
+
+        (self.f)(sender, service);
+        if let Some(data) = val.get("data") {
+            if let Some(schema) = self.schemas.get(&service) {
+                if let Some(action_type) = schema.actions.get(&MethodString(method.to_string())) {
+                    let mut cschema =
```
cool-ant, line 568: `CustomPrettyActionCollectAccounts` silently skips missing nested schemas/actions; is it worth detecting and doing a hard failure/error as at the top-level?

rust/psibase/src/schema.rs
```diff
@@ -385,6 +385,171 @@ pub fn schema_types() -> CustomTypes<'static> {
...
+                }
+            } else {
+                format!("missing schema for {service}")
+            }
+        } else {
+            format!("missing field data")
+        };
+        let raw_data = if let Some(raw_data) = raw_data {
+            raw_data
+        } else {
```
cool-ant, line 510: If I'm reading this all right, Rust is checking data first and falling back to rawData, which is meaningfully different than the C++ impl, which is: prefer non-empty rawData; only pack data if rawData is empty.
This impl would also miss real errors if 1) there's an issue with `data` and 1) `rawData` is present but not meant to be the canonical data (was replaced with `data` without `rawData` being removed).

rust/psibase/src/main.rs
cool-ant, line 2066: `psibase install --local` will fail for packages with nested/data postinstalls.
local should use HttpShemaFetcher, same as non-local, right?

libraries/psibase/common/src/package.cpp
cool-ant, line 366: C++/Rust delta: Rust vs C++ install order can diverge for nested cross-package Actions. Only top-level postinstall services are collected and nested Actions (unlike Rust) are not walked for install ordering

packages/local/XAdmin/ui/wasm/src/lib.rs
```diff
@@ -51,7 +51,11 @@ impl Boot for XAdminPlugin {
                 PackagedService::new(Cursor::new(s)).map_err(|e| e.to_string())?,
             ));
         }
-        sort_package_ops(&mut ops, &PackageList::new()).map_err(|e| e.to_string())?;
+        let mut schemas = SchemaMap::new();
+        for op in &mut ops {
+            op.new().unwrap().get_all_schemas(&mut schemas);
```
cool-ant, line 59: Result being discarded here; potential to swallow errors

libraries/psibase/common/src/package.cpp
```diff
@@ -98,6 +98,133 @@ namespace psibase
...
+               }
+               else if (name == "rawData")
+               {
+                  act.rawData = bytesFromJson(value);
+               }
+               else if (name == "data")
+               {
+                  data = &value;
+               }
+            }
```
cool-ant: C++/Rust delta: This impl
- Hardcodes "sender" / "service" / "method" / "rawData" / "data"; Rust uses schema field names.
- Omitting fields default to 0 where Rust requires them.
- No validation that method was present before schema lookup.

packages/user/Events/src/postinstall.json
```diff
@@ -25,7 +25,7 @@
         "sender": "",
         "service": "events+2",
         "method": "onBlock",
-        "rawData": "0000"
+        "data": {}
       }
     }
   },
```
cool-ant, line 31: Not sure if you're intention is to update all of these (would be nice), e.g., Transact still uses rawData

packages/user/Packages/plugin/src/lib.rs
```diff
@@ -285,27 +286,15 @@ fn get_installed_manifest(
...
-            );
-        }
+struct PluginSchemaFetcher;
+
+#[async_trait(?Send)]
+impl SchemaFetcher for PluginSchemaFetcher {
+    async fn fetch_schema(&self, service: AccountNumber) -> Result<Schema, anyhow::Error> {
+        Ok(serde_json::from_str(&Server::get_json(&format!(
+            "/schema?service={service}"
+        ))?)?)
```
cool-ant, line 297: It looks like `get_required_schemas` might be dead code now

rust/psibase/src/package.rs
```diff
@@ -877,30 +1110,14 @@ impl<R: Read + Seek> PackagedService<R> {
...
-    pub async fn load_schemas(
-        &mut self,
-        base_url: &reqwest::Url,
-        client: &mut reqwest::Client,
+    pub async fn load_schemas<F: SchemaFetcher>(
+        &self,
+        fetcher: &F,
         schemas: &mut SchemaMap,
     ) -> Result<(), anyhow::Error> {
         self.get_all_schemas(schemas)?;
```
cool-ant, line 1107: Should `get_all_schemas` be called here (as well as in `fetch_all_schemas`)?

packages/user/Events/src/postinstall.json
swatanabe: Transact has to use `rawData` in two of the three actions, because it must be empty (the arguments are provided by native).

rust/psibase/src/main.rs
swatanabe: Local packages do not support postinstall. They also can't use HttpSchemaFetcher, because the schema isn't stored anywhere, and if it were, it would be stored in `x-packages`, not `packages`.

rust/psibase/src/schema.rs
swatanabe: Having both `data` and `rawData` isn't really an intended use case. The only reason I added the check for empty `rawData` in C++ is to prevent the generation of invalid packing by appending packed `data` to `rawData`.

packages/local/XAdmin/ui/wasm/src/lib.rs
swatanabe: Actually the `Result` is spurious. `get_all_schemas` always succeeds.

rust/psibase/src/package.rs
swatanabe: I don't think so. The error will show up when trying to pack the action.

rust/psibase/src/package.rs
swatanabe: Yes. The duplication is harmless and `load_schemas` isn't usable by itself without it.

cool-ant (approved):

PR #1964, opened 2026-07-22
Fix null dereference

swatanabe:
This happens when processing a block containing a transaction that isn't in the queue (e.g. syncing the chain). Aborting `onTrx` is mostly harmless in this case, because there can't be any clients waiting on a response for the transaction.


packages/system/Transact/src/RTransact.cpp
```diff
@@ -708,7 +708,10 @@ void RTransact::onTrx(const Checksum256& id, psio::view<const TransactionTrace>
       std::uint32_t txSize;
```
James-Mart: Does this need to be `std::uint32_t txSize{};`?

James-Mart (approved):

PR #1969, opened 2026-07-24
add {{root}} keyword to CSP 

### LINKED ISSUE #1936: User-defined CSP policies need a way to refer to the active root domain
As a prerequisite to https://github.com/gofractally/psibase/issues/977

User-generated CSP strings should be able to use a special keyword in their CSP string that gets automatically replaced with the rootdomain of the request, just like the behavior of the default CSP strings.

Otherwise, a user who needs to loosen the csp restrictions will have to sacrifice all subdomain-specific security clamping.

Velua:
Adds a `{{root}}` keyword to CSP strings in `sites`.

Any `{{root}}` in a CSP (default or user-defined via `setCsp`) is replaced with the root domain of the request, including the port when present:

```
connect-src 'self' {{root}} *.{{root}};
```

becomes, on a local node:

```
connect-src 'self' psibase.localhost:8080 *.psibase.localhost:8080;
```

This lets apps write subdomain-scoped policies without hardcoding a domain or falling back to `*`.


doc/src/default-apps/sites.md
```diff
@@ -28,6 +28,26 @@ If the baseline CSP is insufficient, the `sites` service allows users to customi
 
 The priority of the CSP header applied to content is given to a header set for a specific path, then the header set for an entire site, and finally will default to the default policy if no custom headers were provided.
 
+Note that a custom CSP **completely replaces** the baseline for the content it covers — the policies are not merged — so a custom CSP should be stated as a complete policy.
+
+#### The `{root}` keyword
```
swatanabe: `{{root}}` would make it closer to a typical template language, and therefore easier to generalize.

packages/user/Sites/include/services/user/Sites.hpp
```diff
@@ -130,6 +131,12 @@ namespace SystemService
       /// Sets the Content Security Policy for the specified path (or "*" for a global CSP).
       /// If a specific CSP is set, it takes precedence over the global CSP.
       /// If no specific or global CSP is set, a default CSP is used.
+      ///
+      /// The CSP string may include the keyword `{root}`, which is replaced at
```
swatanabe: I think this should have a boolean argument indicating whether to expand keywords, to make it possible to set the exact text, and more importantly to make it less likely that someone will be surprised by unexpected expansion.

Velua: Adding a param to this signature would cascade in several areas and require storage on-chain won't it? Given `{` and `}` isn't valid CSP grammar anyway, is there really a point? What risk are we trying to guard against? If we've unexpected expansion concerns can we just make the token more distinctive?

James-Mart: > Given { and } isn't valid CSP grammar

Technically I think it *is* valid CSP grammar - The `directive-value` [allows curly braces](https://w3c.github.io/webappsec-csp/#framework-directives).

However to my eye none of the actual user-defined directive values (source-expressions, `report-uri`, `report-to`, `trusted-types`) actually allow curly braces.

For that reason I'm currently leaning towards considering the presence of double curlies `{{ X }}` as the opt-in, with no separate boolean. @swatanabe are you okay with that? It means basically the only users who could be surprised by unexpected expansion would be those who set invalid CSP directive values.

doc/src/default-apps/sites.md
```diff
@@ -28,6 +28,26 @@ If the baseline CSP is insufficient, the `sites` service allows users to customi
...
+default-src 'self'; connect-src 'self' {{root}} *.{{root}}; frame-src supervisor.{{root}};
+```
+
+becomes, on a local node:
+
+```
+default-src 'self'; connect-src 'self' psibase.localhost:8080 *.psibase.localhost:8080; frame-src supervisor.psibase.localhost:8080;
+```
+
+This lets apps loosen or tighten CSP for sibling subdomains without hardcoding a specific deployment domain (and without falling back to `*`, which would discard subdomain-specific clamping).
```
James-Mart: Docs and doc comments refer to this expanded value as the "deployment domain", "deployment root host", and "deployment root domain"

1. I think we should be consistent
2. I think coupling the value to the concept of a "deployment" makes sense in a single-node psibase network, but makes less sense on networks with multiple infra providers and therefore multiple "deployment" domains. I think I recommend just dropping the word "deployment", and just use "root domain" everywhere.

packages/user/Sites/include/services/user/Sites.hpp
swatanabe: The other concern is whether if parts of the CSP are from an untrusted source, expansion can lead to any kind of injection attack. However, `{{root}}` cannot cause an information leak and anything that an attacker could do with it, he could do by inserting whatever string he wants directly. Correct validation should prevent both equally. Therefore I think `{{root}}` is safe, but any additional keywords will need to be evaluated for safety.

James-Mart: Understood. And worth noting because I do think additional keywords will be useful in the future.

packages/user/Sites/src/Sites.cpp
```diff
@@ -92,6 +97,62 @@ namespace SystemService
           "connect-src * blob:;"                                           //
           ;
 
+      // Port suffix from the Host header (e.g. ":8080"), empty if absent or not
+      // a valid port. request.host is stored without the port, so the raw
+      // Host header is the source of truth.
+      //
+      // The Host header is client-controlled and this value is interpolated into
```
swatanabe: `Host` is a forbidden header, so it requires a malicious user agent to set it to something invalid. Since the CSP header is interpreted by the user agent in the first place, I don't see what such an attack could gain. Also, because `Host` is used to dispatch requests, and invalid host like below won't even make it to `Sites`.

packages/user/Sites/src/Sites.cpp
```diff
@@ -92,6 +97,62 @@ namespace SystemService
...
+            std::string_view h = *host;
+            if (auto pos = h.rfind(':'); pos != std::string_view::npos)
+            {
+               // A ']' after the last ':' means that colon is inside a
+               // bracketed IPv6 literal, so there is no port.
+               if (h.find(']', pos) != std::string_view::npos)
+                  return {};
+
+               std::string_view digits = h.substr(pos + 1);
+               if (digits.empty() || digits.size() > 5 ||
```
swatanabe: Use `std::from_chars` with `std::uint16_t`.

doc/src/default-apps/sites.md
```diff
@@ -28,6 +28,26 @@ If the baseline CSP is insufficient, the `sites` service allows users to customi
 
 The priority of the CSP header applied to content is given to a header set for a specific path, then the header set for an entire site, and finally will default to the default policy if no custom headers were provided.
 
+Note that a custom CSP **completely replaces** the baseline for the content it covers — the policies are not merged — so a custom CSP should be stated as a complete policy.
```
swatanabe: "be stated" doesn't add anything

doc/src/default-apps/sites.md
```diff
@@ -28,6 +28,26 @@ If the baseline CSP is insufficient, the `sites` service allows users to customi
...
+
+#### The `{{root}}` keyword
+
+User-defined CSP strings may include the keyword `{{root}}`. At serve time, any `{{root}}` in the CSP that will be sent is replaced with the root domain of the request, including the port when present (e.g. `psibase.localhost:8080` in local development, or `example.com` in production).
```
swatanabe: "serve time" sounds a bit awkward to me.

packages/user/Sites/src/Sites.cpp
James-Mart: In which case there's already a simpler version of this `hostHeaderPortSuffix` function in `HttpServer`.
Consider hoisting into `HttpHeaders.hpp/cpp` and calling it from both `Sites` and `HttpServer`

James-Mart (approved):

Issue #1939, opened 2026-07-10
User avatars should be served from `profiles` subdomain


brandonfancher:
As a follow-up to https://github.com/gofractally/psibase/issues/977

User avatars are currently served from the target account's subdomain--which could be any valid account name and which can also represent third-party/untrusted apps. This forces apps that want to display user account avatars to allow any subdomain `img-src` in their CSP, which can lead to exfiltration methods if XSS is achieved. (See further explanation in #1935.)

Instead, we should serve (proxy?) these from the `profiles` app subdomain. In that way, app CSPs can trust `profiles` (a system app) instead of arbitrary third parties.

PR #1972, opened 2026-07-30
Serve user avatars from Profiles subdomain

cool-ant:
Issue #1939 

Account avatars now served from profiles.{root}/avatar/{account} via proxying to Sites (URL: /profile/avatar.image), so app CSPs can trust profiles instead of *.{root}.
Clients: point shared useAvatar at that URL; keep uploads on the user’s site.
I restricted uploads to image/png|jpeg|webp|gif, and I used the fake `.image` extension so it would be a dotted storage path and so SPA sites still resolve the file.


packages/user/Profiles/query-service/src/lib.rs
```diff
@@ -14,12 +18,82 @@ mod service {
...
+            Hex(Sites::call().getData(account, AVATAR_PATH.into(), true))
+        };
+
+        let mut headers = allow_cors_with_origin("*");
+        headers.push(HttpHeader::new("Cache-Control", "no-cache"));
+        headers.push(HttpHeader::new("ETag", &etag));
+
+        Some(HttpReply {
+            status: HttpStatus::Ok as u16,
+            contentType: props.contentType,
```
swatanabe: Is this vulnerable to XSS is someone creates an "avatar" that isn't an image?

cool-ant: 3 changes to tighten this up:
- don't allow SVG on upload (not directly related to your commend but security-relevant)
- restrict types to the same allowlist as on upload (images only)
- add header to tell browser never to treat this as any other content type than the one being returned.

These 3 should cover your comment and a few other semi-messy, potentially exploitable items.

packages/user/Profiles/query-service/src/lib.rs
```diff
@@ -3,8 +3,13 @@
 mod service {
     use async_graphql::*;
     use profiles::tables::Profile;
+    use psibase::services::sites::Wrapper as Sites;
     use psibase::*;
 
+    const AVATAR_PATH: &str = "/profile/avatar.image";
```
swatanabe: This fake extension seems a bit hacky. Here are a few other ideas:

- Call it `.avatar`
- Change Sites to know that `/profile/avatar` is not an SPA path
- Change Sites to have a standard prefix for non-SPA paths
- Check for multiple extensions: avatar.png, avatar.jpg, etc, instead of a single path

packages/user/Profiles/query-service/src/lib.rs
```diff
@@ -14,12 +19,106 @@ mod service {
...
+
+    fn normalize_content_type(content_type: &str) -> &str {
+        content_type
+            .split(';')
+            .next()
+            .unwrap_or(content_type)
+            .trim()
+    }
+
+    fn is_allowed_content_type(content_type: &str) -> bool {
```
swatanabe: Can we avoid duplicating this here and in the plugin?

packages/user/Profiles/query-service/src/lib.rs
```diff
@@ -14,12 +19,106 @@ mod service {
...
+        let Some(props) = Sites::call().getProps(account, AVATAR_PATH.into()) else {
+            return Some(not_found());
+        };
+
+        if !is_allowed_content_type(&props.contentType) {
+            return Some(not_found());
+        }
+
+        let content_type = normalize_content_type(&props.contentType).to_ascii_lowercase();
+        let etag = etag_for(&props.contentHash);
```
swatanabe: Is there a reason for not using `Sites::serveSys`, which already handles caching?

packages/user/Profiles/query-service/src/lib.rs
```diff
@@ -14,12 +19,106 @@ mod service {
...
+            Hex(Vec::new())
+        } else {
+            Hex(Sites::call().getData(account, AVATAR_PATH.into(), true))
+        };
+
+        Some(HttpReply {
+            status: HttpStatus::Ok as u16,
+            contentType: content_type,
+            body,
+            headers: avatar_headers([
```
swatanabe: Missing `Content-Encoding`. We don't use `Content-Encoding` for images, but that's entirely on the client-side. `Sites` allows it.

brandonfancher (approved):
TS LGTM

packages/user/Profiles/query-service/src/lib.rs
cool-ant: I went with this approach because the other felt more hacky / too heavy.
Changing Sites: we could certainly establish a non-SPA path, but it felt like one-off code that we don't really have many examples of yet, so it felt like premature optimization. Not to mention that no matter what we choose for the path, I figure *someone* will eventually try to put a file there and may not understand why things aren't working as expected (why their avatar suddenly disappear or changed or why that particular path won't fetch from the server).
As for multiple extensions, that's possible, but it also isn't generalizable, i.e., if an app wants to grab the users avatar, what path do they find it at?
The current approach establishes a known path, restricts contents on upload and verifies on download, and--as required by the issue--improves the CSP situation.

packages/user/Profiles/query-service/src/lib.rs
cool-ant: No, good call

packages/user/Profiles/query-service/src/lib.rs
James-Mart: Haven't done a code review here to get the full context. I just read the comment threads. 

But for the record this idea

> Change Sites to have a standard prefix for non-SPA paths

Has independently occurred to me in the past. Might be a good signal.

swatanabe: `avatar.avatar` seems worse to me. It wasn't any of my suggestions.

cool-ant: I miss ready your "call it `.avatar`." comment. all moot now.

packages/user/Profiles/service/src/lib.rs
```diff
@@ -41,4 +95,36 @@ pub mod service {
...
+            image.len() <= MAX_AVATAR_SIZE,
+            "Avatar exceeds max file size of {MAX_AVATAR_SIZE} bytes"
+        );
+        assert!(
+            content_type_mime(contentType).is_some(),
+            "Avatar content type not allowed"
+        );
+
+        let caller = get_sender();
+        AvatarTable::new()
```
James-Mart: You should be calling `Sites::storeSys` instead of storing it in your own table

packages/user/Profiles/query-service/src/lib.rs
```diff
@@ -14,12 +16,67 @@ mod service {
...
+        HttpReply {
+            status: HttpStatus::NotFound as u16,
+            contentType: "text/html".into(),
+            body: Hex(b"Not Found".to_vec()),
+            headers,
+        }
+    }
+
+    /// Serves `GET|HEAD /avatar/<account>` from the Profiles avatar table.
+    fn serve_avatar(request: &HttpRequest) -> Option<HttpReply> {
```
James-Mart: Not needed

packages/user/Profiles/plugin/src/lib.rs
```diff
@@ -108,25 +107,33 @@ impl Api for ProfilesPlugin {
...
-        let file = File {
-            content_type: avatar.content_type,
-            content: avatar.content,
-            path: "/profile/avatar".to_string(),
+        let Some(content_type) = profiles::parse_content_type(&avatar.content_type) else {
+            return Err(ErrorType::InvalidAvatarContentType(avatar.content_type).into());
         };
 
-        bindings::sites::plugin::api::upload(&file, 11)
+        let packed = profiles::action_structs::uploadAvatar {
```
James-Mart, line 118: Sorry, one more thing I noticed: As written we do lose compression that sites plugin does automatically.

Not too hard to add it back in because `compress_content` is exported from `psibase`, and this plugin already depends on `psibase`.

Main thing is we need to thread `content_encoding` through `uploadAvatar` so it can be passed to `storeSys`

swatanabe: The only image type that we actually compress is svg, though. Other image formats have their own internal compression.

cool-ant: I added additional image formats. does that address what you're pointing at? Or are you saying I *shouldn't* add our brotli compression to uploaded images?

James-Mart (approved):

packages/user/Profiles/plugin/src/lib.rs
James-Mart: Approved with compression reverted

PR #1987, opened 2026-08-07
Allow credentials for x-admin accessing x-peers

swatanabe:
- CORS preflight never includes credentials, so we need to handle `OPTIONS` before checking authorization.
- Pass `credentials: include` to fetch
- Add `Access-Control-Allow-Credentials` and don't use `*` in `Access-Control-Allow-Headers`


James-Mart:
This still suffers from 
```
Access to fetch at 'http://x-peers.psibase.localhost:8080/graphql' from origin 'http://x-admin.psibase.localhost:8080' has been blocked by CORS policy: The value of the 'Access-Control-Allow-Origin' header in the response must not be the wildcard '*' when the request's credentials mode is 'include'.
```
Because `x-peers` `serveGraphQL` still uses `Access-Control-Allow-Origin: *` which isn't allowed with credentialed requests

James-Mart (approved):

PR #1993, opened 2026-08-11
XAdmin: Introduce UI for managing allowed peers whitelist

brandonfancher:
## In this PR

Adds UI to the Peers page allowing node operators to manage the incoming peers whitelist. This hooks into existing functionality in `x-peers`. This UI will be further refined when we do an overall ref(actor|inement) of the XAdmin UI soon.

## Follow-up work

- Validate the `x-peers` is doing what it needs to do with these account names

## Future work for XAdmin

- Transition `react-hook-form` forms to `@tanstack/react-form`
- Clean up and reorganize XAdmin UI; allow pre-boot/peered access to all relevant XAdmin settings

## Screenshots

image

image

image

image

image

### DIFF
--- a/packages/local/XAdmin/ui/src/peers/peers-page.tsx
+++ b/packages/local/XAdmin/ui/src/peers/peers-page.tsx
@@ -1,6 +1,5 @@
 import {
     Clipboard,
-    Info,
     MoreHorizontal,
     Plus,
     Trash,
@@ -26,7 +25,6 @@ import {
 import { EmptyBlock } from "@shared/components/empty-block";
 import { zAccount } from "@shared/lib/schemas/account";
 import { cn } from "@shared/lib/utils";
-import { Alert, AlertDescription, AlertTitle } from "@shared/shadcn/ui/alert";
 import { Button } from "@shared/shadcn/ui/button";
 import {
     Dialog,
@@ -46,7 +44,7 @@ import {
 } from "@shared/shadcn/ui/dropdown-menu";
 import { Input } from "@shared/shadcn/ui/input";
 import { Label } from "@shared/shadcn/ui/label";
-import { Switch } from "@shared/shadcn/ui/switch";
+import { RadioGroup, RadioGroupItem } from "@shared/shadcn/ui/radio-group";
 import {
     Table,
     TableBody,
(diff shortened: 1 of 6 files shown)


packages/local/XAdmin/ui/src/peers/peers-page.tsx
```diff
@@ -291,6 +435,89 @@ export const PeersPage = () => {
...
+                    )}
+                >
+                    <div className="flex items-center justify-between py-2">
+                        <div>
+                            <h3 className="text-lg font-semibold tracking-tight">
+                                Allowed peers
+                            </h3>
+                            <p className="text-muted-foreground text-sm">
+                                Accounts allowed to peer with this node when "Accept all
+                                incoming P2P connections" is disabled.
```
swatanabe: I think this should be next to Accept all incoming P2P connections, since they're so closely associated.

Would a radio button make sense instead of separate controls?

brandonfancher: Just to make sure I understand, you're recommending more closely colocating the "Accept all incoming P2P connections" control and the "Allowed peers" list, correct? Yeah, I could see that.

If we did a radio button, then I could see:

- [ ] Accept all incoming P2P connections
- [ ] Accept incoming P2P connections from whitelisted accounts only

The "Allowed peers" list could then show/hide based on what the user selected.

Is that what you're envisioning? That would be more explicit.

I don't want to bury the toggle/radio button below a long list of existing peer connections. Maybe I'd introduce a segmented control (tabs) to the top of the Peers page (as already exists on the Setup page): Connections and Settings... or something like that.

swatanabe: Yeah, that's more or less what I'm imagining.

brandonfancher: Cool, thanks for the feedback. Implemented.

James-Mart (approved):

brandonfancher:
I liked the suggestion yesterday to detect a live network and autocomplete account names when possible. Supporting account names lookup via prefix would require more work. I'd prefer that to be a separate body of work. But it would be a useful thing to have...not just for XAdmin here.

While there are other things we could do (_e.g._, differentiate known account names from unknown ones and flag the latter with a warning), this PR already achieves our objectives, and doing so increases complexity quite a bit. I'm going to move on to other higher-priority tasks for now.

We can circle back on this later.

James-Mart:
> We can circle back on this later.

Suggest you add & link a github issue if you want to remember this

PR #1997, opened 2026-08-13
Ship ubuntu 2604 release sdk

James-Mart:


.github/workflows/ubuntu-builder.yml
```diff
@@ -7,20 +7,33 @@ on:
...
+        value: ${{ jobs.ubuntu-versions.outputs.ubuntu-versions }}
     secrets:
       publish-package-login:
         required: false
 
 jobs:
+  ubuntu-versions:
+    runs-on: ubuntu-latest
+    outputs:
+      ubuntu-versions: ${{ steps.set-ubuntu-versions.outputs.ubuntu-versions }}
```
swatanabe: Can this use a literal string here, instead of jumping though `GITHUB_OUTPUT`?

swatanabe (approved):

.github/workflows/ubuntu-builder.yml
James-Mart: Changed to literal

PR #1999, opened 2026-08-14
Plugin WASM composition PoC

cool-ant:
While James is eliminating plugin hooks, I'm just using the current dynamic call approach to handle the exceptions.
Done in this PR:
- compose Host plugins into a composite plugin
- compose app plugins into a composite plugin
- implement dynamically-generated TracerPkg which "wrap" plugins and make calls out to `host:common` to maintain our callstack.
- export all intfs in composite plugins so calls to the same plugin and between Host and the app composite plugin work


James-Mart:
My goal is that this should not require changes to cascade throughout all the plugins. Changes should mostly be in supervisor, plus a new call-context plugin.

~~Rather than requiring plugins to import call-context, it should be the reverse: a load error if the plugin explicitly  imports call-context, unless the plugin in question is the TracerPkg plugin.~~ Nevermind, I thought `call-context` was a new call-stack management plugin

Don't rename the transact plugin, that adds a ton of noise to the pr.

~~What is going on with host:common/host:authed-http? I assume there's some kind of circularity you're trying to eliminate? If so, that should be a separate PR.~~ Discussed

PR #2000, opened 2026-08-14
Eliminate the Accounts-centered cycle (part of plugin composition work)

cool-ant:


James-Mart:
Congratulations! You opened PR 2K!!
If we planned better we could have made this PR a fix for a timestamp overflow

James-Mart:
```
Build left the working tree dirty:
 M packages/Cargo.lock
```

packages/system/Accounts/plugin/query/src/lib.rs
```diff
@@ -0,0 +1,133 @@
...
+    pub ErrorType
+    InvalidAccountName(msg: String) => "Invalid account name: {msg}",
+    QueryError(msg: String) => "Graphql query error: {msg}",
+    MaxGenerationAttemptsExceeded() => "Max generation attempts exceeded",
+    InvalidPrefix() => "Prefix must be a-z, 1 - 9 chars in length",
+}
+
+struct AccountsQuery;
+
+fn logged_in_user_table() -> Bucket {
```
James-Mart: Oh, interesting, yeah I guess we can read/write tables from different plugins as long as they are served from the same service.

Is there some complication with also moving the management of the logged-in state into this plugin as well, besides the name being "query"?

cool-ant: yes, as soon as you try to move any of the management code into this plugin, we get back to a cycle state.

James-Mart: I think at least `logged_in_user_table` should be imported from the `accounts` plugin, or from a third crate shared by both `accounts/` and `query/` instead of being redefined here. Potentially also any associated read-only interface used by everyone.

cool-ant: I don't see a way to put this in one plugin and import it into both that doesn't cause *some* kind of trouble or be worthwhile, so I just shared a tiny source file between the 2 plugins with that code, which I think is reasonable (although unusual for us, to my knowledge). Thoughts?

packages/system/Accounts/plugin/accounts/wit/world.wit
```diff
@@ -32,32 +23,13 @@ interface active-app {
     connect-account: func() -> result<_, error>;
 }
 
-/// This is the typical API intended to be used by third-party plugins or UIs.
+/// Restricted API for setting the logged-in user's auth service.
```
James-Mart: 1. It's not restricted. The old comment was more correct, although I don't particularly like that either because it *can* be used by third parties but it's not only for third parties. It's just the standard plugin interface.
2. You shouldn't mention "for setting the logged-in user's auth service". That elevates the one plugin function that happens to exist in this interface today to the sole purpose of the interface, which it is not.

packages/system/Accounts/plugin/accounts/src/lib.rs
James-Mart, line 1: Merge with main and verify that the changes made in https://github.com/gofractally/psibase/pull/2018 are not lost

James-Mart (approved):

PR #2001, opened 2026-08-14
Better release automation

James-Mart:
# Description

After this PR, making new psibase releases becomes significantly easier / more automated.

There's basically four steps to making a new release:

1. Open a PR to main that increments all version numbers
2. Merge that PR
3. Merge the (autogenerated) `cut/vX.Y.Z` → `release/vX` PR
4. Publish the (autogenerated) release draft

# Details

Here are more details on the four manual steps and what is happening automatically:

**Open a PR to main that increments all version numbers**

CI detects the change to the `PSIBASE_VERSION_*` macros in `...psibase/version.hpp`. Because it detects a version bump, it runs a dry-run of the crates publish on the PR.

Unfortunately, dry-running multi-workspace publishes in crates isn't great on our cargo version, so the dry run fails. So we detect the expected failure mode of the dry-run script (only the expected error shape is allowed: "failed to select a version for the requirement") and treat it as a passing dry-run. When we upgrade to rust 1.90 we can improve this because there's new affordances for [publishing whole workspaces](https://www.infoworld.com/article/4060262/rust-1-90-brings-workspace-publishing-support-to-cargo.html).

**Merge the version bump PR**

When merged, `cut-release.yml` detects the version bump on main. It creates `release/v<major>` at the pre-bump tip (if DNE), freezes the bump commit on `cut/vX.Y.Z`, and opens a `cut/vX.Y.Z` → `release/vX` PR. The cut branch does not track later merges to main.

**Merge the cut → release PR**

I usually just push directly to the release branch (bypassing branch restrictions) because I'm lazy, but in this automated flow I wanted to leave this as a manual step instead of automatic because it's possible to patch the release branch directly which means it's possible for the cut -> release PR to have merge conflicts. 

Conflict fixes go on the cut branch. A direct push to `release/vX` does not tag or draft. that only runs on a GitHub merge of a PR into `release/v*`.

When merged, `cut-release.yml` tags the merge commit. It also verifies head major version matches `X` in the branch name `release/v<X>` before tagging. Tag name is resolved automatically to `v0.Y.Z-pre` when major is 0, else `vX.Y.Z`.

The bot pushes the tag with `GITHUB_TOKEN`, which does not trigger `on-tag.yml`. The same run continues in `cut-release.yml`:

- `ubuntu-builder.yml` builds at the tag
- We auto-generate the changelog
  - **New: The workflow was updated to automatically prepend the standard disclaimer to the top of any v0 release**
- Create a draft GH release (with attached SDK artifacts)

(`on-tag.yml` still exists for a manual tag push.)

**Publish the release**

This is still a manual step. It's the same manual final check we are used to.

After publishing, `on-release.yml`:

- Dispatches Docker image builds
- Sends telegram notification
- New: Runs `rust/scripts/publish.sh` to crates.io using `CARGO_REGISTRY_TOKEN`


rust/scripts/publish.sh
```diff
@@ -35,6 +35,13 @@ done
...
     cd "$dir"
     echo "Publishing $dir..."
-    cargo publish $1
+    if ! out=$(cargo publish $1 2>&1); then
+        printf '%s\n' "$out" >&2
+        if [[ "$1" != "--dry-run" ]] && ! printf '%s\n' "$out" | grep -q 'already exists on crates.io index'; then
+            exit 1
+        fi
+    else
+        printf '%s\n' "$out"
```
swatanabe, line 44: Is there a reason for writing to STDOUT vs STDERR on the other branch?

.github/scripts/expect-crates-dry-run-errors.sh
```diff
@@ -0,0 +1,115 @@
+#!/usr/bin/env bash
```
swatanabe, line 1: This kind of code is usually really fragile. How is this better than updating our cargo version?

.github/workflows/cut-release.yml
```diff
@@ -0,0 +1,172 @@
...
+          fi
+          if [[ $tag == *-pre ]]; then
+            message="psibase prerelease version v$ver"
+          else
+            message="psibase version v$ver"
+          fi
+          # 41898282+github-actions[bot]@users.noreply.github.com is
+          # GitHub's noreply identity for github-actions[bot] (user id 41898282).
+          git config user.name "github-actions[bot]"
+          git config user.email "41898282+github-actions[bot]@users.noreply.github.com"
```
swatanabe, line 148: This looks really hacky. I can't find this account documented officially anywhere. The tag is not signed and cannot be signed because we don't have access to keys for this email.

It looks like it should be possible to create the tag using the github API, though it seems to require two steps. First [create an annotated tag](https://docs.github.com/en/rest/git/tags?apiVersion=2026-03-10#create-a-tag-object). The response should include the sha of the tag object, which can then be used to [create /refs/tags/$tag](https://docs.github.com/en/rest/git/refs?apiVersion=2026-03-10#create-a-reference).

.github/scripts/expect-crates-dry-run-errors.sh
James-Mart: It's not better. 

If we update our cargo version I'd update this, but I didn't know what kind of unexpected issues there might be with updating cargo.

Do you want to block this until I update cargo or can we push this through and come back for cleanup later?

.github/workflows/cut-release.yml
James-Mart: Fixed

rust/scripts/publish.sh
James-Mart: No, done

.github/scripts/expect-crates-dry-run-errors.sh
swatanabe: Updating cargo shouldn't be a problem. I've always been using a more recent version of cargo. I've only seen two issues:

- The tests that check for exact error messages need to be updated
- Extra warnings (which we can clean up later, as they don't fail the tests)

PR #2002, opened 2026-08-14
X admin sibling fetches present x admin jwt

James-Mart:
Fix various permissions issues with x-admin making subdomain requests in production.

### DIFF
--- a/packages/local/XAdmin/src/XAdmin.cpp
+++ b/packages/local/XAdmin/src/XAdmin.cpp
@@ -613,6 +613,8 @@ namespace LocalService
          bool                    operator()(Auth::Account&& account)
          {
             hasUser = true;
+            if (account.value == XAdmin::service)
+               return true;
             bool result;
             PSIBASE_SUBJECTIVE_TX
             {
--- a/packages/local/XAdmin/src/XAdmin.cpp
+++ b/packages/local/XAdmin/src/XAdmin.cpp
@@ -613,8 +613,6 @@ namespace LocalService
          bool                    operator()(Auth::Account&& account)
          {
             hasUser = true;
-            if (account.value == XAdmin::service)
-               return true;
             bool result;
             PSIBASE_SUBJECTIVE_TX
             {
--- a/packages/local/XAdmin/ui/src/lib/admin-login-jwt.ts
+++ b/packages/local/XAdmin/ui/src/lib/admin-login-jwt.ts
@@ -1,0 +1,26 @@
+import { getJson } from "@psibase/common-lib";
+import { z } from "zod";
+
+const AdminLoginReply = z
+    .object({
+        access_token: z.string(),
+        token_type: z.string(),
+    })
+    .strict();
+
+let accessToken: string | undefined;
+let fetchPromise: Promise<string> | undefined;
+
+export async function getAdminLoginAccessToken(): Promise<string> {
+    if (accessToken !== undefined) {
+        return accessToken;
+    }
+    if (fetchPromise === undefined) {
+        fetchPromise = (async () => {
+            const reply = AdminLoginReply.parse(await getJson("/admin_login"));
+            accessToken = reply.access_token;
+            return accessToken;
+        })();
+    }
+    return fetchPromise;
+}
--- a/packages/local/XAdmin/ui/src/lib/admin-login-jwt.ts
+++ b/packages/local/XAdmin/ui/src/lib/admin-login-jwt.ts
@@ -1,12 +1,10 @@
 import { getJson } from "@psibase/common-lib";
 import { z } from "zod";
 
-const AdminLoginReply = z
-    .object({
-        access_token: z.string(),
-        token_type: z.string(),
-    })
-    .strict();
+const AdminLoginReply = z.object({
+    access_token: z.string(),
+    token_type: z.string(),
+});
 
 let accessToken: string | undefined;
 let fetchPromise: Promise<string> | undefined;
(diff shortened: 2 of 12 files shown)


packages/local/XAdmin/src/XAdmin.cpp
```diff
@@ -613,6 +613,8 @@ namespace LocalService
          bool                    operator()(Auth::Account&& account)
          {
             hasUser = true;
+            if (account.value == XAdmin::service)
```
James-Mart: @swatanabe A bunch of requests use this via XAdmin `checkAuth` rather than `isAdmin`. Without this, even if I attach the JWT I get from `/admin_login`, the requests 403

Is this the right approach?

e.g. shouldn't this work?
```
$ TOKEN=$(curl -sS http://x-admin.psibase.localhost:8080/admin_login | jq -r .access_token)
$ curl -sS -D- -o /dev/null \
  -H 'X-Forwarded-For: 8.8.8.8' \
  -H "Authorization: Bearer $TOKEN" \
  http://x-admin.psibase.localhost:8080/config
HTTP/1.1 403 Forbidden
```

swatanabe: The token was originally intended to be used for on-chain services, which should use `isAdmin`. The tokens are not available until the chain is booted, because they're managed by RTransact, which is fine for on-chain accounts, but doesn't quite work for a local account accessing a local service.

James-Mart: Aha, okay reverted. I guess those endpoints will still 403 for now.

packages/local/XAdmin/ui/src/lib/admin-login-jwt.ts
```diff
@@ -0,0 +1,32 @@
...
+let accessToken: string | undefined;
+let fetchPromise: Promise<string> | undefined;
+
+export async function getAdminLoginAccessToken(): Promise<string> {
+    if (accessToken !== undefined) {
+        return accessToken;
+    }
+    if (fetchPromise === undefined) {
+        fetchPromise = (async () => {
+            const reply = AdminLoginReply.parse(await getJson("/admin_login"));
```
swatanabe, line 18: Should check that `token_type` is `"bearer"` before using it.

packages/local/XAdmin/ui/src/lib/admin-login-jwt.ts
```diff
@@ -0,0 +1,32 @@
+import { getJson } from "@psibase/common-lib";
+import { z } from "zod";
+
+const AdminLoginReply = z
+    .object({
+        access_token: z.string(),
+        token_type: z.string(),
+    })
+    .strict();
```
swatanabe: This probably shouldn't be strict. The format is based on OAuth2, which requires unknown fields to be ignored. At some point we might want to add more fields, like `expires_in`.

packages/local/XAdmin/ui/src/lib/chain-endpoints.ts
```diff
@@ -240,6 +268,10 @@ class Chain {
         const url = siblingUrl(null, "x-peers", "/connect");
         const result = await postJsonGetJson(url, config, {
```
swatanabe, line 256: Can you merge `Content-Type` with the headers passed in explicitly, in `postJsonGetJson`?

packages/local/XAdmin/ui/src/lib/chain-endpoints.ts
```diff
@@ -281,7 +313,13 @@ class Chain {
 
     public async getTransactStats(): Promise<TransactStatsType> {
         const url = siblingUrl(null, "transact", "/stats");
-        return TransactStats.parse(await getJson(url));
```
swatanabe, line 284: `getJson` can have an `options` parameter, too. I only added it where I actually needed it.

packages/local/XAdmin/ui/src/lib/chain-endpoints.ts
James-Mart: Done

packages/local/XAdmin/ui/src/lib/admin-login-jwt.ts
James-Mart: Done

packages/local/XAdmin/ui/src/lib/admin-login-jwt.ts
James-Mart: Removed

packages/local/XAdmin/ui/src/lib/chain-endpoints.ts
James-Mart: Added

packages/user/CommonApi/common/packages/common-lib/src/rpc.ts
```diff
@@ -111,33 +134,31 @@ export async function postGraphQL(
...
         await fetch(url, {
             method: "POST",
-            headers: {
-                "Content-Type": "application/graphql",
-            },
             body: graphql,
-            ...options,
+            ...rest,
+            headers: mergeRequestHeaders(
+                { "Content-Type": "application/graphql" },
```
swatanabe: I had thought that this would be just
```js
...options,
headers: { "Content-Type": "application/graphql", ...options?.headers }
```

- Later fields override earlier fields with the same name
- spreading `undefined` into an object is valid and adds no fields

James-Mart: Done

swatanabe (approved):

brandonfancher (approved):

PR #2003, opened 2026-08-15
host decycle complete and smoke tested

cool-ant:


packages/user/Host/plugin/http/wit/world.wit
```diff
@@ -0,0 +1,31 @@
+package host:http;
+
+/// HTTP to an app's own server, with the current session attached when present.
+/// Session setup must not go through this plugin.
```
James-Mart: ~~Session setup must not go through this plugin~~

packages/user/Host/plugin/http/src/lib.rs
```diff
@@ -0,0 +1,114 @@
...
+        }
+
+        Err(make_error("Invalid graphql response: 'body' must be JSON"))
+    }
+
+    fn post(request: PostRequest) -> Result<BodyTypes, Error> {
+        let endpoint = normalize_endpoint(request.endpoint);
+        let res = do_post(CallContext::get_sender(), endpoint, request.body)?;
+
+        match res.body {
```
James-Mart, line 98: There used to be a comment above this line that should be preserved
```
        // TODO: post should return Option<BodyTypes> because not all posts return a body
```

packages/user/Host/plugin/auth/src/lib.rs
```diff
@@ -29,22 +33,32 @@ fn bucket_id(user: &str) -> String {
     format!("query_tokens-{}", user)
 }
 
+fn cookie_request(app: &str, endpoint: &str, body: String) -> HttpRequest {
```
James-Mart: If the endpoint is parameterized, there's nothing about this function that is inherently a cookie request

packages/user/Host/plugin/client/wit/world.wit
```diff
@@ -0,0 +1,24 @@
+package host:client;
+
+/// Facts about the current plugin call, derived from the service call stack.
+/// This plugin is a leaf: it imports only the bridge (service-stack, root domain).
```
James-Mart: I prefer the old interface comment
```
/// Common calls related to client interactions
```

rust/psibase_plugin/wit/deps/host-client/world.wit
James-Mart, line 1: This is supposed to be a symlink rather than a copy

rust/psibase_plugin/wit/deps/host-http/world.wit
James-Mart, line 1: This is also supposed to be a symlink rather than a copy

James-Mart (approved):

PR #2004, opened 2026-08-15
Transact decycle complete and smoke tested

cool-ant:


packages/system/Transact/plugin/transact/wit/world.wit
```diff
@@ -104,6 +83,34 @@ interface intf {
...
+    /// logged-in user. Only callable by the `packages` plugin.
+    ///
+    /// * `actions`: The actions to propose.
+    /// * `auto-exec`: If true, the staged transaction will execute as soon as it
+    ///                has enough approvals.
+    propose: func(actions: list<action>, auto-exec: bool) -> result<_, error>;
+}
+
+/// Driver-only open-tx ledger. Callable only by service `transact` (the tx admin driver).
+interface ledger {
```
James-Mart, line 97: I'm assuming this interface exists because the transactions are stored in memory instead of being written to the db? 

If so, that's only because the transact plugin existed before plugins could use client-side storage.

So the correct solution is for the transact plugin to simply write transactions to an ephemeral db, and then there's also no need for any explicit clearing (besides the standard clearing of all the ephemeral dbs). Then you can simply share the same table reads from both transact plugins.

packages/system/Transact/plugin/transact/wit/world.wit
```diff
@@ -126,13 +133,16 @@ interface authorized {
...
 
-// Include this world to import the most common transact:plugin imports
+interface network {
+    use host:types/types.{error};
+
+    /// Set snapshot interval period network wide
+    set-snapshot-time: func(seconds: u32) -> result<_, error>;
+}
+
+/// Import this world to append actions / set the propose latch / propose.
```
James-Mart, line 143: "the most common transact:plugin imports" was more general. Now you've coupled the comment to what it implements and it would have to be updated if we add more functionality.

packages/user/Host/plugin/db/src/bucket.rs
```diff
@@ -130,6 +130,7 @@ impl Bucket {
     }
 
     fn validate_identifier(identifier: &str) {
+        // Validate identifier with regex /^[0-9a-zA-Z-]+$/
```
James-Mart, line 133: No comment needed, and this comment is wrong anyway

James-Mart (commented):
Initial high level review

packages/user/Supervisor/ui/src/supervisor.ts
James-Mart: These changes break the supervisor test you added in
`packages/user/Supervisor/ui/src/supervisor-concurrent-entry.test.ts`

James-Mart (changes requested):

packages/user/Supervisor/ui/src/supervisor.ts
cool-ant: fixed

James-Mart (approved):

PR #2014, opened 2026-08-20
add transient key imports and explicit extra signatures for invites

James-Mart:
PR #2 in the series that started with https://github.com/gofractally/psibase/pull/1831

This is progress in the direction of the removal of our custom callback infrastructure. It would be better to simply do without callbacks until the component model defines a standard solution for them, and only then add them, instead of using this non-portable PluginRef solution.

This one replaces `hook-action-auth`, with an explicit `transact.add-signature` plus a host `import-key-transient`. `import-key-transient` allows the host to sign for a pubkey that was imported in the same call context. 

This PR also simplifies the `credentials` plugin (which was the only other consumer of hook-action-auth).

### DIFF
--- a/packages/user/Host/plugin/crypto/wit/deps/supervisor.wit
+++ b/packages/user/Host/plugin/crypto/wit/deps/supervisor.wit
@@ -13,6 +13,11 @@ interface intf {
     /// Imports a PEM-encoded private key into the keyvault.
     /// Returns the corresponding public key in PEM format.
     import-key: func(private-key: pem) -> result<pem, error>;
+
+    /// Imports a PEM-encoded private key into a transient keyvault that
+    /// is automatically cleared at the end of the current transaction
+    /// context. Returns the corresponding public key in PEM format.
+    import-key-transient: func(private-key: pem) -> result<pem, error>;
 }
 
 world imports {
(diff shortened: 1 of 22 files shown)


packages/user/Host/plugin/crypto/wit/deps/supervisor.wit
```diff
@@ -13,6 +13,11 @@ interface intf {
     /// Imports a PEM-encoded private key into the keyvault.
     /// Returns the corresponding public key in PEM format.
     import-key: func(private-key: pem) -> result<pem, error>;
+
+    /// Imports a PEM-encoded private key into a transient keyvault that
+    /// is automatically cleared at the end of the current transaction
```
swatanabe, line 18: I generally prefer to say that the key lives until the end of the current transaction or that it is only usable in the current transaction. Saying that it is cleared is an implementation perspective.

swatanabe (approved):

packages/user/Host/plugin/crypto/wit/deps/supervisor.wit
James-Mart: Will address in forthcoming pr

PR #2017, opened 2026-08-20
Secret storage

swatanabe:
Secrets are encrypted using a master passphrase. There are two methods for passing it to `psinode`:
- A file whose name is provided on the command line. This is useful for [systemd](https://systemd.io/CREDENTIALS/), [docker compose](https://docs.docker.com/compose/how-tos/use-secrets/) and any other environments that can pass secrets securely as files.
- The environmental variable `PSIBASE_PASSPHRASE`.
- If no passphrase is provided, the secrets will only be kept in memory (matching the current behavior)

The pins for unlocked devices are stored as secrets and they will be automatically unlocked when the node starts again (if the device is still present). Secrets are only used by native and are not available to services yet.

### DIFF
--- a/doc/src/run-infrastructure/cli/psinode.md
+++ b/doc/src/run-infrastructure/cli/psinode.md
@@ -42,6 +42,10 @@ psinode - The psibase blockchain server
 
   Use this private key to sign blocks. Any number of keys may be provided, but only the one that matches the public key corresponding to the producer name will be used.
 
+- `--passphrase-file` *filename*
+
+  Provides a passphrase that will be used to encrypt secrets stored by the node. The passphrase can also be passed in the environment variable `PSIBASE_PASSPHRASE`. If no passphrase is provided, secrets will not be stored and will be discarded when the server exits.
+
 - `--pkcs11-module` *filename*
 
   Loads a PKCS #11 module from *filename*. The server will be able to sign blocks using keys from the module. The tokens that the module provides must be unlocked using the HTTP API before they can be used. This option can appear any number of times.
--- a/libraries/psibase/native/src/Secrets.cpp
+++ b/libraries/psibase/native/src/Secrets.cpp
@@ -1,0 +1,206 @@
+#include <psibase/Secrets.hpp>
+
+#include <openssl/evp.h>
+#include <openssl/rand.h>
+#include <cassert>
+#include <cstring>
+#include <fstream>
+#include <psibase/check.hpp>
+#include <psibase/fileUtil.hpp>
+#include <psio/fracpack.hpp>
+#include <psio/reflect.hpp>
+#include <vector>
+
+using namespace psibase;
+
+// AES-GCM
+// magic (4-bytes) || version (4-bytes) || iv (16-bytes) || cyphertext || tag (16-bytes)
+
+namespace
+{
+   constexpr unsigned char prefix[] = {0, 0xcf, 0x88, 's', 1, 0, 0, 0};
+
+   constexpr std::size_t prefix_size = sizeof(prefix);
+   constexpr std::size_t block_size  = 16;
+   constexpr std::size_t iv_size     = block_size;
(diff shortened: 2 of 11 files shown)


programs/psinode/main.cpp
```diff
@@ -1865,6 +1912,28 @@ int main(int argc, char* argv[])
        "Name of this producer");
    opt("key,k", po::value(&keys->provers)->default_value({}, ""),
        "A private key to use for block production");
+   opt("passphrase-file",
+       po::value<std::string>()->default_value("")->notifier(
+           [&](const std::string& filename)
+           {
+              if (filename != "")
+              {
+                 auto contents = readWholeFile(filename);
```
James-Mart: Consider stripping trailing newlines. Generated pw files might have them.

doc/src/run-infrastructure/cli/psinode.md
```diff
@@ -42,6 +42,10 @@ psinode - The psibase blockchain server
 
   Use this private key to sign blocks. Any number of keys may be provided, but only the one that matches the public key corresponding to the producer name will be used.
 
+- `--passphrase-file` *filename*
+
+  Provides a passphrase that will be used to encrypt secrets stored by the node. The passphrase can also be passed in the environment variable `PSIBASE_PASSPHRASE`. If no passphrase is provided, secrets will not be stored and will be discarded when the server exits.
+
 - `--pkcs11-module` *filename*
 
   Loads a PKCS #11 module from *filename*. The server will be able to sign blocks using keys from the module. The tokens that the module provides must be unlocked using the HTTP API before they can be used. This option can appear any number of times.
```
James-Mart, line 51: > The tokens that the module provides must be unlocked using the HTTP API

No longer true

libraries/psibase/native/src/Secrets.cpp
```diff
@@ -0,0 +1,342 @@
...
+// KDF uses scrypt
+// salt (16 bytes) || logn (4-bytes) || r (4-bytes) || p (4-bytes)
+
+namespace
+{
+   constexpr unsigned char prefix[] = {0, 0xcf, 0x88, 's', 1, 0, 0, 0};
+
+   constexpr std::size_t prefix_size = sizeof(prefix);
+   constexpr std::size_t block_size  = 16;
+   constexpr std::size_t iv_size     = block_size;
```
James-Mart: An LLM find is that the `iv_size` here is 16 bytes, but the default length used by openssl appears to be only 12 bytes unless you call `EVP_CIPHER_CTX_ctrl` to specify the length.

> The following _ctrl_s are supported in GCM and OCB modes.
> 
> `EVP_CIPHER_CTX_ctrl(ctx, EVP_CTRL_AEAD_SET_IVLEN, ivlen, NULL)`
> 
> Sets the IV length. This call can only be made before specifying an IV. If not called a default IV length is used.
> 
> For GCM AES and OCB AES the default is 12 (i.e. 96 bits). 

[Source](https://docs.openssl.org/3.0/man3/EVP_EncryptInit/#gcm-and-ocb-modes)

doc/src/run-infrastructure/cli/psinode.md
swatanabe: It's still true. Unlocking is just persistent across server restarts.

libraries/psibase/native/src/Secrets.cpp
swatanabe: Right. The iv is concatenated with the counter to make 16 bytes.

doc/src/run-infrastructure/cli/psinode.md
James-Mart: Well, ok. "must" is a bit strong.
I thought the auto unlocking behavior might be worth a mention here.

swatanabe: I don't think this is the right place for it.

James-Mart (approved):

PR #2021, opened 2026-08-24
Fix namemarket query

brandonfancher:


swatanabe (commented):
Can we add the tests for `NameMarket` to ctest (in packages/CMakeLists.txt)? I even see tests using the query which should have caught this.

swatanabe (approved):

PR #2026, opened 2026-08-25
Fix XProxy service retrieval

brandonfancher:
Note: This code is entirely AI generated; scrutinize accordingly.

XProxy stopped working.

`x-http::serve failed: service 'db' aborted with message: x-proxy cannot read from another service's prefix: http:0` showing up on proxied requests like `GET network.psibase.localhost:8080/ 500`

### DIFF
--- a/packages/local/XProxy/src/XProxy.cpp
+++ b/packages/local/XProxy/src/XProxy.cpp
@@ -141,13 +141,24 @@ std::optional<HttpReply> XProxy::serveSys(HttpRequest req, std::optional<std::in
    }
    else
    {
+      // Load any configured upstream first so bypass prefixes can short-circuit
+      // before we touch HttpServer (e.g. /common should fall through to the
+      // normal HttpServer/common-api path).
+      auto                           table = open<OriginServerTable>();
+      std::optional<OriginServerRow> originServer;
+      PSIBASE_SUBJECTIVE_TX
+      {
+         originServer = table.get(subdomain);
+      }
+      if (originServer && originServer->bypassPrefixes &&
+          pathMatchesBypassPrefix(req.path(), *originServer->bypassPrefixes))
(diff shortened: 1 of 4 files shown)


packages/local/XProxy/src/XProxy.cpp
```diff
@@ -156,57 +167,47 @@ std::optional<HttpReply> XProxy::serveSys(HttpRequest req, std::optional<std::in
...
-                          .to<ServerInterface>(registered->server)
+                          .to<ServerInterface>(*registered)
                           .serveSys(req, socket, user);
 
          if (!to<XHttp>().takeSocket(*socket, false))
             return {};
          if (reply)
             return reply;
       }
-   }
```
swatanabe, line 167: Putting everything after this inside the `else` is wrong.

swatanabe (changes requested):

packages/local/XProxy/src/XProxy.cpp
brandonfancher: Addressed. Thank you.

packages/system/HttpServer/include/services/system/HttpServer.hpp
```diff
@@ -96,6 +96,9 @@ namespace SystemService
       /// * Respond to GraphQL requests
       void registerServer(psibase::AccountNumber server);
 
+      /// Returns the server registered for `service`, if any
+      std::optional<psibase::AccountNumber> getServer(psibase::AccountNumber service);
```
swatanabe, line 100: This needs to be added to the rust service declarations, too.

brandonfancher: Ah, thanks. Added.

swatanabe (approved):

PR #2028, opened 2026-08-27
Fix fractals scoring / ppm bug

Velua:
During distribution of tokens to fractal members, function `member_reward_shares` is responsible for producing a `HashMap` of users and their slice of the token allocation. 

Function `allocations` is used to divide a total / budget across a list of accounts, however it expects each members allocation to be in `ppm` e.g. 500,000 for 50% but instead it's passed the results of `occu_wrapper::call_to(occ.occupation) .get_scores(self.account)` which returns a raw score. 

So in a 3 man Fractal, alice, bob and charlie each with a score of 10, **should** be sending in 333,333 each into `allocations` however it's just `10` instead, resulting in tiny allocations and huge amounts of tokens recycled as dust. 

I've also slipped in renaming away from the `share` idea as it's ambiguous between a share in a company, which is relative to total shares, or ppm share, which is always out of 1,000,000.

### DIFF
--- a/packages/user/Fractals/service/src/tables/fractal.rs
+++ b/packages/user/Fractals/service/src/tables/fractal.rs
@@ -1,7 +1,7 @@
 use async_graphql::connection::Connection;
 use async_graphql::ComplexObject;
 use psibase::services::sites;
-use psibase::services::tokens::{Precision, Quantity};
+use psibase::services::tokens::{Decimal, Precision, Quantity};
 
 use crate::constants::{token_distributions::TOKEN_SUPPLY, TOKEN_PRECISION};
 use crate::constants::{
@@ -11,7 +11,7 @@ use crate::constants::{
 use psibase::services::fractals::distribute::{allocations, combine_group_scores};
 use psibase::services::fractals::weighted_normalization::{
     curves::{get_curve, Curve},
-    weighted_normalization,
+    weighted_normalization, HasScore,
 };
 use psibase::services::fractals::FractalRole::{
     self, Executive, Judiciary, Legislature, Recruitment,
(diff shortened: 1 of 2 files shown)


packages/user/Fractals/service/src/tables/fractal.rs
```diff
@@ -169,20 +180,31 @@ impl Fractal {
...
+            .into_iter()
+            .zip(weighted_occupations)
+            .map(|(occupation, weight)| {
+                let member_scores: Vec<MemberScore> = occu_wrapper::call_to(occupation.occupation)
+                    .get_scores(self.account)
+                    .into_iter()
+                    .filter(|(m, _)| is_member(m))
+                    .map(|(account, score)| MemberScore { account, score })
+                    .collect();
+                let member_ppm =
```
James-Mart, line 201: This is the wrong fix.

There is supposed to be a very strong symmetry between the way guilds computes the scores of its members (aggregating across the ranked guilds) and the way fractals is supposed to compute the scores of its members (aggregating across the occupations).

| | Guilds | Fractals |
|---|---|---|
| Group PPM | Ranked guilds → guild-weight curve → guild PPM | Ranked occupations → dist_strat → occupation PPM |
| Member PPM | Each guild: `Guild::scores()` → member PPM of that guild | Each occupation: `get_scores()` → member PPM of that occupation |
| Combine | `combine_group_scores(guild PPM, member PPM)` → member PPM of the occupation, returned by `<occupation>::get_scores` | `combine_group_scores(occupation PPM, member PPM)` → member PPM of the fractal |

----

So I think guilds and fractals services were fine, the issue was just that the `de-facto` occupation is non-compliant. It should be returning a PPM (it returns 1 for each member instead of a normalizing to PPM).

James-Mart: That's the intended contract. There may be some function/variable renames, and/or doc comment updates that could make it more clear what the public interface is supposed to return.

Velua:
Following discussion with @James-Mart I'll re-approach this by updating `de-facto`'s as it's `get_scores` behaviour is out of line.

PR #2030, opened 2026-08-27
Adjust number handling in AuthDyn

Velua:
I didn't realise when I wrote the code for `AuthDyn` around [overflow](https://doc.rust-lang.org/book/ch03-02-data-types.html#integer-overflow) that Rust only panics when compiling in debug mode, however in release mode it will overflow. 

I'm wondering if we got something that changes this process given the context the services run in, however if not, I've adopted `checked_add` with unwrap to ensure the clean panic I was expecting.

### DIFF
--- a/packages/system/AuthDyn/service/src/lib.rs
+++ b/packages/system/AuthDyn/service/src/lib.rs
@@ -172,7 +172,7 @@ pub mod service {
         let total_possible_weight = policy
             .authorizers
             .iter()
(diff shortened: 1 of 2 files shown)


James-Mart:
> I'm wondering if we got something that changes this process given the context the services run in

It's a compile-time problem, so `checked_add`/`unwrap` is better here.

James-Mart (approved):

PR #2031, opened 2026-08-27
Fix websocket deadlock

James-Mart:
This should resolve the deadlock, and I can't see why any race window between `unlock()` and `error()` should matter.

This is what is currently hanging my testnet node.

This also fixes a `==/=` typo in `handleP2P` that left `p2pState` stuck in `reading`, dropping early p2p messages.

### DIFF
--- a/libraries/psibase_http/websocket.hpp
+++ b/libraries/psibase_http/websocket.hpp
@@ -350,13 +350,12 @@ namespace psibase::http
                    std::unique_lock l{self->mutex};
                    if (ec)
                    {
+                      auto outbox = std::move(self->outbox);
+                      l.unlock(); // `error()` locks the mutex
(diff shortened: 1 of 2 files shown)


libraries/psibase_http/websocket.hpp
```diff
@@ -350,13 +350,12 @@ namespace psibase::http
                    std::unique_lock l{self->mutex};
                    if (ec)
                    {
+                      auto outbox = std::move(self->outbox);
+                      l.unlock(); // `error()` locks the mutex
```
swatanabe: I think this creates a race. Emptying the outbox and setting the error flag (which prevents more items from being pushed into the outbox) need to be done atomically.

swatanabe (approved):

PR #2032, opened 2026-08-28
Add recall shared and sub balances

### LINKED ISSUE #2027: Tokens recall misses sub accounts and shared balances
One can prevent any recall of their tokens simply by putting the tokens into a sub balance, or otherwise just crediting them to a shared balance. 

How do we want to handle this?
Do we allow the token admin to recall shared balances and sub balances too? Can they specify which shared balance or sub balance? Or should the tokens service naively loop through each of them until a sufficient amount of tokens is recalled?

Velua:
Awkwardness around `recall` as it was only able to recall from root balances, anyone could prevent their tokens from being recalled by simply storing them in either a sub balance or a shared balance.

### DIFF
--- a/packages/user/Tokens/plugin/src/lib.rs
+++ b/packages/user/Tokens/plugin/src/lib.rs
@@ -53,12 +53,40 @@ impl Issuer for TokensPlugin {
     }
 
     #[psibase_plugin::authorized(High)]
-    fn recall(token_id: u32, amount: String, memo: String, account: String) -> Result<(), Error> {
+    fn recall(
+        token_id: u32,
+        amount: String,
+        memo: String,
+        account: String,
+        sub_account: Option<String>,
+    ) -> Result<(), Error> {
         let amount = Self::non_zero(token_id, amount)?;
+        let sub_account = sub_account.filter(|s| !s.is_empty());
 
         Tokens::add_to_tx().recall(
             token_id,
             account.parse().unwrap(),
+            amount,
+            memo.as_str().into(),
+            sub_account,
+        );
+        Ok(())
+    }
+
+    #[psibase_plugin::authorized(High)]
+    fn recall_shared(
+        token_id: u32,
+        amount: String,
+        memo: String,
+        creditor: String,
+        debitor: String,
+    ) -> Result<(), Error> {
+        let amount = Self::non_zero(token_id, amount)?;
+
+        Tokens::add_to_tx().recallShared(
+            token_id,
+            creditor.parse().unwrap(),
+            debitor.parse().unwrap(),
             amount,
             memo.as_str().into(),
         );
--- a/packages/user/Tokens/service/src/tables.rs
+++ b/packages/user/Tokens/service/src/tables.rs
@@ -197,7 +197,11 @@ pub mod tables {
         }
 
         pub fn burn(&mut self, amount: Quantity) {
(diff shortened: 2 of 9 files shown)


packages/user/Tokens/plugin/src/lib.rs
```diff
@@ -53,14 +53,42 @@ impl Issuer for TokensPlugin {
...
-    fn recall(token_id: u32, amount: String, memo: String, account: String) -> Result<(), Error> {
+    fn recall(
+        token_id: u32,
+        amount: String,
+        memo: String,
+        account: String,
+        sub_account: Option<String>,
+    ) -> Result<(), Error> {
         let amount = Self::non_zero(token_id, amount)?;
+        let sub_account = sub_account.filter(|s| !s.is_empty());
```
James-Mart: What does this do?

packages/user/Tokens/service/src/tables.rs
```diff
@@ -200,15 +200,46 @@ pub mod tables {
...
+            self.check_can_recall();
+            assert!(amount.value > 0, "burn quantity must be greater than 0");
+
+            SharedBalance::get_assert(creditor, debitor, self.id).recall(amount);
+
+            self.burned_supply = self.burned_supply + amount;
+            self.save();
         }
 
         fn burn_supply(&mut self, amount: Quantity, from: AccountNumber) {
```
James-Mart: Since the only caller of this is now the burn action, it can probably just be inlined.

packages/user/Tokens/service/src/tests.rs
```diff
@@ -143,6 +143,145 @@ mod tests {
...
+        );
+
+        // Recall from a shared balance
+        a.recallShared(tid, bob, carol, 100.into(), memo.clone())
+            .get()?;
+        assert_eq!(0, b.getBalance(tid, bob).get()?.value);
+        assert_eq!(0, get_shared_balance(&chain, tid, bob, carol).value);
+        assert_eq!(0, c.getBalance(tid, carol).get()?.value);
+
+        // Incoming undebited credits are not the debitor's to recall
```
James-Mart: This test is not showing that the debitor can't recall, as the comment suggests. 
It's showing that the issuer can't recall tokens from the wrong shared balance.

packages/user/Tokens/test/src/Tokens-test.cpp
```diff
@@ -246,13 +247,57 @@ SCENARIO("Recalling tokens")
...
+            CHECK(a.getBalance(tokenId, bob).returnVal().value == 0);
+            CHECK(b.getSharedBal(tokenId, bob, carol).returnVal().value == 0);
+            CHECK(c.getBalance(tokenId, carol).returnVal().value == 0);
+         }
+         THEN("Alice recalling Carol as creditor does not take Bob's in-flight credit")
+         {
+            CHECK(a.recallShared(tokenId, carol, bob, 1'000e4, memo).failed(missingSharedBalance));
+            CHECK(b.getSharedBal(tokenId, bob, carol).returnVal().value == 1'000e4);
+         }
+      }
```
James-Mart: If you're testing this in the rust unit tests, we don't need to also test them in C++

There's probably also duplication between what you added and what was already in C++, but I didn't check.

James-Mart (approved):

PR #2034, opened 2026-08-28
Fix min account name length bug

Velua:
Fractals plugin would throw if the account name length was less than 10 instead of 8.

### DIFF
--- a/packages/user/Fractals/plugin/src/helpers.rs
+++ b/packages/user/Fractals/plugin/src/helpers.rs
@@ -1,11 +1,9 @@
 use std::str::FromStr;
 
 use crate::errors::ErrorType::*;
+use psibase::services::accounts::MIN_ALLOWED_ACCOUNT_LENGTH;
 use psibase::AccountNumber;
 use psibase_plugin::{host::client as Client, Error};
-
-/// Free account names are 8-10 characters; shorter names must be purchased.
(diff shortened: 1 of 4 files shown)


packages/user/Fractals/plugin/src/helpers.rs
```diff
@@ -4,14 +4,19 @@ use crate::errors::ErrorType::*;
 use psibase::AccountNumber;
 use psibase_plugin::{host::client as Client, Error};
 
+/// Free account names are 8-10 characters; shorter names must be purchased.
+pub const MIN_FREE_ACCOUNT_NAME_LENGTH: usize = 8;
```
James-Mart: You should import `MIN_ALLOWED_ACCOUNT_LENGTH` directly from the accounts service in `accounts.rs` instead of maintaining a separate variable

brandonfancher (approved):

brandonfancher:
I approve the frontend changes (pending processing of James' comment)

packages/user/Fractals/plugin/src/helpers.rs
Velua: Addressed in d68368611: `validate_account_name` now uses `psibase::services::accounts::MIN_ALLOWED_ACCOUNT_LENGTH` instead of a local constant.

PR #2035, opened 2026-08-28
Clean up Fractal UI components

Velua:


packages/user/FractalCore/ui/src/pages/fractal/guilds/index.tsx
```diff
@@ -20,16 +22,14 @@ import {
 } from "@shared/shadcn/ui/table";
 
 import { ModalCreateGuild } from "./components/modal-create-guild";
-import { useGuilds } from "@/hooks/fractals/use-guilds";
 
 export const Guilds = () => {
     const [showModal, setShowModal] = useState(false);
 
-    const { data: guilds } = useGuilds()
+    const { data: guilds, isLoading } = useGuilds();
```
brandonfancher: I always recommend using `isPending` for cases like this instead of `isLoading`. `isPending` means there is no data yet, while `isLoading` means there is no data yet AND the query is actively fetching. `isPending` can avoid flicker and UI bugs. This comment applies to other uses of `isLoading` in this PR too.

packages/user/Fractals/ui/src/pages/loader.tsx
```diff
@@ -46,27 +42,11 @@ export const Loader = () => {
...
-                    ) : (
-                        <Button disabled={isPending} onClick={() => login()}>
-                            Login
-                        </Button>
-                    )}
-                </CardFooter>
+                <CardDescription>Log in to continue</CardDescription>
             </CardHeader>
+            <CardFooter className="flex justify-end">
+                <Button onClick={() => login()}>Login</Button>
```
brandonfancher: I realize it was "Login" before, but the button action text should say "Log in" since it's a verb.

brandonfancher (approved):
Approved, but I recommend addressing my comments.

PR #2037, opened 2026-08-28
Fix fractal token distribution and paid occupations

Velua:
- Debit TokenStream redeemer NFTs on stream create so `dist_token` and `claim_rew` can actually withdraw. 
- Default paid occupations on fractal create as we are still pushing new fractals directly into Guilds. 
- Cap 25 occupations, must be restricted somewhere given the fibbonaci function limitations.

### DIFF
--- a/packages/user/Fractals/plugin/src/lib.rs
+++ b/packages/user/Fractals/plugin/src/lib.rs
@@ -31,7 +31,10 @@ impl TrustConfig for FractallyPlugin {
                 "Creating a new fractal",
                 "Claiming accrued fractal token rewards",
             ],
-            high: &["Setting the occupation service for a fractal role"],
+            high: &[
+                "Setting the occupation service for a fractal role",
+                "Setting paid occupations for token distribution",
+            ],
         }
     }
 }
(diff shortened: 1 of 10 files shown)


packages/user/Fractals/plugin/src/lib.rs
```diff
@@ -82,6 +85,8 @@ impl AdminFractal for FractallyPlugin {
         set_role_occ(Judiciary.into());
         set_role_occ(Executive.into());
         set_role_occ(Recruitment.into());
+
+        Fractals::add_to_tx().set_paid_occ(vec![account!("guilds")]);
```
James-Mart: Consider taking this `guilds` string from the guild service wrapper directly.

James-Mart (approved):

packages/user/Fractals/plugin/src/lib.rs
Velua: Addressed in e91631985: `create_fractal` now uses `psibase::services::guilds::SERVICE` (same constant FractalGen uses) instead of `account!("guilds")`. Applied it to both `set_paid_occ` and the nearby `set_r_occ` calls so they stay consistent.

PR #2038, opened 2026-08-28
Tokens emit amount actually uncredited, not requested

Velua:
Previously the uncredit action emitted `balChanged` with the amount the caller **requested**. `SharedBalance::uncredit` only moves `min(requested, shared balance)`, so the event could report more than was actually returned.

Emit from the method after that clamp so the history amount matches the transfer. 
If the shared balance is already 0, return without emitting.


James-Mart (approved):
Good find

Issue #2016, opened 2026-08-20
Convert direct fetch requests into authenticated queries


brandonfancher:
We've tightened our CSP in #1935. We provide a restrictive default, but some apps must overwrite the CSP to loosen up the `connect-src` policy because they query other services directly from their respective subdomains.

Instead, we should update these apps to make those requests via supervisor as authenticated queries. Then we won't need the overrides.

See also: #2015

PR #2043, opened 2026-08-31
csp(#2016): route tokens/namemarket/vserver UI reads via plugins

cool-ant:
## Summary
- First PR in a 3-part stack for [#2016](https://github.com/gofractally/psibase/issues/2016): migrate UI GraphQL that already has plugin `authorized::graphql` APIs (`tokens`, `namemarket`, `vserver`) off direct sibling `fetch`.
- Add shared `authorizedPluginGraphql` helper + `vserver` plugin client; point Homepage/Config/shared-ui call sites at those APIs.
- Trim `connect-src` for those services on Homepage/Config; drop Explorer’s unnecessary `tokens` connect-src (keeps `unsafe-inline` for scripts).
- One-line plugin tweak: whitelist `config` on `tokens:plugin/authorized::graphql` so Config can use the shared system-token hook.

## Follow-ups
- PR2: add query surfaces on the ~11 plugins that still need them (invite, token-swap, producers, guilds, …).
- PR3: Config packages/`x-admin`/`http:`/`https:` special case.

## Test plan
- [ ] Boot chain with updated packages; confirm Homepage loads settings (billing + system token) without CSP console errors for `tokens`/`vserver`/`namemarket`
- [ ] Config resources + billing + name markets pages load without those connect-src errors
- [ ] Homepage tokens balances / pending / history still work via `tokens:plugin/authorized`
- [ ] Accounts create prompt still resolves system token
- [ ] Explorer still loads (inline scripts) with tightened connect-src

### DIFF
--- a/packages/shared-ui/hooks/use-billing-config.ts
+++ b/packages/shared-ui/hooks/use-billing-config.ts
@@ -1,7 +1,8 @@
 import { useQuery } from "@tanstack/react-query";
 
-import { GraphQLUrlOptions, graphql } from "@shared/lib/graphql";
+import { authorizedPluginGraphql } from "@shared/lib/graphql/authorized-plugin";
 import QueryKey from "@shared/lib/query-keys";
+import { vserver } from "@shared/lib/plugins";
 
 interface BillingConfigResponse {
     getBillingConfig: {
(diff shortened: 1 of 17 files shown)


cool-ant:
pushed/created in error

Issue #2015, opened 2026-08-20
Proposal: front-ends can only make plugin calls to a plugin on their own domain


James-Mart:
# Task

All UIs should only make plugin calls to plugins on their own subdomain. This implies that if an app needs to make a cross-domain  plugin call today, that it should instead implement a local wrapper function within its own plugin responsible for forwarding the call to the cross-domain plugin.

**Exception**: most apps call some functions like: `get-current-user`, `connect-account`, `logout`, `get-account`. Instead of every top-level app plugin needing to wrap those, let's (for now) leave them as cross-app calls. 

<details>
  <summary>Background info / justification ...</summary>
   
@brandonfancher brought up a good point, that CSP policy `connect-src` prevents data ex-filtration in xss situations, but currently plugin calls are an ex-filtration escape-hatch because the front-end could call `malicious.{{root}}` plugin and upload data to the network.

Which got us thinking whether there should be a CSP equivalent (or at least `connect-src` equivalent) in plugin-land.

Within a plugin, its import set is already explicit so basically has built-in `connect-src` protection, so the only loose-thread is the front-end. The simplest way I can think of to give the same guarantees to the front-end are to simply lock it down so a front-end can only call its own service's plugin, and all x-plugin calls are therefore whitelisted by the plugin's own import set.

</details>

James-Mart:
This also could allow some simplification to `common-lib` which no longer needs to require that the front-end specify the target service of a plugin call. The target service is always the service from which the front-end was served.

James-Mart:
After we have proper user-authentication in query service queries, we should explicitly also then check if the corresponding plugin has an `authorized:graphql` interface.

The rule is:
* If a query service has ANY authenticated queries, then the corresponding plugin's `authorized:graphql` function should be `Low` trust **at minimum**. We can whitelabel any system apps that need to call it today.

PR #2045, opened 2026-09-01
Switching over to authenticated queries only

cool-ant:
Addresses #2016 
Represents the prework / partial impl of #2015 

- no UI-direct-to-foreign-subdomain fetchs
- UI fetches by calling plugin fns via Supervisor to do fetches

NOTE: for easier review, I've split the work up largely into the first 3 commits. The final PR makes a few important tweaks, but the bulk of the work is in the first 3 commits. I suggest you keep the last commit open and if you're going to comment on anything in this PR as you review the first 3 commits, ensure your feedback isn't represented already in the "Cleaning up PR" commit.
Or just review the PR as a whole. It's not a *terribly* big change set, just *lots* of files where calls are redirected, imports change, etc.


brandonfancher:
@cool-ant @James-Mart: I recall that @Velua, in our last discussion on this, brought up some concerns related to migrating all graphql calls to plugin calls--performance considerations. Have we explicitly decided to proceed with this right now?

cool-ant:
> @cool-ant @James-Mart: I recall that @Velua, in our last discussion on this, brought up some concerns related to migrating all graphql calls to plugin calls--performance considerations. Have we explicitly decided to proceed with this right now?

Yes, the gist of it: we need to do this anyway, and we've talked in the past about making this plumbing a lot easier to write. We likely could get this really close to auto-generated for a completely open graphql endpoint. So there's a path here, and we're just choosing to plow forward and add DX improvements later.

packages/shared-ui/lib/plugins/token-swap.ts
```diff
@@ -99,18 +99,32 @@ class Liquidity extends PluginInterface {
...
 }
 
+class Authorized extends PluginInterface {
+    protected override readonly _intf = "authorized" as const;
+
+    get graphql() {
+        return this._call<[query: string], string>("graphql");
+    }
+}
+
```
brandonfancher: Could we just import this `authorized-graphql-plugin.ts`? Same goes for other plugins repeating the code for this class.

packages/shared-ui/lib/plugins/index.ts
```diff
@@ -19,6 +21,18 @@ const nameMarket = new NameMarket("namemarket");
 const tokenSwap = new TokenSwap("token-swap");
 const tokens = new Tokens("tokens");
 const profiles = new Profiles("profiles");
+const vserver = new VirtualServer("vserver");
```
brandonfancher: Could this one also be an `AuthorizedGraphqlPlugin` instance...then we can get rid of `plugins/vserver.ts`.

packages/user/Config/postinstall.json
```diff
@@ -11,7 +11,7 @@
     "method": "setcsp",
     "data": {
       "path": "*",
-      "csp": "default-src 'self';script-src 'self';style-src 'self' 'unsafe-inline';img-src 'self' data: profiles.{{root}}/avatar/ branding.{{root}};font-src 'self';connect-src 'self' producers.{{root}} sites.{{root}} staged-tx.{{root}} transact.{{root}} vserver.{{root}} tokens.{{root}} profiles.{{root}} x-admin.{{root}} packages.{{root}} namemarket.{{root}} http: https:;frame-src supervisor.{{root}};frame-ancestors 'self';base-uri 'none';form-action 'self';object-src 'none';"
+      "csp": "default-src 'self';script-src 'self';style-src 'self' 'unsafe-inline';img-src 'self' data: profiles.{{root}}/avatar/ branding.{{root}};font-src 'self';connect-src 'self' profiles.{{root}};frame-src supervisor.{{root}};frame-ancestors 'self';base-uri 'none';form-action 'self';object-src 'none';"
     }
   }
```
brandonfancher: We can remove the config CSP override here entirely, now, since the default CSP covers all of this.

packages/user/Explorer/postinstall.json
brandonfancher, line 1: We still have to keep this one. But once we rewrite Explorer, we should be able to get rid of it.

packages/user/FractalCore/postinstall.json
```diff
@@ -11,7 +11,7 @@
     "method": "setcsp",
     "data": {
       "path": "*",
-      "csp": "default-src 'self';script-src 'self';style-src 'self' 'unsafe-inline';img-src 'self' data: profiles.{{root}}/avatar/ branding.{{root}};font-src 'self';connect-src 'self' guilds.{{root}} fractals.{{root}} evaluation.{{root}} invite.{{root}} staged-tx.{{root}} profiles.{{root}};frame-src supervisor.{{root}};frame-ancestors 'self';base-uri 'none';form-action 'self';object-src 'none';"
+      "csp": "default-src 'self';script-src 'self';style-src 'self' 'unsafe-inline';img-src 'self' data: profiles.{{root}}/avatar/ branding.{{root}};font-src 'self';connect-src 'self' profiles.{{root}};frame-src supervisor.{{root}};frame-ancestors 'self';base-uri 'none';form-action 'self';object-src 'none';"
```
brandonfancher: We can get rid of this CSP override now

packages/user/Homepage/postinstall.json
```diff
@@ -13,7 +13,7 @@
         "method": "setcsp",
         "data": {
             "path": "*",
-            "csp": "default-src 'self';script-src 'self';style-src 'self' 'unsafe-inline';img-src 'self' data: profiles.{{root}}/avatar/ branding.{{root}};font-src 'self';connect-src 'self' invite.{{root}} tokens.{{root}} token-swap.{{root}} vserver.{{root}} producers.{{root}} profiles.{{root}} namemarket.{{root}};frame-src supervisor.{{root}};frame-ancestors 'self';base-uri 'none';form-action 'self';object-src 'none';"
+            "csp": "default-src 'self';script-src 'self';style-src 'self' 'unsafe-inline';img-src 'self' data: profiles.{{root}}/avatar/ branding.{{root}};font-src 'self';connect-src 'self' profiles.{{root}};frame-src supervisor.{{root}};frame-ancestors 'self';base-uri 'none';form-action 'self';object-src 'none';"
```
brandonfancher: We can get rid of this override

packages/user/Workshop/postinstall.json
```diff
@@ -11,7 +11,7 @@
     "method": "setcsp",
     "data": {
       "path": "*",
-      "csp": "default-src 'self';script-src 'self';style-src 'self' 'unsafe-inline';img-src 'self' data: profiles.{{root}}/avatar/ branding.{{root}};font-src 'self';connect-src 'self' registry.{{root}} setcode.{{root}} sites.{{root}} profiles.{{root}};frame-src supervisor.{{root}};frame-ancestors 'self';base-uri 'none';form-action 'self';object-src 'none';"
+      "csp": "default-src 'self';script-src 'self';style-src 'self' 'unsafe-inline';img-src 'self' data: profiles.{{root}}/avatar/ branding.{{root}};font-src 'self';connect-src 'self' profiles.{{root}};frame-src supervisor.{{root}};frame-ancestors 'self';base-uri 'none';form-action 'self';object-src 'none';"
```
brandonfancher: We can get rid of this override too!

brandonfancher (changes requested):

packages/user/Config/ui/src/hooks/use-available-packages.ts
```diff
@@ -1,82 +1,29 @@
...
         queryKey: QueryKey.availablePackages(),
         queryFn: async () => {
-            return (await getPackageIndex("root")).flatMap((x) => x.index);
+            const res = await supervisor.functionCall({
+                service: "packages",
+                intf: "queries",
+                method: "getAvailablePackages",
+                params: ["root"],
+            });
+
```
brandonfancher, line 26: The UI is failing for me here: `Call to packages:plugin/queries->getAvailablePackages failed`

James-Mart:
> @cool-ant @James-Mart: I recall that @Velua, in our last discussion on this, brought up some concerns related to migrating all graphql calls to plugin calls--performance considerations. Have we explicitly decided to proceed with this right now?

Also keep in mind apps can make gql queries to their own services directly, and 
1. They will still be authenticated.
2. That's compatible with a strict csp
3. Doesn't imply waterfalling 

Which means another path is to reverse proxy queries server side.

packages/user/Host/plugin/http/src/lib.rs
```diff
@@ -110,6 +182,29 @@ impl Api for HostHttp {
...
+    fn fetch_sibling_bytes(url: String) -> Result<Vec<u8>, Error> {
+        assert_packages_caller("fetch-sibling-bytes@host:http/api");
+
+        let res = fetch_sibling(&url, true)?;
+
+        match res.body {
+            Some(BridgeTypes::BodyTypes::Bytes(body)) => Ok(body),
+            _ => Err(make_error("Http response body absent or wrong type")),
+        }
+    }
```
cool-ant, line 207: @James-Mart The additions in this file were needed to 1) route Config's Packages traffic, which must be able to reach other URLs (other than just the packages subdomain) as package srouces and 2) allow binary data to pass through the XHR plumbing (required hex encoding). Lmk if any of this is counter-architecture or otherwise problematic, and note that this pathway will change (to some degree) when we switch to async XHR (which won't need the special header/content type handling)

brandonfancher (changes requested):
The Config package query's working now. But I get this when I try to install packages. (This test was attempting install of Workshop and Fractals.)

Image

packages/local/XAdmin/src/XAdmin.cpp
```diff
@@ -993,6 +993,12 @@ namespace LocalService
       {
          if (target.find("/.") != std::string_view::npos)
             return {};
+         static const std::vector<AccountNumber> packagesCors = {
+             AccountNumber{"supervisor"},
```
James-Mart: Looks wrong

cool-ant:
> The Config package query's working now. But I get this when I try to install packages. (This test was attempting install of Workshop and Fractals.)
> 
> Image

fixed in https://github.com/gofractally/psibase/pull/2070

packages/user/Explorer/postinstall.json
```diff
@@ -5,7 +5,7 @@
     "method": "setcsp",
     "data": {
       "path": "*",
-      "csp": "default-src 'self';script-src 'self' 'unsafe-inline';style-src 'self' 'unsafe-inline';img-src 'self' data: profiles.{{root}}/avatar/ branding.{{root}};font-src 'self';connect-src 'self' tokens.{{root}};frame-src supervisor.{{root}};frame-ancestors 'self';base-uri 'none';form-action 'self';object-src 'none';"
+      "csp": "default-src 'self';script-src 'self' 'unsafe-inline';style-src 'self' 'unsafe-inline';img-src 'self' data: profiles.{{root}}/avatar/ branding.{{root}};font-src 'self';connect-src 'self';frame-src supervisor.{{root}};frame-ancestors 'self';base-uri 'none';form-action 'self';object-src 'none';"
```
Velua: Can we drop `tokens` then? the explorer UI uses it to fetch token balances.

Velua (approved):

brandonfancher (approved):

packages/user/Explorer/postinstall.json
cool-ant: We obviously don't care about this app, which has been declared deprecated numerous times, but no reason to actively break it, so I've restored its CSP exceptions.

packages/shared-ui/lib/graphql/call-graphql-via-plugin.ts
```diff
@@ -0,0 +1,23 @@
...
+    type PluginCall,
+    callPluginFunction,
+} from "@shared/lib/plugins/lib/call-plugin-function";
+
+type GraphqlViaPluginResponse<T> = {
+    data: T;
+    errors?: Array<{ message: string }>;
+};
+
+export async function callGraphqlViaPlugin<T>(
```
James-Mart, line 11: Odd name. A dev wouldn't know why they should call this vs just making a direct graphql query. Something like `graphqlAuth` with an appropriate doc comment would be more useful, and the fact that it goes through a plugin is an implementation detail.

packages/local/XAdmin/src/XAdmin.cpp
```diff
@@ -993,6 +993,12 @@ namespace LocalService
...
+             AccountNumber{"config"},
+         };
+         if (req.method == "OPTIONS")
+            return HttpReply{.headers = allowCors(req, packagesCors)};
```
James-Mart, line 1001: Why keep config? Can't this just be `allowCors(req, "supervisor"_a)`?

And then to keep the same permissions, we made sure the path to make this query through plugins only authorizes the config app.

packages/user/Host/plugin/http/wit/world.wit
```diff
@@ -23,6 +23,20 @@ interface api {
...
+    /// Parameters
+    /// * `url` - A full URL whose host is a sibling service on this chain
+    fetch-sibling-json: func(url: string) -> result<string, error>;
+
+    /// Fetch binary data from another service on the current root domain.
+    /// Only callable by the `packages` plugin.
+    ///
+    /// Parameters
+    /// * `url` - A full URL whose host is a sibling service on this chain
+    fetch-sibling-bytes: func(url: string) -> result<list<u8>, error>;
```
James-Mart, line 39: I would much prefer the interface you deleted here:

https://github.com/gofractally/psibase/pull/2003/changes#diff-cbe39ab0db886cd5d2e9892e2b6cfa10b25d0fa3d94b6468e9f63734fe9a2772L56

And you can just add `get`:
```
interface admin {
    use host:types/types.{post-request, get-request, body-types, error};

    /// Make a post request
    post: func(app: string, request: post-request) -> result<option<body-types>, error>;

    /// Make a get request
    get: func(app: string, request: get-request) -> result<option<body-types>, error>;

}
```

The functions can whitelist allowed callers like `packages`.
And you can leave the `admin` interface out of the `imports` world.

packages/user/Packages/plugin/src/lib.rs
```diff
@@ -416,14 +527,10 @@ impl Queries for PackagesPlugin {
...
-        let json = Server::post_graphql_get_json(&format!(
-            "query {{ sources(account: {}) {{ url account }} }}",
-            serde_json::to_string(&owner).unwrap(),
-        ))?;
-        let result: QueryRoot<SourcesQuery> =
-            serde_json::from_str(&json).map_err(|e| ErrorType::JsonError(e.to_string()))?;
-        Ok(result.data.sources)
+        fetch_sources(owner)
+    }
+    fn get_available_packages(owner: String) -> Result<Vec<types::PackageInfo>, HostTypes::Error> {
```
James-Mart, line 532: This used to live in the js, so was restricted to the config app. 
Any reason to keep that restriction? @swatanabe

packages/user/Supervisor/ui/src/plugin/plugin-host.ts
```diff
@@ -72,20 +72,44 @@ export class PluginHost implements HostInterface {
...
+        const headers = convert(req.headers);
+        const accept = (
+            headers["Accept"] ||
+            headers["accept"] ||
+            ""
+        ).toLowerCase();
+        return (
+            accept.includes("application/octet-stream") ||
+            accept.includes("application/zip") ||
+            accept === "*/*"
```
James-Mart, line 109: This is very broken. 

1. The accept header is what the client accepts and can list more than one type. By themselves, they do not express preference. That's what the `q` value is for.
1. `accept == "*/*` does not mean the requester wants binary

As written, anyone who is okay with a binary response (or expresses no opinion with `*/*`) will cause the response to be forcibly interpreted as binary.

You should always use `overrideMimeType` so you get raw bytes, then interpret from the response Content-Type.

packages/user/Packages/plugin/src/lib.rs
swatanabe: The reason this was in JS is that it cannot be implemented in the plugin. The URL is specifically to handle package repositories that are hosted on a different chain or a regular HTTP server.

James-Mart: So we should either 
1) revert these changes and go back to direct fetch from the UI, or
2) add restricted capabilities to `host:http` to allow making http requests anywhere, with only packages whitelisted

packages/user/Supervisor/ui/src/plugin/plugin-host.ts
cool-ant: yeah, I get it.
ok, one detail semi-hidden here: unknown content types become bytes instead of throwing "Unsupported content type". I think that's the right default, but it is a behavior change worth getting your alignment on.

PR #2048, opened 2026-09-02
Update CI builder image SHA

brandonfancher:

PR #2051, opened 2026-09-02
Fix psibase on new CI image

James-Mart:
# Background

https://github.com/gofractally/psibase/pull/2048 uses a new builder image for psibase CI. The new image has an upgraded yarn, npm, and rust. This PR fixes various issues that surfaced from making those upgrades.

# `bind_env`

On 1.86, a bare `extern "C"`  (e.g. `extern "C" { fn writeConsole(...); }`) on wasm32-wasip1 was an unresolved C symbol. rust-lld’s wasm default:
 1. turned those into `env` imports, and
 2. if a `#[no_mangle]` of the same name existed, bound the reference to that local function instead

On 1.98, that target links `wasi-libc`. A bare extern "C" is a normal C symbol, not a wasm import. Link fails with e.g. `undefined symbol: writeConsole`

`#[link(wasm_import_module = "env")]` is the explicit form needed for those functions to be considered wasm imports. In production that's all that's needed, but the import is now no longer bound to a local `#[no_mangle]` which is needed for tests. Therefore we also need `bind_env` in the cargo-psibase post-processing. `bind_env` will bind those wasi-imports to the same-named local exports and drop the import.

# missing polyfill

 `fd_fdstat_get` is needed for rust services, was already in cpp.
 
# updated error text 

The error text asserted in the `authorized` trybuild test was updated at some point, this updates the snapshot to the latest expected text.

# workspace yarn

- Bump the Yarn workspace to 4.18.0
- `enableScripts`, `approvedGitRepositories`, `npmMinimalAgeGate` were added to keep existing Yarn 4.9 behavior with default changes in latest yarn version

### DIFF
--- a/rust/cargo-psibase/src/link.rs
+++ b/rust/cargo-psibase/src/link.rs
@@ -763,6 +763,100 @@ pub fn link_module(source: &Module, dest: &mut Module) -> Result<(), anyhow::Err
     Ok(())
 }
 
+struct RetargetCalls {
+    from: FunctionId,
+    to: FunctionId,
+}
+
+impl walrus::ir::VisitorMut for RetargetCalls {
+    fn visit_function_id_mut(&mut self, function: &mut FunctionId) {
+        if *function == self.from {
+            *function = self.to;
+        }
+    }
+}
+
+fn retarget(module: &mut Module, from: FunctionId, to: FunctionId) {
+    let mut visitor = RetargetCalls { from, to };
+    for (_, local) in module.funcs.iter_local_mut() {
+        let entry = local.entry_block();
+        walrus::ir::dfs_pre_order_mut(&mut visitor, local, entry);
+    }
+    if module.start == Some(from) {
+        module.start = Some(to);
+    }
+    for elem in module.elements.iter_mut() {
+        if let walrus::ElementKind::Active {
+            offset: walrus::InitExpr::RefFunc(fid),
+            ..
+        } = &mut elem.kind
+        {
+            if *fid == from {
+                *fid = to;
+            }
+        }
+        for member in &mut elem.members {
+            if *member == Some(from) {
+                *member = Some(to);
+            }
+        }
+    }
+    for export in module.exports.iter_mut() {
+        if let walrus::ExportItem::Function(fid) = &mut export.item {
+            if *fid == from {
+                *fid = to;
+            }
+        }
+    }
+}
+
+/// Bind `env` imports to local exports of the same name.
+///
+/// psitest does not provide `env.*`; the tester polyfill does.
+pub fn bind_env(module: &mut Module) -> Result<(), anyhow::Error> {
+    let mut replacements = Vec::new();
+    for import in module.imports.iter() {
+        if import.module != "env" {
+            continue;
+        }
+        let walrus::ImportKind::Function(import_fid) = import.kind else {
+            continue;
+        };
+        let import_fid = import_fid;
+        let Ok(export_fid) = module.exports.get_func(&import.name) else {
+            continue;
+        };
+        if import_fid == export_fid {
+            continue;
+        }
+        if !matches!(module.funcs.get(export_fid).kind, FunctionKind::Local(_)) {
+            continue;
+        }
+        if module.func_type(import_fid) != module.func_type(export_fid) {
+            return Err(anyhow!(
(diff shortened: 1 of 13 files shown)


brandonfancher (approved):

rust/cargo-psibase/src/link.rs
```diff
@@ -763,6 +763,100 @@ pub fn link_module(source: &Module, dest: &mut Module) -> Result<(), anyhow::Err
...
+pub fn bind_env(module: &mut Module) -> Result<(), anyhow::Error> {
+    let mut replacements = Vec::new();
+    for import in module.imports.iter() {
+        if import.module != "env" {
+            continue;
+        }
+        let walrus::ImportKind::Function(import_fid) = import.kind else {
+            continue;
+        };
+        let import_fid = import_fid;
```
swatanabe: Why is this needed?

rust/cargo-psibase/src/link.rs
```diff
@@ -763,6 +763,100 @@ pub fn link_module(source: &Module, dest: &mut Module) -> Result<(), anyhow::Err
...
+            }
+        }
+    }
+    for export in module.exports.iter_mut() {
+        if let walrus::ExportItem::Function(fid) = &mut export.item {
+            if *fid == from {
+                *fid = to;
+            }
+        }
+    }
```
swatanabe, line 819: Replacement of global initializers is missing. eos-vm doesn't implement reference types, but they are part of the wasm spec.

rust/cargo-psibase/src/link.rs
```diff
@@ -763,6 +763,100 @@ pub fn link_module(source: &Module, dest: &mut Module) -> Result<(), anyhow::Err
...
+            continue;
+        };
+        let import_fid = import_fid;
+        let Ok(export_fid) = module.exports.get_func(&import.name) else {
+            continue;
+        };
+        if import_fid == export_fid {
+            continue;
+        }
+        if !matches!(module.funcs.get(export_fid).kind, FunctionKind::Local(_)) {
```
swatanabe, line 841: Is there a reason for distinguishing between a re-export of an import and a local function? I suppose that if it did happen, we would need to resolve the transitive closure. If we don't support it, though, this case should probably be an error.

rust/cargo-psibase/src/link.rs
```diff
@@ -763,6 +763,100 @@ pub fn link_module(source: &Module, dest: &mut Module) -> Result<(), anyhow::Err
     Ok(())
 }
 
+struct RetargetCalls {
+    from: FunctionId,
+    to: FunctionId,
```
swatanabe: I would use a `HashMap` to retarget all the functions in a single pass

swatanabe (approved):

swatanabe:
`extern "C"` + `wasm_import_module` uses normal rust mangling which includes a crate identifier that isn't stable. There doesn't seem to be any way to prevent mangling while also setting it as a wasm import.
```
00000000       F *UND*  00000000 _RNvNtCsj8jRbVMTH0i_7psibase10native_raw8kvOpenAt
```

James-Mart:
> extern "C" + wasm_import_module uses normal rust mangling which includes a crate identifier that isn't stable. There doesn't seem to be any way to prevent mangling while also setting it as a wasm import.

There is a mangled name but I don't think it's relevant to `bind_env`, for example:

Here's the wasm import section:
```
(import "env" "kvOpenAt" (func $_RNvNtCs6AcCgPQwl7k_7psibase10native_raw8kvOpenAt …))
```

`bind_env` reads `import.module` and `import.name` from that section, so it matches `env` + `kvOpenAt`

rust/cargo-psibase/src/link.rs
James-Mart: It's not - removed

rust/cargo-psibase/src/link.rs
James-Mart: Added

rust/cargo-psibase/src/link.rs
James-Mart: Now uses an error

rust/cargo-psibase/src/link.rs
James-Mart: Done

swatanabe (approved):

PR #2066, opened 2026-09-04
Xadmin fixes

James-Mart:
Cleans up xadmin build and console warnings, and cleans up errors related to when the node is isn't live (booted or connected to a network).

Also allows the other pages in xadmin to be visible before boot instead of redirecting, and improves how those pages render when the node isn't live.


swatanabe (commented):
The few service changes look okay. This PR is mostly TS, so I won't give an approval.

brandonfancher (approved):

PR #2067, opened 2026-09-04
Implement runAs using isAuthSys instead of checkAuthSys

swatanabe:
`isAuthSys` seems more appropriate since `runAs` is authorized by another account much like a staged transaction. This also simplifies `checkAuthSys` as it no longer has to do double duty.

- `allowedActions` are authorized by verifying that the caller can authorize each action
- all the type flags `runAsReqesterReq`, `runAsMatchedReq`, etc are mostly handled in the same way by services. We just implement the standard behavior and make it non-configurable. The only thing this breaks is `runAs` an account using `AuthAny`, which I don't think we care about anymore.


packages/system/Transact/src/Transact.cpp
```diff
@@ -346,6 +346,173 @@ namespace SystemService
...
+         auto [iter, _] = authorized.try_emplace(account);
+         if (!iter->second.authorized)
+         {
+            setAuthorized(*iter);
+         }
+      }
+
+      void setAuthorized(AuthItem& self)
+      {
+         self.second.authorized = true;
```
James-Mart, line 407: If it's already authorized, should this be idempotent?

packages/system/Transact/src/Transact.cpp
```diff
@@ -346,6 +346,173 @@ namespace SystemService
...
+         }
+      }
+      void getDelegates(AuthItem& self)
+      {
+         auto delegates = to<AuthInterface>(self.second.authService)
+                              .getDlgsSys(self.first, self.second.getAction());
+         for (auto child : delegates)
+         {
+            auto [iter, inserted] = authorized.try_emplace(child);
+            if (!iter->second.queued)
```
James-Mart, line 445: Couldn't we avoid enqueuing again here if it was already authorized?

packages/system/Transact/src/Transact.cpp
```diff
@@ -346,6 +346,173 @@ namespace SystemService
...
+         while (!stack.empty())
+         {
+            auto item = stack.back();
+            stack.pop_back();
+            item->second.queued = false;
+            if (item->second.refs > 0)
+            {
+               if (item->second.authService == psibase::AccountNumber{})
+               {
+                  item->second.authService = to<Accounts>().getAuthOf(item->first);
```
James-Mart, line 475: Worth caching `getAuthOf`?

packages/system/Transact/src/Transact.cpp
James-Mart: I can't figure out if it's actually a problem but not being idempotent the `decref` worries me

packages/system/Transact/src/Transact.cpp
swatanabe: This *is* where `getAuthOf` is cached

packages/system/Transact/src/Transact.cpp
swatanabe: `setAuthorized` should only be called once for each `AuthItem`

packages/system/Transact/src/Transact.cpp
swatanabe: Yes we should, and it's essential to maintain the invariant that no items that have been authorized are in the queue.

James-Mart (approved):

PR #2071, opened 2026-09-07
Fractal tester fixes

James-Mart:
Found that fractal tester was using the wrong account names in postinstall. This fixes that and also improves error messages to aid in debugging. Packaging now fails if the specified `AccountNumber` is invalid.


rust/psibase/src/package.rs
```diff
@@ -283,7 +283,7 @@ pub type SchemaMap = HashMap<AccountNumber, Schema>;
 #[serde(rename_all = "camelCase")]
 pub struct PrettyAction {
     /// Account sending the action
-    pub sender: AccountNumber,
+    pub sender: ExactAccountNumber,
```
swatanabe, line 286: I can't remember why `service` uses `String` and not `ExactAccountNumber`, but `sender` and `service` should probably be consistent. I do remember that the only reason `sender` is not checked is that I didn't need it when I added the check for `service`.

None of the reasons I can think of for using `String` actually make sense
- ❌ It shouldn't use `ExactAccountNumber`, because we only need it to be valid. It doesn't need to be canonical. It's parsed using `from_exact`, though, and it has been this was since the checking was originally added in 260af90a66.
- ❌ The error should be deferred somewhere. I can't find any use of `PrettyAction` where it doesn't make sense to produce an error for an invalid account immediately.
- ❌ It should be compatible with the C++ declaration. `PrettyAction` only supports JSON serialization, not fracpack, and the JSON format is a string either way.

James-Mart: Updated `service` to `ExactAccountNumber`

swatanabe (approved):

PR #2072, opened 2026-09-07
auto select/unselect dependent packages in config

James-Mart:
Fractaltester was failing to install, and then I realized it was because it was pulling in the fractalgen package, which needs to be installed at boot. The UI did not represent that the fractalgen package was also being installed. This fixes that and now selecting a package also selects any dependent packages. unselecting a dependent package also unselects the depender package.


brandonfancher:
I recently fixed this same problem in the XAdmin boot wizard UI. This rhymes. I wonder if there's an opportunity here to share that logic.

James-Mart:
> I recently fixed this same problem in the XAdmin boot wizard UI. This rhymes. I wonder if there's an opportunity here to share that logic.

Def rhymes. The opportunity is smaller than you might imagine because there are real differences between the implementations, but I extracted the dependent package search.

brandonfancher (approved):
Tested. Seems to be working well...both in x-admin and config

PR #2075, opened 2026-09-08
Fix vulnerability in AuthSig that enables account rekeying

cool-ant:


packages/system/AuthSig/src/AuthSig.cpp
```diff
@@ -100,10 +100,10 @@ namespace SystemService
...
       {
-         auto authTable = open<AuthTable>(KvMode::readWrite);
-         authTable.put(AuthRecord{.account = name, .pubkey = std::move(key)});
-
-         to<Accounts>().newAccount(name, AuthSig::service, true);
+         bool created = to<Accounts>().newAccount(name, AuthSig::service, true);
+         check(created, "account already exists");
+         open<AuthTable>(KvMode::readWrite)
+             .put(AuthRecord{.account = name, .pubkey = std::move(key)});
       }
```
James-Mart, line 107: Hmm, I think the issue might be with the `Accounts` service, actually

James-Mart: Oh, I see.

https://github.com/gofractally/psibase/pull/1824/changes#diff-907e0cc95a5295b9f4d087bd931fbfc6835473b8ad3fa81e5983dd91ff6fc184

This changed the `newAccount` flow so that it returns false instead of aborting if account creation fails, but there were consumers (notably in AuthSig) depending on the abort behavior

James-Mart:
This bug is not limited to AuthSig.

I see `NameMarket::buy` and `AuthDyn::newAccount` were also depending on the `abort` behavior of `Accounts`

James-Mart:
~~Terrifying~~ Great find, by the way.

swatanabe:
> I see `NameMarket::buy` and `AuthDyn::newAccount` were also depending on the `abort` behavior of `Accounts`

`AuthDyn` is okay (well, it's not quite right, but it isn't vulnerable to this bug because it checks whether the manager is already set).

packages/system/Accounts/include/services/system/Accounts.hpp
James-Mart: Doc comment

packages/system/AuthSig/src/AuthSig.cpp
```diff
@@ -103,7 +103,7 @@ namespace SystemService
          auto authTable = open<AuthTable>(KvMode::readWrite);
          authTable.put(AuthRecord{.account = name, .pubkey = std::move(key)});
 
-         to<Accounts>().newAccount(name, AuthSig::service, true);
+         to<Accounts>().newAccount(name, AuthSig::service, NewAccountMode::requireNew);
```
cool-ant, line 106: Any reason not to order this like AuthDelegate and (somewhat) like AuthDyn where `newAccount()` is called, the check is made, *then* the `.put()` is called?
AuthSig, in one sense it the one-off that does the `.put()` first and then *relies* on an abort to reverse what's already been `.put()`.

packages/system/AuthSig/test/src/AuthSig-test.cpp
```diff
@@ -221,3 +221,32 @@ SCENARIO("AuthSig")
...
+      REQUIRE(attacker.to<AuthSig::AuthSig>().newAccount(victim.id, bob_keys.first)
+                  .failed("account already exists"));
+
+      REQUIRE(victim.to<AuthSig::AuthSig>().setKey(alice_keys.first).succeeded());
+
+      auto stolen = t.from(victim.id).with({bob_keys});
+      REQUIRE(stolen.to<AuthSig::AuthSig>().setKey(bob_keys.first).failed(""));
+   }
+
+   THEN("newAccount still creates a genuinely new account")
```
James-Mart, line 246: I don't think this tests what it claims, and also not sure why it's needed. Wouldn't it also have passed before this PR fix?

James-Mart (approved):

packages/system/AuthSig/test/src/AuthSig-test.cpp
swatanabe: It looks like it, but I didn't change these tests at all. @cool-ant

cool-ant: removed

packages/system/AuthSig/src/AuthSig.cpp
cool-ant: @swatanabe ?

swatanabe: The order doesn't matter. That's how transactions work. Relying on it is not anything strange.

PR #2076, opened 2026-09-08
Prep UIs to call only their own plugins

cool-ant:
This PR doesn't yet *restrict* UIs to only calling their plugins, but it does a lot of the prework of rerouting UI calls through their services' plugins.


packages/user/Workshop/plugin/src/lib.rs
```diff
@@ -111,6 +111,12 @@ impl Mail for WorkshopPlugin {
         set_propose_latch(Some(&app)).unwrap();
         chainmail::plugin::api::save(msg_id)
     }
+
+    fn get_msgs(app: String) -> Result<Vec<chainmail::plugin::types::Message>, Error> {
+        assert_authorized(FunctionName::get_msgs)?;
+        set_propose_latch(Some(&app)).unwrap();
```
Velua, line 117: Do we need to `set_propose_latch` for the query?

Velua (changes requested):
Config still calls packages.privateApi — `packages/user/Config/ui/src/hooks/use-install-packages.ts:34` 

Also in `packages/user/FractalCore/ui/src/lib/graphql/fractals/get-invite-by-id.ts` we're calling Invite, guessing we don't want that either

packages/user/Workshop/plugin/src/lib.rs
cool-ant: good catch. removed

packages/shared-ui/components/tables/table-contact.tsx
brandonfancher, line 1: This component encapsulated a lot of internal queries/fetches/errors and loading states prior to this refactor. Now, all of that has to be done manually anywhere this component is invoked. That's a big loss and adds to the risk that the component will not operate consistently wherever it's used.

Can we keep the hook calls (`useCurrentUser`, `useHasProfileReadPermission`, and `useContacts` in this component and instead pass in the functions that `useHasProfilesReadPermission` and `useContacts` needs to invoke (_e.g._, `fractalCorePlugin.contacts.hasReadPermission`_ and `fractalCorePlugin.contacts.get` respectively when invoked from the fractals apps?) Then we'll keep the complexity encapsulated and ensure consistency.

packages/shared-ui/components/show-contacts-button.tsx
brandonfancher, line 1: Same question here.

packages/shared-ui/hooks/use-account-markets.ts
brandonfancher, line 1: This is no longer a hook. The file should be renamed and moved appropriately.

packages/shared-ui/hooks/use-billing-config.ts
brandonfancher, line 1: This is no longer a hook. The file should be renamed and moved appropriately.

packages/shared-ui/hooks/use-system-token.ts
brandonfancher, line 1: This is no longer a hook. The file should be renamed and moved appropriately.

packages/shared-ui/hooks/use-user-token-balances.ts
brandonfancher, line 1: This is no longer a hook. The file should be renamed and moved appropriately.

packages/shared-ui/lib/graphql/tokens.ts
```diff
@@ -32,12 +34,15 @@ const zUserTokenBalanceSchema = z.object({
 
 export type UserTokenBalanceNode = z.infer<typeof zUserTokenBalanceNodeSchema>;
 
-export const fetchUserTokenBalances = async (username: string) => {
+export const fetchUserTokenBalances = async (
+    graphql: GraphqlPluginCall,
```
brandonfancher, line 38: You use `graphql` as the arg name in several instances throughout this PR. That was confusing to me. It makes me think it refers to the GraphQL query string itself. Could we use something like `call` or `pluginCall` instead? Or something more accurate.

packages/shared-ui/hooks/use-user-token-balances.ts
brandonfancher: Actually, if you look at my comments below, I think the hook should be returned to this shared file and the plugin function call passed in. Then it can be consumed by Accounts and Homepage. 

Please check the other hooks where I made the same comment about it no longer being a hook. I suspect the hook was also removed from those files only to be duplicated elsewhere.

packages/system/Accounts/ui/src/hooks/use-marketplace-data.ts
```diff
@@ -0,0 +1,78 @@
...
+        queryFn: async (): Promise<UserTokenBalance[]> => {
+            const nodes = await callPluginFunction(
+                accounts.tokens.getUserBalances,
+                [zAccount.parse(username)],
+            );
+            return toUserTokenBalances(nodes);
+        },
+        enabled: !!username && (options?.enabled ?? true),
+    });
+};
```
brandonfancher, line 58: This is almost entirely duplicated in `Homepage/ui/src/hooks/use-user-token-balances.ts`. Should be put back into the hook file it was removed from. Please check other cases where this happened too.

brandonfancher (changes requested):
I'm not done with review entirely yet. I'll pick back up after you resolve these. I recommend reading all my feedback before addressing any of it as later comments have bearing on earlier ones.

packages/system/Accounts/plugin/accounts/wit/world.wit
```diff
@@ -56,7 +56,25 @@ interface admin {
...
+    get-user-balances: func(user: string) -> result<list<user-balance>, error>;
+}
+
+interface name-market {
+    use host:types/types.{error};
+    use name-market:plugin/types.{markets-overview};
+
+    can-create-account: func() -> bool;
+    get-markets-overview: func() -> result<markets-overview, error>;
+}
```
James-Mart: Are these only called from the accounts prompt UI? If so these functions can just be added into the prompt interface. Then the prompt UI just has one interface with which it interacts for its functionality.

packages/system/Accounts/plugin/accounts/wit/world.wit
```diff
@@ -56,7 +56,25 @@ interface admin {
...
+
 world imports {
     import api;
     import active-app;
+    import tokens;
+    import name-market;
```
James-Mart: This isn't functionality of the accounts plugin you're looking to export to third-party consumers of the accounts plugin. So I'd not include them in the imports world.

packages/system/Accounts/plugin/accounts/src/interfaces/tokens_and_markets.rs
James-Mart: I think the implementations for all of these functions should be limited to be called only by `accounts` itself.

packages/system/VirtualServer/plugin/src/lib.rs
```diff
@@ -233,6 +233,19 @@ impl Authorized for VirtualServerPlugin {
 
         host::server::post_graphql_get_json(&query)
     }
+
+    fn get_billing_config() -> Result<Exports::types::BillingConfig, Error> {
+        assert_caller(
+            &["homepage", "config", &virtual_server::SERVICE.to_string()],
+            "authorized::get_billing_config",
+        );
```
James-Mart: Probably okay if this query is embedded elsewhere, the query itself is public

packages/user/Config/plugin/src/lib.rs
James-Mart: All of the new functions you added should be restricted to be callable only by the receiver (config).

packages/system/Accounts/plugin/accounts/wit/impl.wit
James-Mart: # ❗ New pattern alert❗

A plugin on subdomain `A` now has two relatively distinct purposes.
1) The ingress point for **their own UI** to initiate client-side communication with other plugins (`B`, `C`, `D`).
2) The ingress point for **other plugins** (`B`,`C`,`D`) to access client or server data or functionality of `A`

I think combining these purposes within a plugin is confusing.

**Therefore I propose the following pattern** 

> [!IMPORTANT]
> Apps with both own UIs and services should have a minimum of two plugins:
> 1) `package app:ui` - The plugin used by its own UI, depends on other plugins its UI needs
> 2) `package app:plugin` - The plugin used by other plugins to access `app` functionality

## Example 1 - `Accounts`

The `accounts` app has both a UI (prompt page) and a service. Therefore it has:
* `package accounts:prompt` - It's plugin for its own UI
* `package accounts:plugin` - It's plugin for use by other plugins

> [!NOTE]
> `Accounts` already also has other plugins for other reasons, but the two listed above are the two needed to demonstrate this particular pattern.

## Example 2 - `Explorer`

The `explorer` app has both a UI (the explorer) and a service with queries needed by other plugins. Therefore it ought to have:
* `package explorer:ui` - It's plugin for its own UI - depends on virtual-server, packages, producers, etc., all the plugins from which it fetches data for rendering its own UI.
* `package explorer:plugin` - It's plugin for exposing the interface to which others link when they specifically want data hosted by the explorer server (block log)

#

packages/user/Config/ui/src/hooks/use-reject-staged.ts
```diff
@@ -9,9 +9,9 @@ export const useRejectStaged = () => {
 
     return usePluginMutation<[number]>(
         {
-            intf: "respondent",
+            intf: "staged",
             method: "reject",
-            service: "staged-tx",
+            service: "config",
```
James-Mart: In other cases you used 

> import { CONFIG } from "@/lib/services";

packages/user/FractalCore/plugin/src/lib.rs
James-Mart: * UI/api plugin split
* UI plugin functions not embeddable

packages/user/Homepage/plugin/src/lib.rs
James-Mart: None should be embeddable

packages/user/Tokens/plugin/wit/world.wit
```diff
@@ -148,11 +162,18 @@ interface user-config {
...
+    use types.{decimal, system-token-info, user-balance};
 
     /// Fetches network token, returns TID
     fetch-network-token: func() -> result<option<u32>, error>;
 
+    /// Fetches the configured system/network token with display metadata.
+    /// Returns none when no system token is configured.
+    get-system-token: func() -> result<option<system-token-info>, error>;
+
+    /// Fetches primary token balances for a user (UI-shaped rows).
```
James-Mart: ~~(UI-shaped rows)~~

packages/user/Workshop/plugin/src/lib.rs
```diff
@@ -30,7 +30,7 @@ psibase::define_trust! {
         ",
     }
     functions {
-        None => [],
+        None => [get_msgs],
```
James-Mart: There's a clever way you could make this embeddable safely, but I think for simplicity for now this should just be `Max` (not `None`) implying only `Workshop` can use it for now.

James-Mart (changes requested):

packages/shared-ui/components/tables/table-contact.tsx
cool-ant: Well, we *can*, but that's the exact pattern you and I agreed was a kludge and should work that way. I'll look at potentially some composition approach to keep this code in one place without having to pass in the functions paths.

PR #2090, opened 2026-09-15
add fractal donations / income

Velua:
Allows fractals to receive income + donations in it's own token. 

Using the `donate` action will debit the sender for the fractal token, placing it into a donation sub balance, when token distribution occurs next it will be added to the distribution amount coming from token stream rewards. 

`income` action will add the token amount to the reward stream.


packages/user/FractalCore/plugin/src/lib.rs
```diff
@@ -256,6 +258,16 @@ impl UserFractal for FractalCorePlugin {
         assert_authorized(FunctionName::join_fractal)?;
         GuildsPlugin::user_guild::join_fractal()
     }
+
+    fn donate(token_id: u32, amount: String) -> Result<(), Error> {
+        assert_authorized(FunctionName::donate)?;
+        FractalsPlugin::user_fractal::donate(token_id, &amount)
+    }
+
+    fn income(token_id: u32, amount: String) -> Result<(), Error> {
```
James-Mart: Probably not really needed as a plugin function, is it? Mostly for services that are considered to be services of the fractal to send their income to the fractal?

Doesn't hurt so not blocking merge, but wanted you to consider removing for simpler plugin api.

packages/user/Fractals/plugin/Cargo.toml
```diff
@@ -32,6 +32,7 @@ world = "impl"
 "transact:plugin" = { path = "../../../system/Transact/plugin/wit/world.wit" }
 "guilds:plugin" = { path = "../../Guilds/plugin/wit/world.wit" }
 "accounts:query" = { path = "../../../system/Accounts/plugin/query/wit/world.wit" }
+"tokens:plugin" = { path = "../../Tokens/plugin/wit/world.wit" }
```
James-Mart, line 35: Don't you also need to add a package dep to the Fractals package on the Tokens package?

packages/user/Fractals/service/src/lib.rs
```diff
@@ -161,6 +162,45 @@ pub mod service {
...
+        Wrapper::emit()
+            .history()
+            .donated(fractal.account, get_sender(), token_id, amount);
+    }
+
+    /// Contribute income to the fractal that issued `token_id`.
+    ///
+    /// Debits `amount` from the sender and deposits it into that fractal's
+    /// vesting stream. [`dist_token`] later withdraws vested income and splits
+    /// it to members.
```
James-Mart: "splits it to members" isn't accurate.
Same inaccuracy in bindings

James-Mart (approved):

PR #2092, opened 2026-09-18
Add XKeys service

swatanabe:
This is a local service that can sign transactions on behalf of other local services

How to use this service:

1. Create a key by calling `newKey`
2. Create an account using this key, or change the key of an existing account
3. Call `signTx` to sign a transation with the account as the sender
4. Find the push_transaction URL, e.g. `http://transact.psibase.localhost:8080/push_transaction?wait_for=final`
5. Use `x-http::sendRequest` to `POST` the transaction
6. Handle the response when it comes back


packages/local/XKeys/src/XKeys.cpp
```diff
@@ -0,0 +1,158 @@
...
+SignedTransaction XKeys::signTx(std::vector<psibase::Action> actions)
+{
+   auto sender = getSender();
+
+   // Get available keys
+   std::vector<std::pair<Claim, PrivateKeyInfo>> keys;
+   PSIBASE_SUBJECTIVE_TX
+   {
+      keys.clear();
+      for (auto row : open<KeyTable>().getIndex<0>().subindex(sender))
```
James-Mart: This could add keys to the tx in an order that does not correspond to the order required by the actions (first action auth needs to be first claim)

swatanabe: I think we should eliminate that rule. The logic behind it is flawed and RTransact doesn't actually work that way.

swatanabe: Specifically, a signature does not attest the validity of any other signature, so billing signature validation time after checking the first auth buys nothing.

James-Mart: Sounds good, merging as is is fine with me. Consider a gh issue to note the removal of that rule

James-Mart (approved):

